PR #1, opened 2017-01-25
Set path only for build inside node.

sebastienvas:


sebastienvas:
istio-testing/presubmit passed

sebastienvas:
istio-testing/presubmit passed

sebastienvas:
istio-testing/presubmit passed

PR #104, opened 2017-03-01
Pushing docker images to gcr.io

sebastienvas:
the withDockerRegistry pipeline step does prevent docker push to gcr.io. I tried to find the source code to understand why, but I could not find it. This is the reason we need to run the command twice.

### DIFF
--- a/resources/release-docker
+++ b/resources/release-docker
@@ -2,7 +2,7 @@
 
 # Example usage:
 #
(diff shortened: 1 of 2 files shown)


ayj:
That's unfortunate, but I guess not a huge deal if bazel caches the docker images.

ayj (approved):

PR #155, opened 2017-03-31
Create a new branch for fast forward

yutongz:
https://github.com/istio/istio-testing/issues/113

### DIFF
--- a/tools/github_helper/main.go
+++ b/tools/github_helper/main.go
@@ -94,9 +94,17 @@ func newHelper(r *string) (*helper, error) {
 }
 
 // Create a new branch for fast forward
-func (h helper) createBranchForFastForward(commit *string) error {
-	branchName := fmt.Sprintf("fastForward-%s", (*commit)[0:7])
+func (h helper) createBranchForFastForward(commit *string) (*string, error) {
+	branchName := fmt.Sprintf("fastForward-%s-%s", h.Head, (*commit)[0:7])
 	ref := fmt.Sprintf("refs/heads/%s", branchName)
+
+	if refer, resp, err := h.Client.Git.GetRef(context.TODO(), h.Owner, h.Repo, ref); resp.StatusCode != 404 {
+		if refer != nil {
+			err = fmt.Errorf("Branch %s already exists in repo %s!", branchName, h.Repo)
+		}
+		return nil, err
+	}
+
 	gho := github.GitObject{
 		SHA: commit,
 	}
@@ -106,9 +114,10 @@ func (h helper) createBranchForFastForward(commit *string) error {
 	}
 
 	if _, _, err := h.Client.Git.CreateRef(context.TODO(), h.Owner, h.Repo, &r); err == nil {
-		return nil
+		log.Printf("Created a new branch for fast forward: %s", branchName)
+		return &branchName, nil
 	} else {
-		return err
+		return nil, err
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tools/github_helper/main.go
```diff
@@ -93,16 +93,55 @@ func newHelper(r *string) (*helper, error) {
 	}
 }
 
+// Create a new branch for fast forward
+func (h helper) createBranchForFastForward(commit *string) (*string, error) {
+	count := 0
+
+	for true {
```
sebastienvas: please use something like fastForward/SHORTSHA

sebastienvas (changes requested):
PTAL

tools/github_helper/main.go
```diff
@@ -93,16 +93,42 @@ func newHelper(r *string) (*helper, error) {
 	}
 }
 
+// Create a new branch for fast forward
+func (h helper) createBranchForFastForward(commit *string) error {
+	branchName := fmt.Sprintf("fastForward-%s", (*commit)[0:7])
```
sebastienvas: Can we do something like fastForward/${headBranch ie master}-GIT_SHORT_SHA.

tools/github_helper/main.go
```diff
@@ -93,16 +93,42 @@ func newHelper(r *string) (*helper, error) {
 	}
 }
 
+// Create a new branch for fast forward
+func (h helper) createBranchForFastForward(commit *string) error {
```
sebastienvas: Please pass the branch name as an argument. And generate it in createPullRequestToBase to avoid code duplication

tools/github_helper/main.go
```diff
@@ -93,16 +93,42 @@ func newHelper(r *string) (*helper, error) {
...
+	}
+}
+
 // Creates a pull request from Base branch
 func (h helper) createPullRequestToBase(commit *string) error {
 	if commit == nil {
 		return errors.New("commit cannot be nil.")
 	}
+
+	if err := h.createBranchForFastForward(commit); err != nil {
```
sebastienvas: Move that logging to createBranchForFastForward

tools/github_helper/main.go
```diff
@@ -93,16 +93,42 @@ func newHelper(r *string) (*helper, error) {
...
+		return err
+	}
+
+	newHead := fmt.Sprintf("fastForward-%s", (*commit)[0:7])
+	log.Printf("Created a new branch for fast forward: %s", newHead)
```
sebastienvas: Move that logging to createBranchForFastForward

tools/github_helper/main.go
sebastienvas: We should probably check if the branch already exists and return nil in that case.

sebastienvas (changes requested):
PTAL

tools/github_helper/main.go
yutongz: It turns out we can't use "/" in branch name

tools/github_helper/main.go
```diff
@@ -93,16 +93,50 @@ func newHelper(r *string) (*helper, error) {
 	}
 }
 
+// Create a new branch for fast forward
+func (h helper) createBranchForFastForward(commit *string) (*string, error) {
+	branchName := fmt.Sprintf("fastForward-%s-%s", h.Head, (*commit)[0:7])
+	ref := fmt.Sprintf("refs/heads/%s", branchName)
+
+	if refer, resp, err := h.Client.Git.GetRef(context.TODO(), h.Owner, h.Repo, ref); resp.StatusCode != 404 {
```
sebastienvas: I don t think you need to actually get the ref, you can just look at the reponse from CreateRef.

sebastienvas (approved):
Minor changes.

PR #159, opened 2017-04-06
Add flag to skip cla/codecov failure in fast forward

yutongz:
https://github.com/istio/istio-testing/issues/157

### DIFF
--- a/tools/github_helper/main.go
+++ b/tools/github_helper/main.go
@@ -9,7 +9,7 @@ import (
 	"log"
 	"net/http"
 	"strings"
-
+	"regexp"
 	"github.com/google/go-github/github"
 	"golang.org/x/oauth2"
 )
@@ -21,7 +21,7 @@ var (
 	base        = flag.String("base", "stable", "The base branch used for PR.")
 	head        = flag.String("head", "master", "The head branch used for PR.")
 	pullRequest = flag.Int("pr", 0, "The Pull request to use.")
-	checkToSkip = flag.String("check_to_skip", "", "Lists of check(s) can be skipped, full context separated with comma.")
+	checkToSkip = flag.String("check_to_skip", "", "Lists of check(s) can be skipped, key words separated by comma.")
 	fastForward = flag.Bool("fast_forward", false, "Creates a PR updating Base to Head.")
 	verify      = flag.Bool("verify", false, "Verifies PR on Base and push them if success.")
 	comment     = flag.String("comment", "", "The comment to send to the Pull Request.")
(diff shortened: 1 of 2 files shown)


tools/github_helper/main.go
```diff
@@ -21,6 +21,8 @@ var (
 	base        = flag.String("base", "stable", "The base branch used for PR.")
 	head        = flag.String("head", "master", "The head branch used for PR.")
 	pullRequest = flag.Int("pr", 0, "The Pull request to use.")
+	codecov     = flag.Bool("codecov", false, "Skip code coverage failure")
```
sebastienvas: Please create a unique flag for those that takes in a comma separated list of context to skip.

tools/github_helper/main.go
```diff
@@ -38,12 +40,14 @@ type ghConst struct {
...
-	Base   string
-	Head   string
-	Pr     int
-	Client *github.Client
+	Owner   string
+	Repo    string
+	Base    string
+	Head    string
+	Pr      int
+	Codecov bool
```
sebastienvas: Change this with something like ChecksToSkip

sebastienvas (changes requested):
PLease create a unique flag.

tools/github_helper/main.go
```diff
@@ -284,12 +281,20 @@ func (h helper) updatePullRequest(pr *github.PullRequest, s *github.CombinedStat
...
-	if state == GH.failure && (h.Codecov || h.Cla) {
+	if state == GH.failure && (len(h.CheckToSkip) > 0) {
 		privilege := true
 		for _, status := range s.Statuses {
 			if *status.State != GH.success {
-				if !(h.Codecov && strings.HasPrefix(*status.Context, "codecov")) && !(h.Cla && strings.HasPrefix(*status.Context, "cla")) {
-					//Failed, and not eligible for any privilege
+				skip := false
+				for _, check := range h.CheckToSkip {
+					if *status.Context == check {
```
sebastienvas: Have you tried this ? Should we use some kind of regex instead ?

sebastienvas (commented):
Just a minor comment

tools/github_helper/main.go
yutongz: I tested its logic by creating a fake combinedStatus directly in the updatePullRequest(), working but naive.
And does this pattern work for you:
keyWord := "cla"   //input flag
pattern := fmt.Sprintf("(^|/)%s(/|$)", keyWord)

sebastienvas: what about fmt.Sprintf(".\*%s.\*", keyWord) ?

yutongz: Well, I think it depends on what possible contexts we have. Are we going to have some confused cases. But in this situation, I think we may prefer false negative to false positive. If we can skip one check while the program block it, well, it's easy to manually check and pass it. On contrary, we may miss a really failure.

sebastienvas (dismissed):

sebastienvas:
To plug it to Jenkins, please update the following to take a build parameters for exceptions 
https://github.com/istio/istio-testing/blob/master/src/org/istio/testutils/Utilities.groovy#L99
And update https://testing.istio.io/job/istio-testing/job/verify-stable-prs/

yutongz:
PTAL

sebastienvas (approved):
Awesome

PR #194, opened 2017-04-26
Added a script to clean cache on nodes

sebastienvas:


yutongz (approved):
Just curious, what's the meaning of name changing?

sebastienvas:
some script were using - in their names other were using _, I just renamed them for consistency

PR #208, opened 2017-05-11
Adds repo names in bucket path to avoid collision between repos

sebastienvas:
When creating the 0.1.1 tag I noticed that mixer and manager repo were using the same bucket directory for a given tag which I did not anticipate. Adding the repo to differentiate them.


yutongz (approved):

yutongz:
Nice catch!

PR #214, opened 2017-05-15
Override combine status from github

sebastienvas:


tools/github_helper/main.go
```diff
@@ -309,21 +310,30 @@ func (h helper) updatePullRequest(pr *github.PullRequest, s *github.CombinedStat
...
+	// Not trusting Combined output.
+	// Not counting successes as there could be no required checks.
+	failures := 0
+	pending := 0
+	for _, status := range s.Statuses {
+		if *status.State == gh.error || *status.State == gh.failure {
+			if skipContext(*status.Context) {
+				continue
 			}
+			failures++
```
yutongz, line 321: Can we break out as soon as we find a failure or pending?

sebastienvas: I think it is better if we have a full view of it. I ll had a logging statement to help us debug this.

sebastienvas (commented):
PTAL

yutongz (approved):

PR #250, opened 2017-06-20
remove main()

yutongz:


yutongz:
I am going to make it a library and have program entry in each repos

PR #252, opened 2017-06-21
Metrics fetcher for jenkins and code coverage

yutongz:


toolbox/metrics/BUILD
```diff
@@ -0,0 +1,39 @@
...
+        "@com_github_prometheus_client_golang//prometheus/promhttp:go_default_library",
+        "@com_google_cloud_go//storage:go_default_library",
+        "@org_golang_x_net//context:go_default_library",
+    ],
+)
+
+go_binary(
+    name = "fetcher",
+    gc_linkopts = [
+        "-linkmode",
```
sebastienvas, line 19: Do you know what are those for ?

toolbox/metrics/main.go
```diff
@@ -0,0 +1,352 @@
...
+		"test-infra":  {"presubmit"},
+	}
+
+	codeCovRegister = map[string]bool{
+		"pilot": true,
+		"mixer": true,
+		"auth":  true,
+	}
+
+	codeCovTrackJob = "presubmit"
```
sebastienvas: Dont you want to use postsubmit here ? Or actually fastforward job ? presubmit might not get in.

toolbox/metrics/main.go
```diff
@@ -0,0 +1,352 @@
...
+		client:      &http.Client{},
+	}
+}
+
+func newRepo(name string) *repo {
+	r := &repo{
+		name: name,
+		jobs: make([]*job, 0),
+	}
+	for n := range jobRegister[name] {
```
sebastienvas: what if name is not in the map, would 'nt you get a nil ?

toolbox/metrics/main.go
```diff
@@ -0,0 +1,352 @@
...
+
+	codeCoverage = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "package_code_coverage",
+			Help: "Package code coverage.",
+		},
+		[]string{"package", "repo"},
+	)
+
+	jobRegister = map[string][]string{
```
sebastienvas: Can we create this list dynamically ? Does the API allow job folder listing ?

toolbox/metrics/main.go
```diff
@@ -0,0 +1,352 @@
...
+		"pilot":       {"presubmit", "postsubmit", "e2e-smoketest", "e2e-suite"},
+		"mixer":       {"presubmit", "postsubmit", "e2e-smoketest", "e2e-suite"},
+		"proxy":       {"presubmit", "postsubmit"},
+		"auth":        {"presubmit", "postsubmit"},
+		"mixerclient": {"presubmit"},
+		"istio":       {"presubmit"},
+		"test-infra":  {"presubmit"},
+	}
+
+	codeCovRegister = map[string]bool{
```
sebastienvas: Same comment here, Can we attempt to get code coverage on all the repos and just print 0 on others ?

toolbox/metrics/main.go
```diff
@@ -0,0 +1,352 @@
...
+	return nil
+}
+
+func (j *job) getLatestCompletedBuild() (int, error) {
+	o, err := j.getJenkinsObject("lastCompletedBuild")
+	if err != nil {
+		return 0, fmt.Errorf("failed to get last build info of build, %v", err)
+	}
+	if len(o) == 0 {
+		return 0, errors.New("get empty result from jenkins")
```
sebastienvas: got empty results ...

toolbox/metrics/main.go
yutongz: yes, we get a nil if the key doesn't exist, and when n == nil, it won't go into the loop

sebastienvas (changes requested):
PTAL

toolbox/metrics/main.go
yutongz: let me use postsubmit for now.

toolbox/metrics/BUILD
yutongz: static build is required for building docker image

yutongz:
PTAL

toolbox/metrics/main.go
```diff
@@ -0,0 +1,400 @@
...
+		prometheus.GaugeOpts{
+			Name: "package_code_coverage",
+			Help: "Package code coverage.",
+		},
+		[]string{"package", "repo"},
+	)
+
+	repoRegister = map[string]*repo{}
+
+	codeCovTrackJob = "postsubmit"
```
sebastienvas: I would make this configurable.

toolbox/metrics/main.go
```diff
@@ -0,0 +1,400 @@
...
+const (
+	fetchInterval = 5
+)
+
+var (
+	webAddress = flag.String("listen_port", ":9103", "port on which to expose metrics and web interface.")
+	jenkinsURL = flag.String("jenkins_url", "https://testing.istio.io", "URL to Jenkins API")
+	gcsBucket  = flag.String("bucket", "istio-code-coverage", "gcs bucket")
+
+	succeededBuilds = prometheus.NewSummaryVec(
```
sebastienvas: I feel this should go into a struct

toolbox/metrics/main.go
```diff
@@ -0,0 +1,400 @@
...
+	)
+
+	repoRegister = map[string]*repo{}
+
+	codeCovTrackJob = "postsubmit"
+
+	gcsClient  *storage.Client
+	httpClient = &http.Client{}
+
+	jk = &jkConst{
```
sebastienvas, line 56: Can you just declare those as const ?

yutongz: Honestly, I prefer this way. In my point of view, these strings are from JSON files of Jenkins, so better put them into a structure.

yutongz:
PTAL

toolbox/metrics/main.go
```diff
@@ -0,0 +1,422 @@
...
+		name:               "name",
+		jobs:               "jobs",
+		building:           "building",
+		result:             "result",
+		duration:           "duration",
+		number:             "number",
+		resultFailure:      "FAILURE",
+		resultSUCCESS:      "SUCCESS",
```
sebastienvas: rename to resultSuccess for consistency.

toolbox/metrics/main.go
sebastienvas: Golang does not support const struct which makes it ugly to declare a struct to put in const in.

sebastienvas (approved):

PR #253, opened 2017-06-30
prow: introduce prow

nlandolfi:
Introduce prow related resources and configurations.

1. Introduce the cluster deployment and config files for Istio
2. Introduce docker/prowbazel image; a modified version of the image used on Jenkins, suitable for prow jobs
3. Introduce test-infra-presubmit job to this repository.

Update 7/6/17: prow is running, view the simple deck frontend [here](http://35.190.0.30/). It is running as the `prow` cluster on istio-testing, so with access to the project one can also get the kubectl context and get the pods:
```bash
$ kubectl get pods
 NAME                     READY     STATUS    RESTARTS   AGE
deck-3409711596-8qs1f    1/1       Running   0          8m
deck-3409711596-jq191    1/1       Running   0          8m
deck-3409711596-xjdw3    1/1       Running   0          8m
hook-4091801138-bg7cr    1/1       Running   0          8m
plank-1139864133-2d0g3   1/1       Running   0          8m
sinker-79557021-1d9xm    1/1       Running   0          8m
tot-3401709015-lkwgm     1/1       Running   0          8m
```
This PR adds the test-infra-presubmit job file. This results of test runs from this file are [here](https://k8s-gubernator.appspot.com/builds/istio-prow/pull/istio_test-infra/253/test-infra-presubmit/). Notice that we are piggy-backing on kubernetes instance of `gubernator`, a UI for viewing job results.

I have turned off public reporting for all prow jobs configured in `prow/cluster/configmaps/config.yaml` for the moment, because it was needlessly failing on other PRs (c.f. [254](https://github.com/istio/test-infra/pull/254). 

My intention is to create a pull request against pilot adding its job files in a similar manner. I will wait to see that those are passing (while keeping reporting off), then once that merges master we can turn on the public reporting. As we can turn on public reporting once this PR goes to master (because the jobs/test-infra-presubmit.sh file will be in master).

TODO:
- [x] Changes to bootstrap.py our k8s-test-infra fork
- [x] Look into modifying the checkout directory in bootstrap.py as per [Seb's comment](https://github.com/istio/test-infra/pull/253#discussion_r126207204)

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,7 @@
+This directory contains standard images maintained by Istio for test infrastructure purposes.
+
+* prowbazel - gives an environment with git, gcloud, python, bazel, and go.
+    It is suitable for bazel-based projects. Currently, it is analagous to the
+    Jenkins slave image, and therefore can be used by all jobs for testing
+    istio repositories. In the future, specific jobs may have specific images.
+* bazel-remote & slaves - provide images for Jenkins-based testing.
--- a/docker/prowbazel/Dockerfile
+++ b/docker/prowbazel/Dockerfile
@@ -1,0 +1,50 @@
+FROM ubuntu:xenial
+
+# Note: The build directory is the root of the istio/test-infra repo, not ./
+
+# Installing necessary packages
+RUN rm -rf /var/lib/apt/lists/* \
+    && apt-get update --fix-missing -qq \
+    && apt-get install -qqy git iptables procps sudo xz-utils \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Adding sudo group user no password access.
+# This is used by bootstrap user to start docker service
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+
+# Installing
+ADD scripts /tmp/istio_tmp/scripts
+RUN chmod +x /tmp/istio_tmp/scripts/linux-install-software
+RUN /tmp/istio_tmp/scripts/linux-install-software \
+    && rm -rf /tmp/istio_tmp
+
+ENV PATH /usr/local/go/bin:/usr/lib/google-cloud-sdk/bin:${PATH}
+
+# Create bootstrap user
+ENV HOME /home/bootstrap
+RUN useradd -c "Bootstrap user" -d ${HOME} -G docker,sudo -m bootstrap -s /bin/bash
+USER bootstrap
+WORKDIR ${HOME}
+
+# Explicitly set bazel cache, to avoid cache issues with checking out
+# code and running bazel in different dir than the calling script.
+ENV TEST_TMPDIR /home/bootstrap/.cache/bazel
+
+# Add bootstrap, the test harness (checks out code, captures log and status)
+ADD https://raw.githubusercontent.com/nlandolfi/test-infra-1/d7900388fa1fb69738e231a21230373fd4b5ede9/jenkins/bootstrap.py ${HOME}
+RUN sudo chmod +rx ${HOME}/bootstrap.py
+
+# Add entrypoint that runs bootstrap with appropriate arguments
+ADD ./docker/prowbazel/entrypoint /usr/local/bin/entrypoint
+RUN sudo chmod +rx /usr/local/bin/entrypoint
+
+# Final GOPATH configuration
+RUN mkdir -p /home/bootstrap/go/src
+ENV GOPATH=/home/bootstrap/go
+ENV PATH ${GOPATH}/bin:${PATH}
+
+# Set CI variable which can be checked by test scripts to verify
+# if running in the continuous integration environment.
+ENV CI bootstrap
+
+ENTRYPOINT ["entrypoint"]
--- a/docker/prowbazel/Makefile
+++ b/docker/prowbazel/Makefile
@@ -1,0 +1,17 @@
+PROJECT = istio-testing
+VERSION = 0.1.5
+
+# Note: The build directory is the root of the istio/test-infra repository, not ./
+image:
+	docker build -t "gcr.io/$(PROJECT)/prowbazel:$(VERSION)" -f Dockerfile ../../
+
+push:
+	gcloud docker -- push "gcr.io/$(PROJECT)/prowbazel:$(VERSION)"
+
+# Open shell to running prowbazel image for experimentation.
+run:
+	docker run -it --entrypoint bash --privileged gcr.io/$(PROJECT)/prowbazel:$(VERSION)
+# Example of running a job:
+# ~/bootstrap.py --repo="github.com/<org>/<repo>" --job="<job-name>" --clean --root="${GOPATH}/src" --pull=master:<sha>,<pr-#>:<sha>
+
+.PHONY: image push
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -1,0 +1,76 @@
+# Copyright Istio Authors 2017.
+#
+# Based off a similar file used to manage the k8s cluster
+# https://github.com/kubernetes/test-infra/blob/master/prow/Makefile
+
+# GKE variables.
+PROJECT ?= istio-testing
+ZONE    ?= us-west1-a
+CLUSTER ?= prow
+
+# ensures that kubectl has access to the prow cluster's config context
+get-cluster-credentials:
+	gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
+
+update-config: get-cluster-credentials
+	kubectl create configmap config --from-file=config=./config.yaml --dry-run -o yaml | kubectl replace configmap config -f -
+
+update-plugins: get-cluster-credentials
+	kubectl create configmap plugins --from-file=plugins=./plugins.yaml --dry-run -o yaml | kubectl replace configmap plugins -f -
+
+# use to guarantee that on command completion all deployments see new configs
+update-configs:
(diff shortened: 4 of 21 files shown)


kyessenov (commented):
I hope you don't expect us to review 600k lines of code :)

nlandolfi:
😂 certainly not -- advice on the right way to vendor another Git repo that
we may want to modify in the future welcome...currently approaching with
git subtrees ....

nlandolfi:
for now axing k8s-test-infra subtree because I don't modify here, and we use the k8s-prow images now (rather than build our own from source).

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

nlandolfi:
@istio-testing bazel test this

docker/README.md
```diff
@@ -0,0 +1,4 @@
+This directory contains standard images maintained by Istio for test infrastructure purposes.
+
+* prowbazel - gives an environment with git, gcloud, python, bazel, and go. It is suitable for bazel-based projects.
```
rshriram: is this deeply tied to gcloud? Can this be run on other platforms?

rshriram: The reason I ask is because eventually folks would want to run tests against istio/istio on their own environments. Having a jenkins and a prow seems ripe for all sorts of issues such as test environment divergence.

docker/prowbazel/bootstrap.py
```diff
@@ -0,0 +1,965 @@
+#!/usr/bin/env python
+
+# Copyright 2016 The Kubernetes Authors.
```
ldemailly: any way to reference rather than copy this large file ? or did you make changes to it (which change compared to the original ? where is the original?)

ldemailly (commented):
it's exciting to see this taking shape, nice work!

quick random question below:

prow/cluster/configmaps/config.yaml
```diff
@@ -150,6 +150,7 @@ presubmits:
 
   istio/test-infra:
   - name: test-infra-presubmit
+    skip_report: true
```
ldemailly: why ? I'm curious to see how prow reports work and I hope it doesn't have the bug that our jenkins report has of posting a comment that triggers a lot of useless github notifications ?

docker/README.md
sebastienvas: The prowbazel docker image will be used in every job. We are actually uploading artifacts and logs to google cloud storage. Those would be publicly available. If folks want to upload to other cloud, they can modify this image to add new tools or create their own image. In a near future we ll probably use specific image for specific jobs.

docker/prowbazel/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN useradd -c "Bootstrap user" -d ${HOME} -G docker,sudo -m bootstrap -s /bin/bash
+USER bootstrap
+WORKDIR ${HOME}
+
+# Explicitly set bazel cache, to avoid cache issues with checking out
+# code and running bazel in different dir than the calling script.
+ENV TEST_TMPDIR /home/bootstrap/.cache/bazel
+
+# Add bootstrap, the test harness (checks out code, captures log and status)
+ADD ./docker/prowbazel/bootstrap.py ${HOME}
```
sebastienvas: Based on our discussion with @ldemailly, this is where you would put an URL. https://docs.docker.com/engine/reference/builder/#add. The url might something like https://raw.githubusercontent.com/kubernetes/test-infra/master/jenkins/bootstrap.py

docker/prowbazel/Makefile
```diff
@@ -0,0 +1,17 @@
...
+# Note: The build directory is the root of the istio/test-infra repository, not ./
+image:
+	docker build -t "gcr.io/$(PROJECT)/prowbazel:$(VERSION)" -f Dockerfile ../../
+
+push:
+	gcloud docker -- push "gcr.io/$(PROJECT)/prowbazel:$(VERSION)"
+
+# Open shell to running prowbazel image for experimentation.
+run:
+	docker run -it --entrypoint bash --privileged gcr.io/istio-testing/prowbazel:0.1
```
sebastienvas: replace istio-testing with $(PROJECT) and 0.1 with $(VERSION)

jobs/test-infra-presubmit.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -ex
+
+if [ "$CI" == "bootstrap" ]; then
+    # ensure correct path
+    mkdir -p $GOPATH/src/istio.io
```
sebastienvas: It kind of sucks that we need to do that on every job. Could we move that feature into bootstrap once we have our own fork ?

prow/README.md
```diff
@@ -0,0 +1,47 @@
...
+In the repository of interest, add a `/jobs/` directory. The test harness will look for `/jobs/<job-name>.sh` and execute this script.
+
+For example, a configuration that referenced the jobs `my-presubmit`, `my-postsubmit` and `my-job` would require a corresponding directory structure:
+
+```bash
+$ tree
+.
+├── anotherdir
+│   └── ...
+├── jobs
```
sebastienvas: let s rename this to prow

prow/cluster/configmaps/config.yaml
```diff
@@ -0,0 +1,102 @@
+---
```
sebastienvas: let s move this to prow/config.yaml as discussed.

prow/cluster/configmaps/plugins.yaml
```diff
@@ -0,0 +1,9 @@
+# Plugin repository whitelist.
```
sebastienvas: let move this to prow/plugins.yaml as discussed.

sebastienvas (changes requested):
PTAL

prow/Makefile
```diff
@@ -0,0 +1,76 @@
...
+	make update-plugins
+	make startd
+
+.PHONY: update-config update-plugins get-cluster-credentials update-configs
+
+deck-deployment: get-cluster-credentials
+	kubectl apply -f cluster/deck_deployment.yaml
+
+deck-service: get-cluster-credentials
+	kubectl create -f cluster/deck_service.yaml
```
kyessenov: why not make everything apply?

prow/cluster/prow_job.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: extensions/v1beta1
+kind: ThirdPartyResource
```
kyessenov, line 2: should probably switch to `CustomResourceDefinition` since TPRs are deprecated. I'd expect the upstream project to do the work fairly soon.

kyessenov (approved):

docker/README.md
nclando: good point @rshriram -- the prowbazel docker file in particular, not tied to GCP, just a docker file (we do host it on GCR). 

to reiterate as @sebastienvas mentioned, the entrypoint of the docker file uploads logs to a gcs bucket, which is google specific. the bucket is read public.

other folks running infrastructure would likely use an image that uploads logs to their preferred storage.

prow/cluster/configmaps/config.yaml
nclando: This diff introduces the `prow/test-infra-presubmit.sh`, but other PRs on these repos don't yet have the file, and the job would fail, therefore I turned off reporting.

re: notifications, Prow does comment on the PR for test failures.

nlandolfi:
@istio-testing bazel test this

prow/Makefile
nclando: good call. done

docker/prowbazel/bootstrap.py
nclando: We have forked the kubernetest/test-infra repo [here](https://github.com/istio/k8s-test-infra). I have a PR against our fork [here](https://github.com/istio/k8s-test-infra/pull/1). I cleaned up my changes so that they are flags which, when unspecified, do not modify current behavior, but also allow us to modify/turn off the things we need to.

docker/prowbazel/Makefile
nclando: good catch, done.

prow/README.md
nclando: done

prow/cluster/prow_job.yaml
nclando: nice catch. I created an[ issue upstream](https://github.com/kubernetes/test-infra/issues/3400), I assume the plan is to make the switch.

nlandolfi:
@istio-testing bazel test this

sebastienvas (approved):

nlandolfi:
@istio-testing bazel test this

nlandolfi:
the prow results for the presubmit [are here](https://k8s-gubernator.appspot.com/builds/istio-prow/pull/istio_test-infra/253/test-infra-presubmit/).

@sebastienvas

PR #254, opened 2017-07-06
Create tmp dir for go test

yutongz:
Didn't create tmp dir so that it failed to create test files


toolbox/pkg_check/main_test.go
```diff
@@ -149,10 +149,21 @@ func TestFailedCheck(t *testing.T) {
 }
 
 func TestMain(m *testing.M) {
+	if _, err := os.Stat(tmpDir); os.IsNotExist(err) {
```
sebastienvas: You might want to use https://golang.org/pkg/io/ioutil/#TempFile to make your test more predictable. As of now if I run tests concurrently they will fail. You want to make sure that each file name is unique.

sebastienvas (changes requested):
PTAL

yutongz:
PTAL

sebastienvas (approved):

toolbox/pkg_check/main_test.go
```diff
@@ -149,10 +153,17 @@ func TestFailedCheck(t *testing.T) {
...
-	os.Exit(m.Run())
+	var err error
+	if tmpDir, err = ioutil.TempDir("", tmpDirPrefix); err != nil {
+		log.Printf("Failed to create tmp directory: %s, %s", tmpDir, err)
+		os.Exit(4)
+	}
+
+	exitCode := m.Run()
+
+	if err := os.RemoveAll(tmpDir); err != nil {
```
nclando, line 164: why move this out of a defer?

yutongz: cause os.Exit() will end it immediately without going through defer.

PR #263, opened 2017-07-10
Adds istio/istio presubmit

sebastienvas:


nlandolfi (changes requested):
add trigger to plugins.yaml

prow/plugins.yaml
```diff
@@ -2,6 +2,10 @@
 # Keys: Full repo name: "org/repo".
 # Values: List of plugins to run against the repo.
 ---
+istio/istio:
+- trigger
+- config-updater
```
nclando: I don't think you need this, only on the config the test-infra repo

nlandolfi (approved):
LGTM

PR #268, opened 2017-07-10
update README with triggers/ block and execution bit gotchas

nlandolfi:
Add two gotchas to the README. 

1. The first is that in order for triggers to work, you must also add them to the config block at the top of `config.yaml`.

2. Second, in order for the prowbazel image to execute the job script, the execution bit must be set. Otherwise you will recieve a Permission error, as [here](https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_istio/464/istio-presubmit/1/). This is a bit of a gotcha, so we call it out in the README.

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -27,10 +27,15 @@ Both of these types of jobs can be configured using the config configmap [here](
 
 ```yaml
 # in config.yaml
+triggers:
+- repos:
+  - istio/istio
+  - istio/test-infra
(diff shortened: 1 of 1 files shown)


yutongz (approved):
Nice!

nlandolfi:
@sebastienvas PTAL

sebastienvas (approved):

PR #271, opened 2017-07-11
Add shared kubeconfig secret to istio presubmit

sebastienvas:


yutongz (approved):

nlandolfi (approved):
LGTM, this will turn on istio/istio reporting which is cool now because the other PR I just reviewed showed that it works?

But I still think we *shouldn't* merge this until that change goes into istio/istio?

nlandolfi:
Nice!

PR #274, opened 2017-07-12
Skip broken builds

yutongz:
When hit a build whose message is broken for some reasons, skip it and move on. This happens when we manually delete a build at Jenkins UI.


sebastienvas (approved):

yutongz:
@istio-testing test this

PR #298, opened 2017-07-20
prow: add pilot e2e smoke test config

nlandolfi:


prow/config.yaml
```diff
@@ -116,6 +116,56 @@ presubmits:
...
+      rerun_command: "@istio-testing smoke test this"
+      trigger: "((?m)^@istio-testing (smoke )?test this,?(\\s+|$)|(?m)^/test( all| bazel),?(\\s+|$))"
+      skip_report: true
+      branches:
+      - master
+      spec:
+        containers:
+        - image: gcr.io/istio-testing/prowbazel:0.1.5
+          args:
+          - "--pull=$(PULL_REFS)"
```
sebastienvas, line 130: I think you want to do "--repo=istio/test-infra=master" instead of  "--pull=$(PULL_REFS)" since you are checking out istio master to run the test from.

prow/config.yaml
```diff
@@ -116,6 +116,56 @@ presubmits:
...
+          - containerPort: 9999
+            hostPort: 9998
+        volumes:
+        - name: service
+          secret:
+            secretName: service-account
+        - name: e2e-testing-kubeconfig
+          secret:
+            secretName: e2e-testing-kubeconfig
+        - name: pilot-codecov-token
```
sebastienvas: You don't need codecov for this job.

prow/config.yaml
```diff
@@ -116,6 +116,56 @@ presubmits:
...
+          - name: GOOGLE_APPLICATION_CREDENTIALS
+            value: /etc/service-account/service-account.json
+          volumeMounts:
+          - name: service
+            mountPath: /etc/service-account
+            readOnly: true
+          - name: e2e-testing-kubeconfig
+            mountPath: /home/bootstrap/.kube
+            readOnly: true
+          - name: pilot-codecov-token
```
sebastienvas: No codecov for this job.

prow/config.yaml
```diff
@@ -116,6 +116,56 @@ presubmits:
...
         hostPath:
           path: /mnt/disks/ssd0
+    run_after_success:
+    - name: pilot-e2e-smoketest
+      context: prow/pilot-e2e-smoketest.sh
+      always_run: true
+      rerun_command: "@istio-testing smoke test this"
+      trigger: "((?m)^@istio-testing (smoke )?test this,?(\\s+|$)|(?m)^/test( all| bazel),?(\\s+|$))"
+      skip_report: true
+      branches:
```
sebastienvas: You can probably remove branches from here right ?

sebastienvas (changes requested):

prow/config.yaml
nclando: I clone the istio/istio repo in the script. If this job checked out the istio repo then we wouldn't have a local script for running this job in the pilot repository.

thoughts?

prow/config.yaml
nclando: Yes, I believe so.

nlandolfi:
@sebastienvas PTAL

prow/config.yaml
sebastienvas: but the script would be in istio/istio right ?

sebastienvas (changes requested):
PTAL

nlandolfi:
Ok, so to record what we discussed.

It makes sense to put the prow file for this job in the repo whose presubmit it blocks. In this case, we put it in pilot. Rationale includes: 
1. being able to run the job locally
2. editing the job file can be done in that repo and and affect that repo's presubmit during editing (i.e., to add debugging lines or to actually change the presubmit, the effects can be observed on the PR) (lose this ability if editing the job on a different repo than the one the presubmit blocks)
3. logically, all jobs for a particular repo are in that repo

Trade-off is that if we make a change to the actualy e2e tests which live in istio/istio, we will need to change files across several repos independently (pilot, mixer, and auth).

If we feel good about this, let's commit this config change so that the job file introduced [here](https://github.com/istio/pilot/pull/946) can be tested.

cc @sebastienvas

sebastienvas (approved):

PR #306, opened 2017-07-24
Enable auth presubmit in Prow

sebastienvas:


prow/config.yaml
```diff
@@ -25,6 +26,59 @@ presubmits:
...
   #                  Otherwise, run a Jenkins job.
 
+  istio/auth:
+  - name: auth-presubmit
+    context: prow/auth-presubmit.sh
+    skip_report: true
+    always_run: true
+    rerun_command: "@istio-testing bazel test this"
+    trigger: "((?m)^@istio-testing (bazel )?test this,?(\\s+|$)|(?m)^/test( all| bazel),?(\\s+|$))"
+    branches:
```
nclando, line 36: I think we have decided that we don't need this?

nlandolfi (approved):
LGTM

prow/config.yaml
sebastienvas: I prefer stating which branch this is testing for. We only care about master and release branches and those may have different test requirements.

PR #314, opened 2017-07-25
Mungegithub config files for istio/test-infra

yutongz:


yutongz:
@sebastienvas   when approve, try use "/approve" and you can skip normal github-way approve. Lucky it will work just fine.

yutongz:
@sebastienvas  since this pr is not associated with any issue, we kind of have to use "/approve no-issue" :)

PR #319, opened 2017-07-25
Add Nick as a prow approver

yutongz:


prow/OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
+  - nlandolfi 
+
```
nclando, line 3: do we need this extra new line?

nlandolfi (approved):
LGTM, maybe kill the extra new line if not needed.

prow/OWNERS
yutongz: DONE!

PR #336, opened 2017-07-27
Automated dependency update binary

chxchx:
**Release note**:

```release-note
NONE
```

### DIFF
--- a/toolbox/deps_update/dependency.go
+++ b/toolbox/deps_update/dependency.go
@@ -59,7 +59,7 @@ func deserializeDeps(depsFilePath string) ([]dependency, error) {
 
 // Writes in-memory dependencies to file
 func serializeDeps(depsFilePath string, deps *[]dependency) error {
-	pickled, err := json.Marshal(*deps)
+	pickled, err := json.MarshalIndent(*deps, "", "\t")
 	if err != nil {
 		return err
 	}
--- a/toolbox/deps_update/dependency.go
+++ b/toolbox/deps_update/dependency.go
@@ -14,6 +14,11 @@
 
 package main
 
+import (
+	"encoding/json"
+	"io/ioutil"
+)
+
 /*
 	Assumes Dependency Hoisting in all bazel dependency files. Example:
 
@@ -38,4 +43,15 @@ type dependency struct {
 	RepoName   string `json:"repoName"`
 	ProdBranch string `json:"prodBranch"` // either master or stable
 	File       string `json:"file"`       // where in the *parent* repo such dependecy is recorded
+}
+
+// Get the list of dependencies of a repo
+func getDeps(depsFilePath string) ([]dependency, error) {
+	var deps []dependency
+	raw, err := ioutil.ReadFile(depsFilePath)
+	if err != nil {
+		return deps, err
+	}
+	err = json.Unmarshal(raw, &deps)
+	return deps, err
 }
--- a/toolbox/deps_update/dependency.go
+++ b/toolbox/deps_update/dependency.go
@@ -1,0 +1,45 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+/*
+	Assumes Dependency Hoisting in all bazel dependency files. Example:
+
+		new_git_repository(
+			name = "mixerapi_git",
+			commit = "ee9769f5b3304d9e01cd7ed6fb1dbb9b08e96210",
+			remote = "https://github.com/istio/api.git",
+		)
+
+	becomes
+
+		MIXERAPI = "ee9769f5b3304d9e01cd7ed6fb1dbb9b08e96210"
+		new_git_repository(
+			name = "mixerapi_git",
+			commit = MIXERAPI,
+			remote = "https://github.com/istio/api.git",
+		)
+*/
+
+type dependency struct {
+	name       string
+	repoName   string
+	prodBranch string // either master or stable
+	file       string // where in the *parent* repo such dependecy is recorded
+}
+
+var (
+	deps = make(map[string][]dependency)
+)
--- a/toolbox/deps_update/dependency.go
+++ b/toolbox/deps_update/dependency.go
@@ -34,12 +34,8 @@ package main
 */
 
 type dependency struct {
-	name       string
-	repoName   string
-	prodBranch string // either master or stable
-	file       string // where in the *parent* repo such dependecy is recorded
+	Name       string `json:"name"`
+	RepoName   string `json:"repoName"`
+	ProdBranch string `json:"prodBranch"` // either master or stable
+	File       string `json:"file"`       // where in the *parent* repo such dependecy is recorded
 }
-
-var (
-	deps = make(map[string][]dependency)
-)
--- a/toolbox/deps_update/dependency.go
+++ b/toolbox/deps_update/dependency.go
@@ -39,14 +39,15 @@ import (
 */
 
 type dependency struct {
-	Name       string `json:"name"`
-	RepoName   string `json:"repoName"`
-	ProdBranch string `json:"prodBranch"` // either master or stable
-	File       string `json:"file"`       // where in the *parent* repo such dependecy is recorded
+	Name          string `json:"name"`
+	RepoName      string `json:"repoName"`
+	ProdBranch    string `json:"prodBranch"`    // either master or stable
+	File          string `json:"file"`          // where in the *parent* repo such dependecy is recorded
+	LastStableSHA string `json:"lastStableSHA"` // sha used in the latest stable build of parent
 }
 
-// Get the list of dependencies of a repo
-func getDeps(depsFilePath string) ([]dependency, error) {
+// Get the list of dependencies of a repo by deserializing the file on depsFilePath
+func deserializeDeps(depsFilePath string) ([]dependency, error) {
 	var deps []dependency
 	raw, err := ioutil.ReadFile(depsFilePath)
 	if err != nil {
@@ -54,4 +55,13 @@ func getDeps(depsFilePath string) ([]dependency, error) {
 	}
 	err = json.Unmarshal(raw, &deps)
 	return deps, err
+}
+
+// Writes in-memory dependencies to file
+func serializeDeps(depsFilePath string, deps *[]dependency) error {
+	pickled, err := json.Marshal(*deps)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(depsFilePath, pickled, 0600)
 }
--- a/toolbox/deps_update/githubClient.go
+++ b/toolbox/deps_update/githubClient.go
@@ -1,0 +1,111 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"strings"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+)
+
+var (
+	tokenFile = flag.String("token_file", "", "File containing Auth Token.")
+	owner     = flag.String("owner", "istio-testing", "Github Owner or org.")
+	token     = ""
+)
+
+type githubClient struct {
+	client *github.Client
+}
+
+// Get github api token from tokenFile
+func getToken() (string, error) {
+	if token != "" {
+		return token, nil
+	}
+	if *tokenFile == "" {
+		return "", fmt.Errorf("token_file not provided")
+	}
+	b, err := ioutil.ReadFile(*tokenFile)
+	if err != nil {
+		return "", err
+	}
+	token = strings.TrimSpace(string(b[:]))
+	return token, nil
+}
+
+func newGithubClient() (*githubClient, error) {
+	token, err := getToken()
+	if err != nil {
+		return nil, err
+	}
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(context.Background(), ts)
+	if err != nil {
+		return nil, err
+	}
+	client := github.NewClient(tc)
+	return &githubClient{client}, nil
+}
+
+func (g githubClient) createPullRequest(branch, repo string) error {
+	if branch == "" {
+		return errors.New("branch cannot be empty")
+	}
+	title := fmt.Sprintf("[DO NOT MERGE] Auto PR to update dependencies of %s", repo)
+	body := "This PR will be merged automatically once checks are successful."
+	base := "master"
+	req := github.NewPullRequest{
+		Head:  &branch,
+		Base:  &base,
+		Title: &title,
+		Body:  &body,
+	}
+	log.Printf("Creating a PR with Title: \"%s\" for repo %s", title, repo)
+	pr, _, err := g.client.PullRequests.Create(context.TODO(), *owner, repo, &req)
+	if err != nil {
+		return err
+	}
+	log.Printf("Created new PR at %s", *pr.HTMLURL)
+	return nil
+}
+
+func (g githubClient) getListRepos() ([]string, error) {
+	opt := &github.RepositoryListOptions{Type: "owner"}
+	repos, _, err := g.client.Repositories.List(context.Background(), *owner, opt)
+	if err != nil {
+		return nil, err
+	}
+	var listRepoNames []string
+	for _, r := range repos {
+		listRepoNames = append(listRepoNames, *r.Name)
+	}
+	return listRepoNames, nil
+}
+
+func (g githubClient) getHeadCommitSHA(repo, branch string) (string, error) {
+	ref, _, err := g.client.Git.GetRef(context.Background(), *owner, repo, "refs/heads/"+branch)
+	if err != nil {
+		return "", err
+	}
+	return *ref.Object.SHA, nil
+}
--- a/toolbox/deps_update/githubClient.go
+++ b/toolbox/deps_update/githubClient.go
@@ -25,12 +25,15 @@ import (
 
 	"github.com/google/go-github/github"
 	"golang.org/x/oauth2"
+
+	"istio.io/test-infra/toolbox/util"
 )
 
 var (
 	tokenFile = flag.String("token_file", "", "File containing Auth Token.")
 	owner     = flag.String("owner", "istio-testing", "Github Owner or org.")
 	token     = ""
+	ci        = util.NewCIState()
 )
 
 type githubClient struct {
@@ -100,6 +103,31 @@ func (g githubClient) getListRepos() ([]string, error) {
 		listRepoNames = append(listRepoNames, *r.Name)
 	}
 	return listRepoNames, nil
+}
+
+func (g githubClient) getListBranches(repo string) ([]string, error) {
+	branches, _, err := g.client.Repositories.ListBranches(
+		context.Background(), *owner, repo, nil)
+	if err != nil {
+		return nil, err
+	}
+	var branchNames []string
+	for _, b := range branches {
+		branchNames = append(branchNames, b.GetName())
+	}
+	return branchNames, nil
+}
+
+func (g githubClient) hasFailedAnyCICheck(repo, branch string) (bool, error) {
+	// TODO (chx) list pr, use pr commit sha to get combined status
+	// TODO (chx) test with istio token
+	combinedStatus, _, err := g.client.Repositories.GetCombinedStatus(
+		context.Background(), *owner, repo, branch, nil)
+	if err != nil {
+		return false, err
+	}
+	finalState := util.GetCIState(combinedStatus, nil)
+	return (finalState == ci.Failure), nil
 }
 
 func (g githubClient) getHeadCommitSHA(repo, branch string) (string, error) {
--- a/toolbox/deps_update/githubClient.go
+++ b/toolbox/deps_update/githubClient.go
@@ -19,8 +19,10 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/google/go-github/github"
+	multierror "github.com/hashicorp/go-multierror"
 	"golang.org/x/oauth2"
 
 	"istio.io/test-infra/toolbox/util"
@@ -28,6 +30,10 @@ import (
 
 var (
 	ci = util.NewCIState()
+)
+
+const (
+	prTitlePrefix = "[DO NOT MERGE] Auto PR to update dependencies of "
 )
 
 type githubClient struct {
@@ -52,11 +58,12 @@ func (g githubClient) remote(repo string) string {
 	)
 }
 
+// Create a pull request within repo from branch to master
 func (g githubClient) createPullRequest(branch, repo string) error {
 	if branch == "" {
 		return errors.New("branch cannot be empty")
 	}
-	title := fmt.Sprintf("[DO NOT MERGE] Auto PR to update dependencies of %s", repo)
+	title := prTitlePrefix + repo
 	body := "This PR will be merged automatically once checks are successful."
 	base := "master"
 	req := github.NewPullRequest{
@@ -66,7 +73,8 @@ func (g githubClient) createPullRequest(branch, repo string) error {
 		Body:  &body,
 	}
 	log.Printf("Creating a PR with Title: \"%s\" for repo %s", title, repo)
-	pr, _, err := g.client.PullRequests.Create(context.Background(), g.owner, repo, &req)
+	pr, _, err := g.client.PullRequests.Create(
+		context.Background(), g.owner, repo, &req)
 	if err != nil {
 		return err
 	}
@@ -74,6 +82,7 @@ func (g githubClient) createPullRequest(branch, repo string) error {
 	return nil
 }
 
+// Returns a list of repos under the provided owner
 func (g githubClient) listRepos() ([]string, error) {
 	opt := &github.RepositoryListOptions{Type: "owner"}
 	repos, _, err := g.client.Repositories.List(context.Background(), g.owner, opt)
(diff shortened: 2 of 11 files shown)


chxchx:
The base of this code change was migrated from [PR#493 on istio/istio](https://github.com/istio/istio/pull/493)

Additionally, we assume each repo now hosts a file where its dependency info is recorded (see [PR #89 on istio/mixerclient](https://github.com/istio/mixerclient/pull/89)). Hence, the dependency update script now accesses dependency info from such file.

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Assumes at the root of istio directory
+// Runs the updateVersion.sh script
+func updateIstioDeps() error {
+	pilotSHA, err := githubClnt.getHeadCommitSHA("pilot", "stable")
+	if err != nil {
+		return err
+	}
+	mixerSHA, err := githubClnt.getHeadCommitSHA("mixer", "stable")
```
sebastienvas: What about auth for istio-ca ?

toolbox/deps_update/dependency.go
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+/*
+	Assumes Dependency Hoisting in all bazel dependency files. Example:
```
nclando: s/Hoisting/Hosting

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2017 Istio Authors
```
nclando, line 1: I think it might be a go thing to name files snake case `github_client.go`

but not a  big deal i don't think.

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,111 @@
...
+	body := "This PR will be merged automatically once checks are successful."
+	base := "master"
+	req := github.NewPullRequest{
+		Head:  &branch,
+		Base:  &base,
+		Title: &title,
+		Body:  &body,
+	}
+	log.Printf("Creating a PR with Title: \"%s\" for repo %s", title, repo)
+	pr, _, err := g.client.PullRequests.Create(context.TODO(), *owner, repo, &req)
```
nclando: i think either parameterize this function by a context or use context.Background() -- not TODO

toolbox/deps_update/util.go
```diff
@@ -0,0 +1,44 @@
...
+	"strings"
+)
+
+// GetMD5Hash generates an MD5 digest of the given string
+func GetMD5Hash(text string) string {
+	hash := md5.Sum([]byte(text))
+	return hex.EncodeToString(hash[:])
+}
+
+// Shell run command on shell and get back output and error if get one
```
nclando: nit: s/run/runs

toolbox/deps_update/util.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"log"
+	"os/exec"
+	"strings"
+)
+
+// GetMD5Hash generates an MD5 digest of the given string
+func GetMD5Hash(text string) string {
+	hash := md5.Sum([]byte(text))
+	return hex.EncodeToString(hash[:])
```
nclando: is the [:] making a copy of the hash bytes and do we need to make a copy since you don't do anything with hash?

toolbox/deps_update/util.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Shell run command on shell and get back output and error if get one
+func Shell(format string, args ...interface{}) (string, error) {
+	command := fmt.Sprintf(format, args...)
+	parts := strings.Split(command, " ")
+	log.Printf("Running command %s", command)
+	c := exec.Command(parts[0], parts[1:]...) // #nosec
```
nclando: question: what does nosec mean?

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if err := os.Chdir(".."); err != nil {
+		return err
+	}
+	if _, err := Shell("rm -rf " + repo); err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
```
nclando, line 200: yeah so another option is for handling the contexts is to have ctx := context.Background() here and then thread that through all the network calls.

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+	cmd := fmt.Sprintf("./install/updateVersion.sh -p %s,%s -x %s,%s -i %s", hub, pilotSHA, hub, mixerSHA, istioctlURL)
+	_, err = Shell(cmd)
+	return err
+}
+
+// Update the given repository so that it uses the latest dependency references
+// push new branch to remote, create pull request on master,
+// which is auto-merged after presumbit
+func updateDeps(repo string) error {
+	if _, err := Shell("rm -rf " + repo); err != nil {
```
nclando: heads up you can do this with https://golang.org/pkg/os/#RemoveAll, right?

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,111 @@
...
+	owner     = flag.String("owner", "istio-testing", "Github Owner or org.")
+	token     = ""
+)
+
+type githubClient struct {
+	client *github.Client
+}
+
+// Get github api token from tokenFile
+func getToken() (string, error) {
```
nclando: is the token file expected to exist at the beginning of the program execution? is it not expected to change?

if so...maybe read it in main, and then set it...or use a golang `init()` function? then make the program crash if it doesn't exist and then you can avoid having to ever return an error for get token and instead of getToken you can just reference the token variable. or put it on a struct somewhere...

nclando: like githubClient should probably have it on the struct right?

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	b, err := ioutil.ReadFile(*tokenFile)
+	if err != nil {
+		return "", err
+	}
+	token = strings.TrimSpace(string(b[:]))
+	return token, nil
+}
+
+func newGithubClient() (*githubClient, error) {
```
nclando: probably should take argument `token`?

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,111 @@
...
+	log.Printf("Creating a PR with Title: \"%s\" for repo %s", title, repo)
+	pr, _, err := g.client.PullRequests.Create(context.TODO(), *owner, repo, &req)
+	if err != nil {
+		return err
+	}
+	log.Printf("Created new PR at %s", *pr.HTMLURL)
+	return nil
+}
+
+func (g githubClient) getListRepos() ([]string, error) {
```
nclando: nit: s/getListRepots/listRepos

nclando: this function should maybe take a context? or put the context on the client struct?

toolbox/deps_update/util.go
nclando: possibly can delete, you only use helper once, and might be able to be replaced with

`hex.EncodeToString(md5.Sum([]byte(text)))`

maybe

toolbox/deps_update/util.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// GetMD5Hash generates an MD5 digest of the given string
+func GetMD5Hash(text string) string {
+	hash := md5.Sum([]byte(text))
+	return hex.EncodeToString(hash[:])
+}
+
+// Shell run command on shell and get back output and error if get one
+func Shell(format string, args ...interface{}) (string, error) {
```
nclando: my suggestion here is maybe don't bother with writing a helper unless you need the extra logging

rather than doing the `fmt.Sprintf`  with the formatting you can just use the underlying signature and pass in the executable and list of args right? seems funky that you always create the string only to split it again...


like `exec.Command("./some/script.sh", []string{"-p", "flag", "-d", fmt.Sprintf("%d", integerSomething)}).CombinedOutput()` seems nearly equivalent

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+)
+
+var (
+	repo       = flag.String("repo", "", "Update dependencies of only this repository")
+	githubClnt *githubClient
```
nclando: doesn't seem like this should be a global package level variable

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", acct, token, acct, repo), nil
+}
+
+// Update the commit SHA reference in a given line from dependency file
+// to the latest stable version
+// returns the updated line
+func replaceCommit(line string, dep dependency) (string, error) {
```
nclando: this should probably either take as argument a pointer to a client or have a client reciever

i.e., `replaceCommit(gc *githubClient, line string, dep dependency)` or `func (gc *githubClient) replaceCommit(line string, dep dependency)`

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+	commitSHA, err := githubClnt.getHeadCommitSHA(dep.RepoName, dep.ProdBranch)
+	if err != nil {
+		return line, err
+	}
+	idx := strings.Index(line, "\"")
+	return line[:idx] + "\"" + commitSHA + "\",", nil
+}
+
+// Get the list of dependencies of a repo
+// Assumes repo has been cloned to local
```
nclando: cloned to local? or cloned locally? does this program assume that we are in root directory of the repo?

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return deps, err
+	}
+	err = json.Unmarshal(raw, &deps)
+	return deps, err
+}
+
+// Generates an MD5 digest of the version set of the repo dependencies
+// useful in avoiding making duplicate branches of the same code change
+func fingerPrint(deps []dependency) (string, error) {
```
nclando: i think also should either take a github client or be a method with a gc receiver.

nlandolfi (commented):
nice!

most of my comments are a function of
1. trying to get away from package level variables
2. trying to be idiomatic
3. trying to fail the program early if one of assumptions isn't met (like loading token immediately)

just suggestions!

toolbox/deps_update/githubClient.go
sebastienvas: or githubclient

toolbox/deps_update/dependency.go
chxchx: I actually do want to say hoisting here. It means pulling the commit SHA up and out of the bazel rules, defining is as constant string, and referencing it in the bazel rule.

nlandolfi:
ohhhhh gotcha -- SGTM

toolbox/deps_update/githubClient.go
chxchx: I actually see a mix of camel case and snake case in our repos for go files and got confused too. But I guess I will leave it like this unless people have a strong opinion

toolbox/deps_update/githubClient.go
chxchx: This is such a good point. I find myself calling getToken() for quite some times and always have to check for error in case file was not provided. Doing what you said reduces such work to once and clean. Thanks!

nlandolfi:
ya seb pointed out to me that mixer does camel case

toolbox/deps_update/githubClient.go
chxchx: Done

toolbox/deps_update/githubClient.go
chxchx: Done

toolbox/deps_update/githubClient.go
chxchx: Done

toolbox/deps_update/util.go
chxchx: The [:] operator creates a slice (type []byte) from an array (type [16]byte). (See full explanation [here](https://blog.golang.org/go-slices-usage-and-internals)) It is necessary to make such conversion here since md5.Sum() returns [16]byte but hex.EncodeToString wants []byte.

I think this function is generically useful to not just this particular feature/script but also the rest of test-infra so putting it under util is in my opinion a good idea.

toolbox/deps_update/util.go
chxchx: haha I took this code from istio/istio. I blame Seb on this.

toolbox/deps_update/main.go
chxchx: It is declared on a global scope so that other functions in main.go could easily access githubClient functions. The intention is not providing package level visibility. Any way we could do better?

toolbox/deps_update/main.go
chxchx: Any reason why such indirection is necessary?

toolbox/deps_update/main.go
chxchx: The assumption has been removed

toolbox/deps_update/main.go
chxchx: Done.

chxchx:
PTAL

Comments above have been addressed. Specifically, failed PRs are now closed in subsequent runs, istio-ca is taken care of, code style has improved, and duplicate code with github_helper has been refactored. An example PR created by this script is [here](https://github.com/chxchx/istio/pull/6). This update script has to work with `istio.deps` file at root, and please also take a look at https://github.com/istio/istio/pull/510

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+const (
+	istioDepsFile = "istio.deps"
+)
+
+// Update the commit SHA reference in a given line from dependency file
+// to the latest stable version
+// Returns the updated line
+func replaceCommit(line string, dep dependency) (string, error) {
+	idx := strings.Index(line, "\"")
+	return line[:idx] + "\"" + dep.LastStableSHA + "\",", nil
```
yutongz: No need to have an error return if the function never returns an error.

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+	AuthSHA, err := githubClnt.getHeadCommitSHA("auth", "stable")
+	if err != nil {
+		return err
+	}
+	caHub := "docker.io/istio"
+	istioctlURL := fmt.Sprintf(
+		"https://storage.googleapis.com/istio-artifacts/pilot/%s/artifacts/istioctl", pilotSHA)
+	hub := "gcr.io/istio-testing"
```
yutongz: Please make something like docker hub, gcs storage configurable from flags or read from somewhere, try not hardcoded.

yutongz: Try not to hardcoded something like docker hubs and gcs storage, make it configurable and read it from flags or somewhere else.

toolbox/deps_update/main.go
yutongz: If no error is going to be returned, then remove error from return values.

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+	if err := updateDeps(repo, deps); err != nil {
+		return err
+	}
+	if err := serializeDeps(istioDepsFile, &deps); err != nil {
+		return err
+	}
+	if _, err := util.Shell("git add *"); err != nil {
+		return err
+	}
+	if _, err := util.Shell("git commit -m Update_Dependencies"); err != nil {
```
yutongz: Maybe you can use "git commit -am "

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	} else { // update dependencies of all repos in the istio project
+		repos, err := githubClnt.listRepos()
+		if err != nil {
+			log.Panicf("Error when fetching list of repos: %v\n", err)
+			return
+		}
+		for _, r := range repos {
+			if err := updateDependenciesOf(r); err != nil {
+				log.Panicf("Failed to udpate dependency: %v\n", err)
```
yutongz: Careful using panic, you may still want to update other repos even the first one hit an error

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+	if _, err := util.Shell("git push --set-upstream origin " + branch); err != nil {
+		return err
+	}
+	if err := githubClnt.createPullRequest(branch, repo); err != nil {
+		return err
+	}
+	return nil
+}
+
+func main() {
```
yutongz, line 200: I suppose it is a periodic job, then why not use goroutine and wake up every midnight

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+func replaceCommit(line string, dep dependency) (string, error) {
+	idx := strings.Index(line, "\"")
+	return line[:idx] + "\"" + dep.LastStableSHA + "\",", nil
+}
+
+// Generates an MD5 digest of the version set of the repo dependencies
+// useful in avoiding making duplicate branches of the same code change
+// Also updates dependency objects deserialized from istio.deps
+func fingerPrintAndUpdateDepSHA(repo string, deps *[]dependency) (string, error) {
+	digest, err := githubClnt.getHeadCommitSHA(repo, "master")
```
sebastienvas: We want this to work also for release branches, so you should make the branch as a flag.

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,165 @@
...
+	return false, nil
+}
+
+// Checks all auto PRs to update dependencies
+// Close the ones that have failed the presubmit
+// Deletes the remote branches from which the PRs are made
+func (g githubClient) closeFailedPullRequests(repo string) error {
+	log.Printf("Search for failed auto PRs to update dependencies in repo %s", repo)
+	options := github.PullRequestListOptions{
+		Base:  "master",
```
sebastienvas: Same comment here. Use the branch as a flag.

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+	output := strings.Join(lines, "\n")
+	return ioutil.WriteFile(dep.File, []byte(output), 0600)
+}
+
+// Assumes at the root of istio directory
+// Runs the updateVersion.sh script
+func updateIstioDeps() error {
+	pilotSHA, err := githubClnt.getHeadCommitSHA("pilot", "stable")
```
sebastienvas: Do you have to hard code this? Can you look at the istio.deps files and generate this from the json files?

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	AuthSHA, err := githubClnt.getHeadCommitSHA("auth", "stable")
+	if err != nil {
+		return err
+	}
+	caHub := "docker.io/istio"
+	istioctlURL := fmt.Sprintf(
+		"https://storage.googleapis.com/istio-artifacts/pilot/%s/artifacts/istioctl", pilotSHA)
+	hub := "gcr.io/istio-testing"
+	cmd := fmt.Sprintf("./install/updateVersion.sh -p %s,%s -x %s,%s -i %s -c %s,%s",
```
sebastienvas: Maybe dynamically update the flags based which deps needs to be updated.

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Create a pull request within repo from branch to master
+func (g githubClient) createPullRequest(branch, repo string) error {
+	if branch == "" {
+		return errors.New("branch cannot be empty")
+	}
+	title := prTitlePrefix + repo
+	body := "This PR will be merged automatically once checks are successful."
+	base := "master"
```
sebastienvas: Same comment here. Make the branch a flag.

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,165 @@
...
+		State: "open",
+	}
+	prs, _, err := g.client.PullRequests.List(
+		context.Background(), g.owner, repo, &options)
+	if err != nil {
+		return err
+	}
+	var multiErr error
+	for _, pr := range prs {
+		if strings.Contains(*pr.Title, prTitlePrefix) {
```
sebastienvas: Don't you want to look only those that have the prefix ?

toolbox/deps_update/main.go
chxchx: You are absolutely right. There has been a refactor and I forgot to take the error out.

toolbox/deps_update/main.go
chxchx: Done

toolbox/deps_update/main.go
chxchx: Done

toolbox/deps_update/main.go
chxchx: I agree. It is now read from istio.VERSION directly.

toolbox/deps_update/main.go
chxchx: I thought the intention is to set up the cron jobs thru jenkins/prow and this binary does not care when it is invoked (ordering of repo) or how it is triggered (cron versus PR). I was trying to leave for room for the future if we want to update istio project by building from leaves.

chxchx:
PTAL
An example PR on istio/istio created by this script can be found here https://github.com/chxchx/istio/pull/9/files

sebastienvas:
you should use pretty print on the json file. That being said, this looks promising

sebastienvas (changes requested):

toolbox/deps_update/main.go
chxchx: Done

toolbox/deps_update/main.go
chxchx: There has to be some hadcoding in order to use `install/updateVersion.sh` because each dependency is passed in with different flag and hub. Such flags and hubs are only applicable to istio/istio so they should not be put into the json file. I did refactor this such that SHAs are read from deps array (read from json file) but I still have to specify the name such as "pilot" and "mixer" to properly use the flags. Let me know if you have smarter ways to do this.

toolbox/deps_update/githubClient.go
chxchx: Done

toolbox/deps_update/githubClient.go
chxchx: Done

toolbox/deps_update/githubClient.go
chxchx: Done

toolbox/deps_update/main.go
chxchx: It is hard since there is also correspondence between dependency and hub, and istioctlURL does not have a hub. The `updateVersion.sh` is just a bit ad hoc so I think what I have here is actually the most concise

chxchx:
PTAL
Example PR made with this script plus pretty-print json: https://github.com/chxchx/istio/pull/10/files

toolbox/deps_update/githubClient.go
```diff
@@ -0,0 +1,164 @@
...
+			context.Background(), g.owner, repo, *pr.Head.SHA, nil)
+		multiErr = multierror.Append(multiErr, err)
+		if util.GetCIState(combinedStatus, nil) == ci.Failure {
+			log.Printf("Closing PR %s/%s#%d and deleting branch %s",
+				g.owner, repo, *pr.Number, *pr.Head.Ref)
+			*pr.State = "closed"
+			if _, _, err := g.client.PullRequests.Edit(
+				context.Background(), g.owner, repo, *pr.Number, pr); err != nil {
+				multiErr = multierror.Append(multiErr, err)
+				log.Printf("Failed to delete branch %s in repo %s", *pr.Head.Ref, repo)
```
sebastienvas: Did you mean failed to close PR ?

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+		digest += commitSHA
+		(*deps)[i].LastStableSHA = commitSHA
+	}
+	return util.GetMD5Hash(digest), nil
+}
+
+// Read the exported environment variable named :key in istio.VERSION
+// and return its value
+func istioVersionGet(key string) (string, error) {
```
sebastienvas: Could you just read it once, and then create a map ?

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+		return ""
+	}
+	pilotSHA := getSHAByName("pilot")
+	mixerSHA := getSHAByName("mixer")
+	AuthSHA := getSHAByName("auth")
+	if pilotSHA == "" || mixerSHA == "" || AuthSHA == "" {
+		return fmt.Errorf("incomplete dependencies in %s", istioDepsFile)
+	}
+	caHub, err := istioVersionGet("CA_HUB")
```
sebastienvas: I think there is a better way to do this. You can just update directly the istio.VERSION file like you do for the other cases, but when you update a deps, you have to make sure to update the docker registry to gcr.io/istio-testing (this should be a flag somewhere) and then you can run ./install/updateVersion.sh without args.

sebastienvas (changes requested):
PTAL

toolbox/deps_update/githubClient.go
chxchx: Indeed. I have corrected this.

toolbox/deps_update/main.go
chxchx: I like your idea of updating istio.VERSION directly since ./install/updateVersion.sh will pick up the latest changes I made to istio.VERSION. However, there is a minor twist around istioctlURL as you will see from my comment in code

toolbox/deps_update/main.go
chxchx: It is difficult to create such a map. I can source the exported vars in istio.VERSION to the environment, but I won't know which variables are exactly the ones I just exported. To make a map, I could either read the entire env and put it into map (huge), or I could manually parse istio.VERSION line by line (file and format specific, create source of bugs once istio.VERSION changes). Neither of those solutions is as simple and maintainable as the one I have now. But do let me know if there is better way to do this.

chxchx:
PTAL

Minor dispute over the map thing but I agree with everything else. An Example PR created by the script is here https://github.com/chxchx/istio/pull/11/files

sebastienvas:
The PR that you pointed is still using gcr.io/istio-io in some of those update.

toolbox/deps_update/main.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"istio.io/test-infra/toolbox/util"
+)
+
+var (
+	repo       = flag.String("repo", "", "Optional. Update dependencies of only this repository")
+	owner      = flag.String("owner", "istio", "Github Owner or org")
+	tokenFile  = flag.String("token_file", "", "File containing Github API Access Token")
+	baseBranch = flag.String("base_branch", "master", "Branch from which the deps update commit is based")
+	caHub      = flag.String("ca_hub", "", "Optional. Where new CA image is hosted")
```
sebastienvas: Just create one hub for all of them. It is the same we use for testing.

sebastienvas (changes requested):
I can t see your comments. Let s do an hour of pair programming tomorrow so that I can understand where you are going

toolbox/deps_update/main.go
chxchx: Are you sure? Because on istio master, the three hubs in istio.VERSION are all different. Unless you are saying they all resolve to the same host but maybe having the option of updaing only one of the hubs and not others could be useful.

chxchx:
You are absolutely right. I forgot to rebase from istio/istio. Please take a look on this https://github.com/chxchx/istio/pull/12/files

Also, this PR now contingent on the correction I just made at https://github.com/istio/istio/pull/534 so please also take a look on that.

toolbox/deps_update/main.go
nclando: shoot just saw this -- gotcha, the other option is to pass it down as arguments to functions.

chxchx:
PTAL

I hope I understand what @sebastienvas wants for the hub flag. Example PR created by this script can be found here  https://github.com/chxchx/istio/pull/14/files

sebastienvas (approved):

PR #338, opened 2017-07-27
Update prowbazel image.

sebastienvas:
* Remove --repo bootstrap flag from entrypoint
* Update bazel to 0.5.2 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docker/prowbazel/Makefile
+++ b/docker/prowbazel/Makefile
@@ -1,5 +1,5 @@
 PROJECT = istio-testing
-VERSION = 0.1.5
+VERSION = 0.1.7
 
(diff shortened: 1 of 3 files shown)


nlandolfi:
do we not need the `--repo` flag?

sebastienvas:
@nlandolfi I think I figured it out. I updated the template so now prow.istio.io should have the right logs link. (Famous last words)

sebastienvas:
I don't think we need the --repo flag since we set it in every job already (repo or pull)

nlandolfi (approved):
assuming you ran `make image && make push` from the prowbazel dir!

PR #339, opened 2017-07-27
Final Fix for cleanup cluster script

sebastienvas:
**Release note**:
NONE
```release-note
```

### DIFF
--- a/prow/test-infra-cleanup-cluster.sh
+++ b/prow/test-infra-cleanup-cluster.sh
@@ -27,4 +27,4 @@ set -u
 set -x
 
 echo "=== Cleaning up cluster ==="
-./scripts/cleanup-cluster -h 2
+./scripts/cleanup-cluster -c '' -h 2
--- a/scripts/cleanup-cluster
+++ b/scripts/cleanup-cluster
@@ -25,9 +25,11 @@ done
 [[ -n "${PROJECT_ID}" ]] && ARGS+=(--project "${PROJECT_ID}")
 [[ -n "${ZONE}" ]] && ARGS+=(--zone "${ZONE}")
 
-echo 'Getting credentials for Kubernetes cluster'
-gcloud container clusters get-credentials ${CLUSTER_NAME} ${ARGS[@]}\
-  || error_exit 'Could not get kubectl config for cluster'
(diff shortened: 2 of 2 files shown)


sebastienvas:
@nlandolfi Finally got it to work
https://pantheon.corp.google.com/storage/browser/istio-prow/test-infra-cleanup-cluster/28/. 

But the path from prow is wrong. Could we match k8s hiearchy and store all the logs in istio-prow/logs ?

nlandolfi:
heads up you have to put the NONE in the code fence for the release-note

/release-note-none

yeah we can do that! I'm sort of confused as to how they discriminate on where to store what...since we have it working by storing the presubmit stuff in the root dir

nlandolfi (approved):
LGTM

prow/test-infra-cleanup-cluster.sh
```diff
@@ -27,4 +27,4 @@ set -u
 set -x
 
 echo "=== Cleaning up cluster ==="
-./scripts/cleanup-cluster -h 2
+./scripts/cleanup-cluster -c '' -h 2
```
nclando, line 30: @sebastienvas why add `-c ''` isn't' that the same as no flag at all in this situation?

nlandolfi (changes requested):

prow/test-infra-cleanup-cluster.sh
nclando: or I see that disables getting the credentials basically?

sebastienvas: CLUSTER_NAME is set to e2e-testing by default. Because we are mounting kubeconfig we don't need to call get-credentials. I was also getting an error before changing this
ERROR: (gcloud.container.clusters.get-credentials) ResponseError: code=403, message=Required "container.clusters.get" permission for "projects/istio-testing/zones/us-west1-a/clusters/e2e-testing".

sebastienvas (commented):
PTAL

nlandolfi (approved):

nlandolfi:
@istio-testing test this

nlandolfi:
@istio-testing test this

PR #348, opened 2017-07-28
Revert "use deck url for github status"

nlandolfi:
Reverts istio/test-infra#347

/release-note-none
/lgtm
/approve no-issue


nlandolfi:
heads up: I released to check the URL for the case where, once the status changes, the URL is changed to Prow. Deck currently only evaluates the template once so this did not work. Will need more work to get the different URLs than just changing the template.

PR #357, opened 2017-08-03
Enable Prow for Mixer

sebastienvas:
```release-note
NONE
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -120,6 +120,7 @@ presubmits:
   istio/mixer:
   - name: mixer-presubmit
     context: prow/mixer-presubmit.sh
+    skip_reports: true
     always_run: true
     rerun_command: "@istio-testing bazel test this"
     trigger: "((?m)^@istio-testing (bazel )?test this,?(\\s+|$)|(?m)^/test( all| bazel),?(\\s+|$))"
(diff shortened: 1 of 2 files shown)


prow/config.yaml
```diff
@@ -116,6 +117,88 @@ presubmits:
...
+          secretName: mixer-codecov-token
+      - name: cache-ssd
+        hostPath:
+          path: /mnt/disks/ssd0
+    run_after_success:
+    - name: mixer-e2e-smoketest
+      context: prow/mixer-e2e-smoketest.sh
+      always_run: true
+      rerun_command: "@istio-testing smoke test this"
+      trigger: "((?m)^@istio-testing (smoke )?test this,?(\\s+|$)|(?m)^/test( all| bazel),?(\\s+|$))"
```
nclando: can make this `/test( all| smoke)` please...also thinking we can move away from `@istio-testing smoke test` style comments in favor of promoting the `/test smoke` style  comments and maybe eventually require the `@istio-testing` ones

nlandolfi (changes requested):
one minor thing

nlandolfi:
also we should skip_reports until the job files go to master

prow/config.yaml
sebastienvas: Can we change all of those in another PR ?

sebastienvas:
PTAL

nlandolfi (approved):
yeah just skip reports for e2e as well as discussed

prow/config.yaml
nclando: ya

nlandolfi (approved):

PR #361, opened 2017-08-04
Install CMake in Jenkins image

lizan:
To use envoy native build.

**Release note**:
```release-note
NONE
```


nlandolfi:
@sebastienvas should we rebuild the prowbazel image?

sebastienvas:
proxy is still on Jenkins so we don t have to.

PR #362, opened 2017-08-04
Add istio/istio e2e-suite

yutongz:
And post-submit job on istio/istio to run e2e-test suite with rbac and auth.
Will add kube-config of **e2e-testing-rbac**(cluster with rbac) as a secret to prow cluster after Seb re-create all clusters.
Add do-not-merge label to block merge until everything is in.

**Release note**:

```release-note
None
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -39,7 +39,7 @@ presubmits:
     - master
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.1.7
+      - image: gcr.io/istio-testing/prowbazel:0.2.1
         args:
         - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
         - "--clean"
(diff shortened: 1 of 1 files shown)


sebastienvas (approved):
you need to create the secret on the cluster. Let me know if you need help with that.

ldemailly:
if this goes in before #365 I will put 0.2.0 also for post submit after rebase

looks like the bot doesn't auto merge this repo?

nlandolfi:
@ldemailly auto merge has been disabled for this PR because of the do-not-merge label -- not sure why the label was put there but otherwise it would be merged by the submit queue.

ldemailly:
looks like my pr went it before this (thanks!) so can you put 0.2.0 for all istio/istio images/config
before pushing

nlandolfi (changes requested):
as I recall it, you wanted to run these two jobs in parallel right?

This is expressing that you first run the non rbac and then _after succes_ the rbac. So if no rbac takes 30 mins and rbac takes 30 mins it takes an hour to see two successes...whereas both in parallel could take more like ~30 mins.

Of course the tradeoff is it uses more resources.

Thoughts?

nlandolfi:
also as Laurent mentioned, we should make these be the prowbazel 0.2 image

ldemailly:
btw the image I made is still not enough so I'm looking at building another one and pulling my hair...

ldemailly:
make that 0.2.1 (#366) - thanks !

ldemailly:
@rshriram mentioned that Envoy doesn't work/crashes on the latest Docker CE so we might need to keep 0.1.7 for proxy build (or try locally/confirm and/or change it but be ready to put back 0.1.7 if something goes wrong) - it's unclear also if it's a build or run time issue (as istio/istio build works fine with 0.2.1)

sebastienvas:
proxy is currently build on Jenkins so your change is already active for Proxy.

ldemailly:
jenkins is using the prow image 0.2.1 ?

sebastienvas:
Jenkins build its own image as a post submit.

On Aug 9, 2017 4:07 PM, "Laurent Demailly" <notifications@github.com> wrote:

jenkins is using the prow image 0.2.1 ?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/istio/test-infra/pull/362#issuecomment-321405705>, or mute
the thread

nlandolfi (approved):
lgtm

yutongz:
@sebastienvas PTAL and which prowbazel we should use? I am lost in that part.

sebastienvas (approved):
You can use 0.2.1 for all the pods that have privilege mode.

sebastienvas:
Since you are running all combination as part of run after, does it mean that we can remove the e2e from the presubmit ?

yutongz:
Yes, that's the plan. I will remove it until I am confident about the new tests.

PR #369, opened 2017-08-10
Binary to fast forward stable branch

chxchx:
Given that there is no longer any PR from master to stable, this binary is required to fast forward stable branch. It is structured in a flexible way that is extensible for additional features in the near future.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -1,0 +1,75 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


toolbox/githubctl/main.go
```diff
@@ -0,0 +1,75 @@
...
+	if value == nil || *value == "" {
+		log.Panicf("%s must be specified\n", name)
+	}
+}
+
+func fastForward(repo, baseBranch, refBranch *string) error {
+	assertNotEmpty("repo", repo)
+	assertNotEmpty("baseBranch", baseBranch)
+	assertNotEmpty("refBranch", refBranch)
+	sha, err := githubClnt.GetHeadCommitSHA(*repo, *refBranch)
```
sebastienvas: I think in our case, you don't want to use the Head commit, but the commit you are testing with, so maybe the ref(commit) should be passed as an argument.

sebastienvas (changes requested):

toolbox/githubctl/main.go
chxchx: Done. This is a really good point. Especially because a reference branch might be moving and by the time we run this script its head points to a different SHA than what we want.

chxchx:
PTAL

sebastienvas (approved):

sebastienvas:
/lgtm
/approve no-issue
Try to create an issue on github next time your work on something like that and map it to our spreadsheet.

PR #372, opened 2017-08-11
Upgrading prow to match k8s deployment

sebastienvas:
```release-note
NONE
```


sebastienvas:
Thanks @nlandolfi for this. The upgrade had some hiccup, but we were able to resolve all of them without downtime.

PR #373, opened 2017-08-11
url points to deck, then gubernator

nlandolfi:
This obviously won't get in before I leave, but once kubernetes merges the PR that re-executes the URL template when the ProwJob's status changes, we can deploy this I believe to have it first point to deck and then to gubernator.

**Release note**:

```release-note
none
```


nlandolfi:
https://github.com/kubernetes/test-infra/pull/3876 went in, so whenever we update plank to `0.39`, then we can consider merging this.

sebastienvas:
Ryan fixed gubernator to work properly with istio, so we don't need this anymore.

PR #385, opened 2017-08-16
Dependency freshness metrics from a stable build

chxchx:
Generates statistics about how stale the dependency SHAs are compared to their respective latest stable versions.
Also refactored utilities to be more universal applicable.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/toolbox/metrics/buildFreshness.go
+++ b/toolbox/metrics/buildFreshness.go
@@ -1,0 +1,96 @@
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"strings"
+	"sync"
+	"time"
+
+	multierror "github.com/hashicorp/go-multierror"
+	u "istio.io/test-infra/toolbox/util"
+)
+
+const (
+	istioDepsFile = "istio.deps"
+)
+
+// DepFreshness stores how many days behind the dependency SHA used by the last
+// stable build is compared to the HEAD of the production branch of that dependency
+type DepFreshness struct {
+	Dep       u.Dependency
+	Freshness time.Duration
+}
+
+func getBranchHeadTime(gclient *u.GithubClient, repo, branch string) (*time.Time, error) {
+	sha, err := gclient.GetHeadCommitSHA(repo, branch)
+	if err != nil {
+		return nil, err
+	}
+	t, err := gclient.GetSHATime(repo, sha)
+	if err != nil {
+		return nil, err
+	}
+	return t, nil
+}
+
+func getStableBuildFreshness(owner, repo, branch string) ([]DepFreshness, error) {
+	githubClnt := u.NewGithubClientNoAuth(owner)
+	var stats []DepFreshness
+	pickledDeps, err := githubClnt.GetFileContent(repo, branch, istioDepsFile)
+	if err != nil {
+		return stats, err
+	}
+	deps, err := u.DeserializeDepsFromString(pickledDeps)
+	if err != nil {
+		return stats, err
+	}
+	var wg sync.WaitGroup
+	var mutex = &sync.Mutex{} // used to synchronize access to stats and multiErr
+	var multiErr error
+	for _, dep := range deps {
+		wg.Add(1)
+		go func(dep u.Dependency) {
+			latestTime, err := getBranchHeadTime(githubClnt, dep.RepoName, dep.ProdBranch)
+			if err != nil {
+				// multierror not thread safe
+				mutex.Lock()
+				multiErr = multierror.Append(multiErr, err)
+				mutex.Unlock()
+				return
+			}
+			var stableTime *time.Time
+			if strings.Contains(dep.LastStableSHA, "-") || strings.Contains(dep.LastStableSHA, ".") {
+				// the reference is a tag
+				stableTime, err = githubClnt.GetTagPublishTime(dep.RepoName, dep.LastStableSHA)
+			} else {
+				stableTime, err = githubClnt.GetSHATime(dep.RepoName, dep.LastStableSHA)
+			}
+			if err != nil {
+				mutex.Lock()
+				multiErr = multierror.Append(multiErr, err)
+				mutex.Unlock()
+				return
+			}
+			lag := latestTime.Sub(*stableTime)
+			mutex.Lock()
+			stats = append(stats, DepFreshness{dep, lag})
+			mutex.Unlock()
+			wg.Done()
+		}(dep)
+	}
+	wg.Wait()
+	return stats, multiErr
+}
--- a/toolbox/metrics/buildFreshness.go
+++ b/toolbox/metrics/buildFreshness.go
@@ -27,37 +27,56 @@ const (
 	istioDepsFile = "istio.deps"
 )
 
-// DepFreshness stores how many days behind the dependency SHA used by the last
-// stable build is compared to the HEAD of the production branch of that dependency
+// DepFreshness stores how long the dependency SHA used by the last
+// stable build is behind the HEAD of the production branch of that dependency
 type DepFreshness struct {
-	Dep       u.Dependency
-	Freshness time.Duration
+	Dep u.Dependency
+	Age time.Duration
 }
 
-func getBranchHeadTime(ghClient *u.GithubClient, repo, branch string) (*time.Time, error) {
-	sha, err := ghClient.GetHeadCommitSHA(repo, branch)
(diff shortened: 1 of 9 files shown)


chxchx:
/assign @yutongz 
/assign @sebastienvas 
PTAL

toolbox/metrics/buildFreshness.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// DepFreshness stores how many days behind the dependency SHA used by the last
+// stable build is compared to the HEAD of the production branch of that dependency
+type DepFreshness struct {
+	Dep       u.Dependency
+	Freshness time.Duration
+}
+
+func getBranchHeadTime(gclient *u.GithubClient, repo, branch string) (*time.Time, error) {
```
yutongz: Could you change "gclient" to "ghClient"

toolbox/metrics/buildFreshness.go
```diff
@@ -0,0 +1,108 @@
...
+				multiErrAppend(err)
+				return
+			}
+			latestTime, err := getBranchHeadTime(githubClnt, dep.RepoName, dep.ProdBranch)
+			if err != nil {
+				multiErrAppend(err)
+				return
+			}
+			var stableTime *time.Time
+			if strings.Contains(dep.LastStableSHA, "-") || strings.Contains(dep.LastStableSHA, ".") {
```
yutongz: You may want to see if it's a SHA: [a-z0-9]{40}, if not then assume it's a tag

toolbox/metrics/buildFreshness.go
```diff
@@ -0,0 +1,108 @@
...
+	dockerImageRegex := regexp.MustCompile(
+		"^[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}$")
+	multiErrAppend := func(err error) {
+		// multierror not thread safe
+		mutex.Lock()
+		multiErr = multierror.Append(multiErr, err)
+		mutex.Unlock()
+	}
+	for _, dep := range deps {
+		wg.Add(1)
```
yutongz: It's pretty cool to get data parallel, but is it necessary? It's a pain to collect and read log and error message.

toolbox/util/githubClient.go
```diff
@@ -164,10 +172,49 @@ func (g GithubClient) CloseFailedPullRequests(prTitlePrefix, repo, baseBranch st
 
 // GetHeadCommitSHA finds the SHA of the commit to which the HEAD of branch points
 func (g GithubClient) GetHeadCommitSHA(repo, branch string) (string, error) {
-	ref, _, err := g.client.Git.GetRef(
-		context.Background(), g.owner, repo, "refs/heads/"+branch)
+	return g.getReferenceSHA(repo, "refs/heads/"+branch)
+}
+
+// GetSHATime gets the time when sha is made
```
yutongz: when the sha is created

toolbox/util/githubClient.go
```diff
@@ -164,10 +172,49 @@ func (g GithubClient) CloseFailedPullRequests(prTitlePrefix, repo, baseBranch st
...
+	commit, _, err := g.client.Git.GetCommit(
+		context.Background(), g.owner, repo, sha)
+	if err != nil {
+		return nil, err
+	}
+	return (*(*commit).Author).Date, nil
+}
+
+// GetTagPublishTime finds the time a tag is published
+func (g GithubClient) GetTagPublishTime(repo, tag string) (*time.Time, error) {
```
yutongz: Kind of suspicious, the time when a tag created is not necessarily to be the time when the corresponding SHA is created.  Could you double check the which time the api gives you. SHA time seems to have more value.

toolbox/util/githubClient.go
```diff
@@ -164,10 +172,49 @@ func (g GithubClient) CloseFailedPullRequests(prTitlePrefix, repo, baseBranch st
...
+
+// GetFileContent retrieves the file content from the hosted repo
+func (g GithubClient) GetFileContent(repo, branch, path string) (string, error) {
+	opt := github.RepositoryContentGetOptions{branch}
+	fileContent, _, _, err := g.client.Repositories.GetContents(
+		context.Background(), g.owner, repo, path, &opt)
+	if err != nil {
+		return "", err
+	}
+	return fileContent.GetContent()
```
yutongz, line 212: return fileContent.GetContent(), nil

yutongz (changes requested):

toolbox/util/githubClient.go
chxchx: fileContent.GetContent() returns both a string and error so the trailing nil would be unnecessary.

toolbox/metrics/buildFreshness.go
chxchx: The thing is, this program can handle either sha or tag, but if it isa docker image name, then github api is not sufficient to know the time created, thats why I wanted to check this specific case.

toolbox/metrics/buildFreshness.go
chxchx: I would say collecting age metrics in parallel is a big win for us. For each dependency, we make quite a few RPCs thru github API and suffer delay from multiple RTT, but the age metrics of each repo has no inter-dependency that requires serial execution and should be executed concurrently. But you do have a point about logs and errors so I will make sure they clearly states what has gone wrong in which part.

toolbox/metrics/buildFreshness.go
chxchx: Done

toolbox/util/githubClient.go
chxchx: Done

toolbox/util/githubClient.go
sebastienvas: I think we talked about this lately. We should find the SHA the tag points to, and returns the commit date of it.

toolbox/metrics/buildFreshness.go
sebastienvas: I think we can create a regex for SHA and tag that matches X.Y.Z. if anything else we should print a warning that the tag provided cannot be mapped to a git reference.

toolbox/util/githubClient.go
chxchx: You are absolutely right. I overlooked the fact that the sha of a commit is different from that of a tag. I has been fixed in the latest commit.

toolbox/metrics/buildFreshness.go
```diff
@@ -0,0 +1,114 @@
...
+				e := fmt.Errorf(
+					"failed to get the publish time of the stable dependency named %s: %v",
+					dep.Name, err)
+				multiErrAppend(e)
+				return
+			}
+			lag := latestTime.Sub(*stableTime)
+			mutex.Lock()
+			stats = append(stats, DepFreshness{dep, lag})
+			mutex.Unlock()
```
sebastienvas: defer mutex.Unlock()

toolbox/metrics/buildFreshness.go
```diff
@@ -0,0 +1,114 @@
...
+			latestTime, err := getBranchHeadTime(githubClnt, dep.RepoName, dep.ProdBranch)
+			if err != nil {
+				e := fmt.Errorf(
+					"failed to get the committed time of HEAD of branch %s on repo %s: %v",
+					dep.ProdBranch, dep.RepoName, err)
+				multiErrAppend(e)
+				return
+			}
+			var stableTime *time.Time
+			if strings.Contains(dep.LastStableSHA, "-") || strings.Contains(dep.LastStableSHA, ".") {
```
sebastienvas: It would be good to create a regex for SHA and for release tags (X.Y.Z). Or just create a function that returns CreationTime from a ref (try first sha, then on error tag, then error) and print an error or a warning that we could not find the creation time from the provided ref.

toolbox/metrics/buildFreshness.go
sebastienvas: Could you create a function that does everything sequentially, and another that call that sequential function and add the mutex and errors. I think It ll be easier to read.

sebastienvas (changes requested):
PTAL

toolbox/metrics/buildFreshness.go
chxchx: Done

toolbox/metrics/buildFreshness.go
chxchx: Done

toolbox/metrics/buildFreshness.go
chxchx: Done

toolbox/metrics/buildFreshness.go
chxchx: Done

chxchx:
PTAL

toolbox/util/githubUtils.go
```diff
@@ -17,13 +17,18 @@ package util
...
 
 	"github.com/google/go-github/github"
 )
 
 var (
 	ci = NewCIState()
+	// SHARegex matches commit SHA's
+	SHARegex = regexp.MustCompile("^[a-z0-9]{40}$")
+	// ReleaseTagRegex matches release tags
+	ReleaseTagRegex = regexp.MustCompile("^[0-9]*.[0-9]*.[0-9]*$")
```
sebastienvas: [0-9]+.[0-9]+.[0-9]+ Right ? you expect at least one digit.

sebastienvas (approved):
minor change

yutongz (approved):

toolbox/util/githubUtils.go
chxchx: Done

PR #393, opened 2017-08-18
Automate release tagging, archives uploading, and create PR to update istio.RELEASE and downloadIstio.sh

chxchx:
**Release note**:

```release-note
None
```

### DIFF
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -104,7 +104,7 @@ func updateDependenciesOf(repo string) error {
 		return err
 	}
 	defer func() {
-		if err := u.RemoveLocalRepo(repoDir); err != nil {
+		if err = u.RemoveLocalRepo(repoDir); err != nil {
 			log.Fatalf("Error during clean up: %v\n", err)
 		}
 	}()
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -99,11 +99,12 @@ func updateDeps(repo string, deps *[]u.Dependency, depChangeList *[]u.Dependency
 // which is auto-merged after presumbit
 func updateDependenciesOf(repo string) error {
 	log.Printf("Updating dependencies of %s\n", repo)
-	if err := u.CloneRepoCheckoutBranch(githubClnt, repo, *baseBranch); err != nil {
+	repoDir, err := u.CloneRepoCheckoutBranch(githubClnt, repo, *baseBranch, *baseBranch)
+	if err != nil {
 		return err
 	}
 	defer func() {
-		if err := u.RemoveLocalRepo(repo); err != nil {
+		if err := u.RemoveLocalRepo(repoDir); err != nil {
 			log.Fatalf("Error during clean up: %v\n", err)
 		}
 	}()
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -17,10 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"log"
-	"os"
-	"strings"
 
 	u "istio.io/test-infra/toolbox/util"
 )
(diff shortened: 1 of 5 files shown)


chxchx:
/assign @yutongz
/assign @sebastienvas
PTAL

toolbox/deps_update/main.go
```diff
@@ -188,10 +141,7 @@ func updateDependenciesOf(repo string) error {
 	if err := u.SerializeDeps(istioDepsFile, &deps); err != nil {
 		return err
 	}
-	if _, err := u.Shell("git commit -am Update_Dependencies"); err != nil {
-		return err
-	}
-	if _, err := u.Shell("git push --set-upstream origin " + branch); err != nil {
+	if err := u.CreateCommitPushToRemote(branch, "Update_Dependencies"); err != nil {
```
sebastienvas, line 145: Note that there is an API to update files directly. So you could just use that instead of having the create a local repo.

sebastienvas: Let 's do that in a cleanup PR.

toolbox/githubctl/main.go
```diff
@@ -45,14 +60,111 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
+	for _, dep := range deps {
+		if err := githubClnt.CreateAnnotatedTag(
+			dep.RepoName, releaseTag, dep.LastStableSHA, releaseMsg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func findPathToArchives(consoleLog *string) (string, error) {
```
sebastienvas: I think it is easier to update  ./release/create_release_archives.sh to take a flag on where the archive should be created.

toolbox/githubctl/main.go
```diff
@@ -45,14 +60,111 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
+	if err := u.CloneRepoCheckoutBranch(
+		githubClnt, istioRepo, *baseBranch); err != nil {
+		return err
+	}
+	defer func() {
+		if err := u.RemoveLocalRepo(istioRepo); err != nil {
+			log.Fatalf("Error during clean up: %v\n", err)
+		}
+	}()
+	output, err := u.Shell("bash ./release/create_release_archives.sh")
```
sebastienvas: Please update create_release_archives to take in a flag for where to generate artifacts.

toolbox/githubctl/main.go
```diff
@@ -45,14 +60,111 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
+		dir := lastLine[len(logPrefix):]
+		return dir, nil
+	}
+	return "", fmt.Errorf("failed to find path to release archives from console log")
+}
+
+func updateIstioVersion(releaseTag string) error {
+	hubCommaTag := fmt.Sprintf("%s,%s", dockerHub, releaseTag)
+	istioctl := fmt.Sprintf(
+		"https://storage.googleapis.com/istio-artifacts/pilot/%s/artifacts/istioctl", releaseTag)
```
sebastienvas: For release we might use a different bucket. In any way make it a constant. Sync with  @rvkubiak on where things are going to be uploaded.

toolbox/util/githubUtils.go
```diff
@@ -87,3 +88,47 @@ func GetAPITokenFromFile(tokenFile string) (string, error) {
...
+	if err := os.Chdir(repo); err != nil {
+		return err
+	}
+	_, err := Shell("git checkout " + branch)
+	return err
+}
+
+// RemoveLocalRepo deletes the local git repo just cloned
+func RemoveLocalRepo(repo string) error {
+	if err := os.Chdir(".."); err != nil {
```
sebastienvas: I am kind of uncomfortable with those chdir, as if I call those method in a random order I might not know where I end up. Could you see how difficult it would be to use the github api to changes files ? One option is that for some files changes (ie increasing version) we might not need to create a PR, we can just have the bot do it direclty.

toolbox/githubctl/main.go
```diff
@@ -45,14 +60,111 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
+	}
+	releaseTag, err := githubClnt.GetFileContent(istioRepo, *baseBranch, releaseTagFile)
+	if err != nil {
+		return err
+	}
+	releaseTag = strings.Trim(releaseTag, "\n")
+	releaseMsg := "Istio Release " + releaseTag
+	for _, dep := range deps {
+		if err := githubClnt.CreateAnnotatedTag(
+			dep.RepoName, releaseTag, dep.LastStableSHA, releaseMsg); err != nil {
```
sebastienvas: Are we guaranteed to have dep.LastStableSHA in sync with the value in istio.VERSION ?

sebastienvas (changes requested):
Nice. Almost there!

chxchx:
@david-mcmahon Besides the comments above, two PRs need to be created instead of one as stated in https://github.com/istio/istio/blob/master/release/README.md, which I am working on. Other than that, feel free to take a look.

toolbox/githubctl/main.go

toolbox/githubctl/main.go

toolbox/githubctl/main.go
chxchx: This is a valid concern. There is a chance that someone may run updateVersion.sh directly without also updating istio.deps. In latest commit, I read the sha from istio.VERSION to make sure we get the right thing.

toolbox/githubctl/main.go

toolbox/util/githubUtils.go
chxchx: In latest commit, I use absolute path instead to mitigate the ordering problem.

chxchx:
PTAL

toolbox/deps_update/main.go
sebastienvas: As we talked. Let s consider doing this in anther PR.

sebastienvas (approved):

PR #397, opened 2017-08-19
Add manually rerun command

yutongz:
**Release note**:

```release-note
None
```


sebastienvas (changes requested):
That s not enough. The rerun command does not do anything (it is just used for the report). You need to add the trigger. 
trigger: "(?m)^/(retest|test (all|istio-presubmit))?,?(\\s+|$)"

sebastienvas (approved):

PR #400, opened 2017-08-22
Periodic Prow Job to Update Istio Dependency SHA

chxchx:
An example PR created is https://github.com/istio/istio/pull/588

Notice that it reads SHA from stable branch of each dependency. For the entire dependency update scheme to be automated, it also needs the fast-forward jobs that move stable up-to-date with master, which will be done in another PR shortly after.

**Release note**:

```release-note
NONE
```


chxchx:
/assign @sebastienvas 
/assign @yutongz 
PTAL

sebastienvas (approved):

PR #401, opened 2017-08-22
Bazel 0.5.2 + Vendoring

sebastienvas:
Update to bazel 0.5.2
Properly vendor our tools

```release-note
NONE
```


sebastienvas:
@chxchx You should be able to use this instead of declaring all the deps in repos.

yutongz:
/lgtm
/approve
Nice job!

PR #402, opened 2017-08-22
Mount github token to postsubmits

chxchx:
**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -538,6 +538,9 @@ postsubmits:
           readOnly: true
         - name: cache-ssd
           mountPath: /home/bootstrap/.cache
+        - name: oauth
+          mountPath: /etc/github
+          readOnly: true
         ports:
         - containerPort: 9999
           hostPort: 9998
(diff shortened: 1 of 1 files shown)


nlandolfi (approved):
LGTM

One thing I meant to voice before I left is the security semi-concern implicit in mounting credentials to a pod. Hypothetically someone could write a PR that modified a prow job to cat the credential and then steal it...that said I realize we _need_ to mount these.

postsubmit feels safer tho because it only gets run after someone from the org approves it which is a good filter. But for presubmits I've thought about this...thoughts?

cc @sebastienvas

chxchx:
/assign @yutongz
/assign @sebastienvas
PTAL

chxchx:
@nlandolfi I just checked prow cluster config and was surprised to see that we do mount secrets to pod even during presubmit. You are absolutely right about the security concern. I thought about encrypting on our secrets but then malicious users could just cat the key instead. It boils down to the fact that we faithfully execute whatever source code that is given to us in presubmit and that is dangerous. Will definitely followup on this.

nlandolfi:
yeah totally I think when Laurent and I talked about this we decided that it was for the time ok-ish because we don't run code from people not in the org...automatically at least.

chxchx:
[Thompson K. Reflections on Trusting Trust. _Communications of the ACM_. 1984 Aug 1;27(8):761-3.](https://www.ece.cmu.edu/~ganger/712.fall02/papers/p761-thompson.pdf)

sebastienvas (approved):

Issue #359, opened 2017-08-03
Allow repo-wide code coverage requirement


lookuptable:
Currently the `pkg_check` tool reads per-package code coverage requirement from a local file. If a package does not have an entry in the file, the code coverage requirement is effectively 0 (a.k.a. no tests are needed at all).

It would be better if we can set code coverage requirement for the whole repo, and uses the file to lower the coverage for certain package (e.g. a package only containing a main method).

cc @yutongz @sebastienvas

sebastienvas:
That s a nice add on.

PR #409, opened 2017-08-23
Add default requirement

yutongz:
1. Fix https://github.com/istio/test-infra/issues/359
Add default check for package without requirement

2. Change regex to match new [requirement file](https://github.com/yutongz/mixer/blob/codecov/codecov.requirement)

```release-note
none
```

### DIFF
--- a/toolbox/pkg_check/main.go
+++ b/toolbox/pkg_check/main.go
@@ -184,6 +184,7 @@ func (c *codecovChecker) checkPackageCoverage() (code int) {
 				code = 3 //If no other error code, Error code 3: Failed to upload coverage
 			}
 		}
+
 	}()
 
 	if err := c.parseReport(); err != nil {
--- a/toolbox/pkg_check/main.go
+++ b/toolbox/pkg_check/main.go
@@ -41,11 +41,12 @@ var (
 )
 
 type codecovChecker struct {
-	codeCoverage  map[string]float64
-	report        string
-	requirement   string
-	failedPackage []string
-	bucket        string
+	codeCoverage    map[string]float64
+	codeRequirement map[string]float64
+	report          string
+	requirement     string
+	failedPackage   []string
+	bucket          string
 }
 
 func (c *codecovChecker) parseReport() error {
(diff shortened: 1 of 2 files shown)


toolbox/pkg_check/main.go
```diff
@@ -125,6 +123,21 @@ func (c *codecovChecker) checkRequirement() error {
 	return scanner.Err()
 }
 
+func (c *codecovChecker) checkRequirement() {
+	for pkg, cov := range c.codeCoverage {
+		if req, exist := c.codeRequirement[pkg]; !exist {
+			//There is no entry for this package in requirement file, set default requirement
+			if cov < c.codeRequirement["Default"] {
```
chxchx: Try to make "Default" a constant

toolbox/pkg_check/main.go
```diff
@@ -98,24 +101,19 @@ func (c *codecovChecker) checkRequirement() error {
...
-				re, err := strconv.ParseFloat(m[2], 64)
-				if err != nil {
-					log.Printf("Failed to get requirement for package %s: %s", m[1], m[2])
-					return err
-				}
-				if cov < re {
-					c.failedPackage = append(c.failedPackage, fmt.Sprintf("%s\t%.2f\t%s", m[1], cov, m[2]))
-				}
-
+		if len(m) == 5 || len(m) == 3 {
```
chxchx: There are a lot of hardcoded indices and constants here and in a few other places too, which is not ideal for 2 reasons.

* Any future index change due to a different file format requires one to track down all occurrences of such index and other indices shifted by such change, which is not scalable and prone to bugs.  
* It is not immediately clear what m[2] really means just by looking at it. It needs some more explanatory name/comments

I suggest having a few simple accessor functions, like `getRequirement(m)` instead pf `m[2]`, so that you could keep a single copy of that index 2 in the function (single point of change) and it is more readable.

chxchx (commented):
Just one suggestion and good job yutong!

yutongz:
PTAL

PR #415, opened 2017-08-24
Automatically merge dep_update PR with green CI status

chxchx:
Automation of merging is implemented by adding /lgtm and /approave labels directly to the PR
 
**Release note**:

```release-note
None
```


chxchx:
/assign @sebastienvas
/assign @yutongz 
PTAL

toolbox/util/githubClient.go
```diff
@@ -96,9 +96,76 @@ func (g GithubClient) CreatePullRequest(
...
 		context.Background(), g.owner, repo, &req)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	log.Printf("Created new PR at %s", *pr.HTMLURL)
+	return pr, nil
+}
+
+// MergePullRequestOnceCIGreen polls CI status of a PR periodically
```
yutongz: You don't need to pull pr status and wait for ci green by yourself. What you need to do is add those two labels when you create that kind of pr. Mungegithub will be there checking ci status.

toolbox/util/githubClient.go
```diff
@@ -96,9 +96,76 @@ func (g GithubClient) CreatePullRequest(
...
+	}
+	return nil
+}
+
+// AddlLGMTandApproveLabelsToPR adds /lgtm and /approve labels to a PR,
+// essentially automatically merges the PR without review
+func (g GithubClient) AddlLGMTandApproveLabelsToPR(repo string, pr *github.PullRequest) error {
+	labels := []string{
+		"lgtm",
+		"approve",
```
yutongz: The label is "approved" :)

toolbox/util/githubClient.go
sebastienvas: Yes don't do the merge yourself.

toolbox/util/githubClient.go
```diff
@@ -96,9 +96,76 @@ func (g GithubClient) CreatePullRequest(
...
+		log.Printf("Polling. time elapsed: %s\n", timeElapsedMin)
+		if timeElapsedMin.Nanoseconds() > timeoutMin.Nanoseconds() {
+			return fmt.Errorf("failed to get CI results before timeout")
+		}
+		combinedStatus, _, err := g.client.Repositories.GetCombinedStatus(
+			context.Background(), g.owner, repo, *pr.Head.SHA, nil)
+		if err != nil {
+			return err
+		}
+		switch GetCIState(combinedStatus, nil) {
```
sebastienvas: Can you rename GetCIState, to get GetRequiredCheckStatuses, and use this API to find them https://developer.github.com/v3/repos/branches/#get-required-status-checks-of-protected-branch

toolbox/deps_update/main.go
```diff
@@ -146,7 +146,11 @@ func updateDependenciesOf(repo string) error {
...
 	prTitle := prTitlePrefix + repo
-	return githubClnt.CreatePullRequest(prTitle, prBody, branch, *baseBranch, repo)
+	pr, err := githubClnt.CreatePullRequest(prTitle, prBody, branch, *baseBranch, repo)
+	if err != nil {
+		return err
+	}
+	return githubClnt.MergePullRequestOnceCIGreen(repo, pr)
 }
 
 func main() {
```
sebastienvas: Can you make sure that error and converted in log.Fatalf. That way the binary will return a non 0 exit code, which will failure the prow job. Currently it fails silently.

sebastienvas (changes requested):

sebastienvas (changes requested):
Make sure the binary fails in case of unrecoverable error.

toolbox/util/githubClient.go
chxchx: Done

toolbox/util/githubClient.go
chxchx: You are right! The istio-merge-bot does the check for me. Now I feel stupid.

toolbox/deps_update/main.go
chxchx: Done

toolbox/util/githubClient.go
chxchx: The API you suggested remains in perview stage and using postman to ping it has not shown what we would like. Instead, as @yutongz suggests, since the update_deps job runs every 12 hours, any PR that is still open after 12 hours (aka when the job runs the next time) pretty much fails the presubmit and was not merged. We could safely close them in this case.

toolbox/util/githubClient.go
chxchx: Done

toolbox/util/githubClient.go
chxchx: You are right! istio-merge-bot will check the status for me. Now I feel stupid.

toolbox/deps_update/main.go
chxchx: Done

chxchx:
PTAL

toolbox/util/githubClient.go
```diff
@@ -152,26 +182,9 @@ func (g GithubClient) CloseFailedPullRequests(prTitlePrefix, repo, baseBranch st
 		if !strings.HasPrefix(*pr.Title, prTitlePrefix) {
 			continue
 		}
-		combinedStatus, _, err := g.client.Repositories.GetCombinedStatus(
```
sebastienvas, line 155: Can you only close PR that has been open for say 24 hours. such that if we need to run the command manually we don't accidently close PRs in flight.

chxchx: Done

sebastienvas (approved):

toolbox/deps_update/main.go
```diff
@@ -129,54 +129,61 @@ func updateDependenciesOf(repo string) error {
...
+func main() {
 	if *repo != "" { // only update dependencies of this repo
 		if (*repo == "istio") == (*hub == "") {
-			log.Printf("The hub flag hub must be set for istio/istio and must not for other repos\n")
+			log.Fatalf("The hub flag hub must be set for istio/istio and must not for other repos\n")
 			return
 		}
 		if err := updateDependenciesOf(*repo); err != nil {
-			log.Printf("Failed to udpate dependency: %v\n", err)
+			log.Fatalf("Failed to udpate dependency: %v\n", err)
```
yutongz, line 176: I don't understand why changed back to fatalf? @chxchx

chxchx: Because in calling fatal will have the binary return a nonzero status code to the caller, which is useful in prow so that prow will mark this particular run as failed rather than silently failing.

Issue #396, opened 2017-08-18
Prow failed to create _bazel_bootstrap dir


yutongz:
In some case, prow failed to create the directory '/home/bootstrap/.cache/bazel/_bazel_bootstrap' which skip all tests and exit without error.

`W0818 21:11:19.327] INFO: $TEST_TMPDIR defined: output root default is '/home/bootstrap/.cache/bazel'.
W0818 21:11:19.327] Error: mkdir('/home/bootstrap/.cache/bazel/_bazel_bootstrap'): (error: 13): Permission denied`

nlandolfi:
:( I saw this once or twice -- I am not positive of the cause. The one theory I had was that it might have something to do with the hostport trick we use...

yutongz:
Hi Nick, glad to hear from you! Yes we have the same guess. After we started to run multiple tests on different environment combinations, this issue pops up more than it did before. Currently we increased node number of testing clusters and things are getting better.

ldemailly:
for instance: https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_istio/592/istio-presubmit/412/

which script exactly generates that error message ? can it ls -l on error (or always) so we know why is there a permission denied

ldemailly:
kuat / andra suggested to look into volume claim as a way to ensure assignment of volume

https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolumeclaim

sebastienvas:
I looked at volume claim before, but this is not what we need unless we implement some kind of pool to recycle PVs. Using a volume claim will force us to use PVs, and therefore we would have to create multiple volume claims and then force jobs to use a given volume claim.

The way we do this now, is that we a use the host mount point and once a pod is started we stick to a node port which should prevent other pods from being scheduled, but apparently that does not work.

@spxtr the pods still starts and fails, I think it should be rescheduled in that case right ?

spxtr:
The only time we've seen something similar was when we forgot to specify the HostPort for some of our jobs. Then we got a confusing message saying that the bazel server was shutting down when another job scheduled on the same node. That doesn't sound quite like what you're seeing here.

sebastienvas:
Yeah I double checked, every pod mounting the cache also set the hostPort. One other different is that we run as a bootstrap user, instead of running with root.

ldemailly:
we can also look into sharing the cache out of pod : 
[...]

PR #417, opened 2017-08-25
Explicitly set node selector

sebastienvas:
Prow has 2 node pool. It is possible that some jobs are scheduled on the wrong pool explaining why we experience #396.

```release-note
NONE
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -78,6 +78,8 @@ presubmits:
           limits:
             memory: "20Gi"
             cpu: "5000m"
+      nodeSelector:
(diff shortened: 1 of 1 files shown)


yutongz:
It doesn't seem to explain why job running in e2e-testing clusters also hit this issue.

sebastienvas:
Yes it does. No jobs run on e2e-testing. All the jobs runs inside the Prow cluster. A job can have a test that talk to the e2e-testing cluster, but the framework is actually running inside the Prow cluster.

nlandolfi:
nice catch seb

PR #422, opened 2017-08-27
Separate code coverage from presubmit for pilot

yutongz:
Separate coverage processes (codecov.io, package check) with presubmit:
    1) Speed up presubmit
    2) Coverage will not block presubmit, we can easily make it required or not.

**Release note**:

```release-note
none
```
Connect to https://github.com/istio/pilot/pull/1110

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -553,9 +553,9 @@ presubmits:
         - name: cache-ssd
           hostPath:
             path: /mnt/disks/ssd0
-    - name: pilot-package-coverage
+    - name: pilot-coverage
       agent: kubernetes
-      context: prow/pilot-package-coverage.sh
+      context: prow/pilot-coverage.sh
       always_run: true
      spec:
         containers:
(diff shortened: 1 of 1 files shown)


sebastienvas:
Could you please update the description and give a reason for splitting this?

Also you might want to do this in 2 steps, ie leave the presubmit the way it is such that you don't break it once you push this, split the presubmit on pilot and then, update the config to remove codecov.

yutongz:
I safely remove the "code coverage" part from istio/pilot prow/pilot-presubmit.sh, so I think as long as we merge that pr first, we can keep pilot presubmit green. And I mute pilot-coverage job for now.

P.S. It seems when you "github approve" [pr in pilot](https://github.com/istio/pilot/pull/1110#pullrequestreview-58814741), prow add "lgtm" on this pr since I referred that pr. It's fun and kind of smart, but a bit of confusing. I am recording it.

sebastienvas (approved):

sebastienvas (changes requested):
Actually are you doing another presubmti or another after_success

yutongz:
Another after success, e2e doesn't depend on code coverage so they can work concurrently.

sebastienvas (approved):

yutongz:
hi robot

PR #423, opened 2017-08-27
Allow dependencies update pr self-approves in istio/istio

yutongz:
**Release note**:

```release-note
none
```

### DIFF
--- a/toolbox/util/githubClient.go
+++ b/toolbox/util/githubClient.go
@@ -119,7 +119,8 @@ func (g GithubClient) ApproveAutoPR(repo string, pr *github.PullRequest) error {
 	review, _, err := g.client.PullRequests.CreateReview(context.Background(), g.owner, repo, pr.GetNumber(),
 		&github.PullRequestReviewRequest{})
 	if err != nil {
-		log.Printf("Failed to create a pr review for %s #%d", repo, pr.GetNumber())
+		log.Printf("Failed to create a pr review for %s #%d: %s", repo, pr.GetNumber(), err)
+		return err
 	}
 
 	e := "APPROVE"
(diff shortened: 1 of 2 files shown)


toolbox/util/githubClient.go
```diff
@@ -105,13 +110,31 @@ func (g GithubClient) CreatePullRequest(
...
-		"approved",
-		"release-note-none",
-	}
 	_, _, err := g.client.Issues.AddLabelsToIssue(
-		context.Background(), g.owner, repo, pr.GetNumber(), labels)
+		context.Background(), g.owner, repo, pr.GetNumber(), autoMergeLabels)
+	return err
+}
+
+func (g GithubClient) ApproveAutoPR(repo string, pr *github.PullRequest) error {
```
sebastienvas, line 118: This is cool, but you cannot approve your own PR :). You would need another bot with admin priviledge to approve it. I would advise to create a new label like review-not-required, and auto approve all PRs with that label.

sebastienvas (changes requested):
Separate Approval from current workflow as another bot needs to do that.

toolbox/util/githubClient.go
chxchx: @sebastienvas @yutongz How about we initialize another githubClient right here with the token of another bot/identity so the pr we just created gets approved right away? I am worried about timing issues if we were to have another bot to approve auto PRs asynchronously, since we assume any PR that remains pending/open for an extended period has failed presubmit and we close them after that.

toolbox/util/githubClient.go
```diff
@@ -105,13 +110,31 @@ func (g GithubClient) CreatePullRequest(
...
+	review, _, err := g.client.PullRequests.CreateReview(context.Background(), g.owner, repo, pr.GetNumber(),
+		&github.PullRequestReviewRequest{})
+	if err != nil {
+		log.Printf("Failed to create a pr review for %s #%d: %s", repo, pr.GetNumber(), err)
+		return err
+	}
+
+	e := "APPROVE"
+	b := "Self-approve for auto PR"
```
chxchx, line 127: nit: maybe say "Self-approve without review"

yutongz:
/close
@chxchx will pick it up fix it in a new pr.

PR #433, opened 2017-08-29
Prevent fast forwarding stable to commits other than master or its ancestor

chxchx:
**Release note**:

```release-note
none
```

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -41,6 +41,7 @@ const (
 	downloadScript       = "downloadIstio.sh"
 	istioRepo            = "istio"
 	masterBranch         = "master"
+	stableBranch         = "stable"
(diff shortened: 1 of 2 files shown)


sebastienvas (changes requested):
I think that does not match the requirement. 
Here is postsubmit file
https://github.com/istio/auth/blob/master/prow/auth-postsubmit.sh

The issue is that I want this to be trigered only from master. with the current change if we create a branch of master today, that statement will not change and the postsbumit will still fastforward stable to the release-0.2. SHA. I think we need to find a way to verify that the SHA is a parent of the HEAD of master.

chxchx:
@sebastienvas PTAL

sebastienvas (approved):
Thanks for fixing this.

PR #448, opened 2017-09-01
Release tool: new skip_edit_download_script flag and better approach to add new line before eof

chxchx:
**Release note**:

```release-note
none
```

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -208,7 +208,7 @@ func CreateIstioReleaseUploadArtifacts() error {
 		if err := u.WriteFile(releaseTagFile, *nextRelease); err != nil {
 			return err
 		}
-		if skipEditDownloadScript {
+		if *skipEditDownloadScript {
 			return nil
 		}
 		return u.UpdateKeyValueInFile(
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -24,15 +24,15 @@ import (
 )
 
 var (
-	owner                  = flag.String("owner", "istio", "Github owner or org")
-	tokenFile              = flag.String("token_file", "", "File containing Github API Access Token")
-	op                     = flag.String("op", "", "Operation to be performed")
-	repo                   = flag.String("repo", "", "Repository to which op is applied")
-	baseBranch             = flag.String("base_branch", "", "Branch to which op is applied")
-	refSHA                 = flag.String("ref_sha", "", "Reference commit SHA used to update base branch")
-	nextRelease            = flag.String("next_release", "", "Tag of the next release")
-	skipEditDownloadScript = flag.Bool("skip_edit_download_script", false, "Have download script point to old version")
-	githubClnt             *u.GithubClient
+	owner                = flag.String("owner", "istio", "Github owner or org")
+	tokenFile            = flag.String("token_file", "", "File containing Github API Access Token")
+	op                   = flag.String("op", "", "Operation to be performed")
+	repo                 = flag.String("repo", "", "Repository to which op is applied")
+	baseBranch           = flag.String("base_branch", "", "Branch to which op is applied")
+	refSHA               = flag.String("ref_sha", "", "Reference commit SHA used to update base branch")
+	nextRelease          = flag.String("next_release", "", "Tag of the next release")
+	udpateDownloadScript = flag.Bool("update_download_script", false, "Have download script point to latest version")
+	githubClnt           *u.GithubClient
 )
 
 const (
@@ -195,6 +195,7 @@ func CreateIstioReleaseUploadArtifacts() error {
 		return err
 	}
 	releaseBranch := "finalizeRelease-" + releaseTag
+	prBody := fmt.Sprintf("Finalize release %s on istio", releaseTag)
 	edit := func() error {
 		if _, err := u.Shell(
 			"./release/create_release_archives.sh -d " + releaseBaseDir); err != nil {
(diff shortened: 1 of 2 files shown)


toolbox/githubctl/main.go
```diff
@@ -24,14 +24,15 @@ import (
...
-	nextRelease = flag.String("next_release", "", "Tag of the next release")
-	githubClnt  *u.GithubClient
+	owner                  = flag.String("owner", "istio", "Github owner or org")
+	tokenFile              = flag.String("token_file", "", "File containing Github API Access Token")
+	op                     = flag.String("op", "", "Operation to be performed")
+	repo                   = flag.String("repo", "", "Repository to which op is applied")
+	baseBranch             = flag.String("base_branch", "", "Branch to which op is applied")
+	refSHA                 = flag.String("ref_sha", "", "Reference commit SHA used to update base branch")
+	nextRelease            = flag.String("next_release", "", "Tag of the next release")
+	skipEditDownloadScript = flag.Bool("skip_edit_download_script", false, "Have download script point to old version")
```
ldemailly: I'd make the default to not edit that script.

The idea there is you make the release first - then spend some days making sure it's good/better than the previous one and ready for prime time - then update the download script

so I see that automatic updates as very rare - also that file is special as in there is only 1 so while once we cut the 0.2 branch, we will change istio.RELEASE in the branch, the download change is on master,...

toolbox/util/commonUtils.go
```diff
@@ -36,6 +36,9 @@ func ReadFile(filePath string) (string, error) {
 
 // WriteFile overwrites the file on the given path with content
 func WriteFile(filePath, content string) error {
```
ldemailly: nit: might need to be changed to WriteTextFile if you assume you can add a \n
or are you sure it's never used to write binary data? (I guess it takes a string and not bytes so maybe that's implicit)

toolbox/util/commonUtils.go
```diff
@@ -36,6 +36,9 @@ func ReadFile(filePath string) (string, error) {
...
+	if len(content) == 0 || content[len(content)-1] != '\n' {
```
ldemailly: are you sure you want to change 0 bytes file (which is fine, sometimes marker files are like that) into 1 byte file for the new line ?

toolbox/util/commonUtils.go
```diff
@@ -36,6 +36,9 @@ func ReadFile(filePath string) (string, error) {
...
+		content += "\n"
```
ldemailly: yeah ! fix all the files ! (not perf critical)

toolbox/util/commonUtils.go
chxchx: This is a good point.

toolbox/util/commonUtils.go
chxchx: Renaming to `WriteTextFile` totally makes sense. It is just difficult to know all the places this function has been used and also refactor every such occurrence. Maybe we could leave this text input as implicit defined as you have pointed out, and create a `WriteBinaryFile` when we need it.

toolbox/githubctl/main.go
chxchx: Got it. Thanks for the detailed explanation!

toolbox/util/commonUtils.go
ldemailly: looks like it's used exactly once so far:
https://github.com/istio/test-infra/search?l=Go&q=WriteFile&type=&utf8=%E2%9C%93
so I would rename it now :-)

toolbox/githubctl/main.go
```diff
@@ -24,14 +24,15 @@ import (
...
-	nextRelease = flag.String("next_release", "", "Tag of the next release")
-	githubClnt  *u.GithubClient
+	owner                = flag.String("owner", "istio", "Github owner or org")
+	tokenFile            = flag.String("token_file", "", "File containing Github API Access Token")
+	op                   = flag.String("op", "", "Operation to be performed")
+	repo                 = flag.String("repo", "", "Repository to which op is applied")
+	baseBranch           = flag.String("base_branch", "", "Branch to which op is applied")
+	refSHA               = flag.String("ref_sha", "", "Reference commit SHA used to update base branch")
+	nextRelease          = flag.String("next_release", "", "Tag of the next release")
+	udpateDownloadScript = flag.Bool("update_download_script", false, "Have download script point to latest version")
```
ldemailly: thanks for flipping the flag around!

typo: udpate -> update
probably should turn on the spell checker linter on this repo

toolbox/githubctl/main.go
```diff
@@ -204,16 +206,18 @@ func CreateIstioReleaseUploadArtifacts() error {
...
 		}
-		if err := u.WriteFile(releaseTagFile, *nextRelease+"\n"); err != nil {
+		if err := u.WriteFile(releaseTagFile, *nextRelease); err != nil {
 			return err
 		}
-		return u.UpdateKeyValueInFile(
-			downloadScript, "ISTIO_VERSION", releaseTag)
+		if *udpateDownloadScript {
+			log.Printf("Updating download script with latest release")
+			return u.UpdateKeyValueInFile(
```
ldemailly: why return ?
also probably want to prBody += " and updating download script"

sebastienvas: he is doing a func inside a func.

ldemailly: oic, missed that, thx

toolbox/util/commonUtils.go
chxchx: Done

toolbox/githubctl/main.go

chxchx:
PTAL

ldemailly (approved):
/lgtm
thank you

sebastienvas (approved):

PR #488, opened 2017-09-12
Update pilot presubmit to use rbac cluster

sebastienvas:
Remove cluster for code coverage

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -492,7 +492,7 @@ presubmits:
         - name: service
           mountPath: /etc/service-account
           readOnly: true
-        - name: e2e-testing-kubeconfig
+        - name: pilot-e2e-rbac-kubeconfig
           mountPath: /home/bootstrap/.kube
         - name: cache-ssd
           mountPath: /home/bootstrap/.cache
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -593,8 +593,6 @@ presubmits:
           - name: pilot-codecov-token
             mountPath: /etc/codecov
             readOnly: true
-          - name: e2e-testing-kubeconfig
```
yutongz: You may want to keep this. Pilot needs the existence of this file to pass bazel build before calculate coverage.

prow/config.yaml
```diff
@@ -593,8 +593,6 @@ presubmits:
...
-            mountPath: /home/bootstrap/.kube
```
yutongz, line 597: You may want to keep this. Pilot needs the existence of this file to pass bazel build before calculate coverage.

yutongz (changes requested):

prow/config.yaml
sebastienvas: Code coverage should not need it
https://github.com/istio/pilot/pull/1254

yutongz (approved):

PR #502, opened 2017-09-15
Try saving the logs

sebastienvas:
Saving bazel logs

```release-note
none
```


sebastienvas:
@xiaolanz This needs to be added to all our presubmits :(

xiaolanz (approved):

xiaolanz:
does --test_output=all flag do the job too?

sebastienvas:
Does it forces test to run sequentially ?

xiaolanz:
I don't think so. I can run it with --features=race  flag. not sure if it's what you mean?

sebastienvas:
ok test_output=streamed, is the one that forces test to run sequentially.
Should we use "errors" ? to only get logs on failed tests ?
https://docs.bazel.build/versions/master/command-line-reference.html

xiaolanz:
if we are not restricted by resource, I prefer printing out logs for succeed tests too. It's easier for debugging when we have a easy to access baseline.

PR #512, opened 2017-09-20
Skip rather than abort if tag already exists during release retry

chxchx:
**Release note**:

```release-note
none
```


toolbox/util/githubClient.go
```diff
@@ -214,6 +214,21 @@ func (g GithubClient) ExistBranch(repo, branch string) (bool, error) {
 	return false, nil
 }
 
+// ExistTag checks if a given tag has already existed on remote repo
+func (g GithubClient) ExistTag(repo, tag string) (bool, error) {
+	tags, _, err := g.client.Repositories.ListTags(
```
sebastienvas: Have you looked at https://developer.github.com/v3/git/tags/#get-a-tag

toolbox/githubctl/main.go
```diff
@@ -125,6 +125,13 @@ func TagIstioDepsForRelease() error {
 		if !exists {
 			return fmt.Errorf("ill-defined %s: unable to find %s", istioVersionFile, dep.Name)
 		}
+		if tagExists, err := githubClnt.ExistTag(dep.RepoName, releaseTag); err != nil {
+			log.Printf("Tailed to check if tag [%s] exists on repo %s\n", releaseTag, dep.RepoName)
+			return err
+		} else if tagExists {
```
sebastienvas: don't you want to check if reference the tag points to is the same as the new one ? if is the same, continue, otherwise fail

sebastienvas (changes requested):
PTAL

toolbox/util/githubClient.go
chxchx: Yes the problem is that if the tag does not exist then the go github client returns an error, which is hard to distinguish from other rpc errors like unauthorized, timeout, etc.

chxchx:
PTAL
I switch to a different approach where I check the error message to determine if tag already exists (rather than do this check myself). If tag does exists, I delete it and retag at whatever the given sha is so I don't need to check if sha has changed.

sebastienvas:
You need to change if the SHA has changed. Overwriting tag should never happen in best practice. If people need to overwrite tags they should delete them manually.

chxchx:
PTAL

sebastienvas (approved):

chxchx:
tickle

chxchx:
poke

PR #518, opened 2017-09-25
add code coverage for broker

xiaolanz:
```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -144,6 +144,60 @@ presubmits:
       - name: cache-ssd
         hostPath:
           path: /mnt/disks/ssd0
+  run_after_success:
+  - name: broker-coverage
+    agent: kubernetes
+    context: prow/broker-coverage.sh
+    always_run: true
+    spec:
+      containers:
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -144,6 +144,60 @@ presubmits:
...
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /etc/service-account/service-account.json
+        volumeMounts:
+        - name: service
+          mountPath: /etc/service-account
+          readOnly: true
+        - name: broker-codecov-token
+          mountPath: /etc/codecov
+          readOnly: true
+        - name: e2e-testing-kubeconfig
```
sebastienvas, line 172: Do your test need to run on a cluster ? If not please remove the e2e-testing-kubeconfig volume and volumeMount

sebastienvas (commented):
PTAL

prow/config.yaml
xiaolanz: yes. It runs on cluster for crd configs.

PR #521, opened 2017-09-26
Update downloadIstioCandidate.sh when finalizing release

chxchx:
**Release note**:

```release-note
none
```

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -38,7 +38,7 @@ const (
 	istioVersionFile     = "istio.VERSION"
 	istioDepsFile        = "istio.deps"
 	releaseTagFile       = "istio.RELEASE"
-	downloadScript       = "downloadIstioCandidate.sh"
+	downloadScript       = "release/downloadIstioCandidate.sh"
 	istioRepo            = "istio"
 	masterBranch         = "master"
 	export               = "export "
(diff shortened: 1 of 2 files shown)


toolbox/githubctl/main.go
```diff
@@ -24,29 +24,28 @@ import (
...
+	nextRelease = flag.String("next_release", "", "Tag of the next release")
+	githubClnt  *u.GithubClient
 )
 
 const (
 	istioVersionFile     = "istio.VERSION"
 	istioDepsFile        = "istio.deps"
 	releaseTagFile       = "istio.RELEASE"
-	downloadScript       = "downloadIstio.sh"
+	downloadScript       = "downloadIstioCandidate.sh"
```
ldemailly: don't you need it to be `release/downloadIstioCandidate.sh` ?
(I put the new script under `release/` to limit the chance of accidental changes to it - we can't move the other one as the URL is burned)

chxchx: You are right. I didn't notice the two download scripts are in different directories.

ldemailly (approved):
I guess it'll get tested for 0.2.5 shortly ?
otherwise how do you test it ?

chxchx:
I tested it with my forked repos. An example PR is here https://github.com/chxchx/istio/pull/35/files.

chxchx:
I tested it with my forked repos. An example PR created by the script is here https://github.com/chxchx/istio/pull/36/files

sebastienvas:
@ldemailly Instead of updating that script for each release, can we just update script to take the latest - like here:
https://github.com/istio/istio/pull/409/files

ldemailly:
@sebastienvas that would slow down the download by adding an extra step for every user running it, if we already modify files when finalizing I think modifying that one is good too

PR #522, opened 2017-09-26
Change cluster default name and add more configurable

yutongz:
**Release note**:

```release-note
none
```
Cluster name changed by https://github.com/istio/istio/pull/902/files#diff-4d4b343aca91a15218a827b3b0385f10R36

### DIFF
--- a/prow/test-infra-cleanup-GKE.sh
+++ b/prow/test-infra-cleanup-GKE.sh
@@ -27,4 +27,4 @@ set -u
 set -x
 
 echo "=== Cleaning up GKE clusters ==="
-./scripts/cleanup-GKE "${@}"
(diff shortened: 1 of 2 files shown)


prow/test-infra-cleanup-GKE.sh
```diff
@@ -27,4 +27,4 @@ set -u
 set -x
 
 echo "=== Cleaning up GKE clusters ==="
-./scripts/cleanup-GKE
+./scripts/cleanup-GKE "${@}"
```
sebastienvas: we discovered that ${@} and $@ are not equivalent. plus use $@.

sebastienvas (approved):

PR #526, opened 2017-09-28
Skips release-note check for Release PR.

sebastienvas:
We had to manually to this, so making it easier for next release engineer

```release-note
none
```

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -173,9 +173,12 @@ func cloneIstioMakePR(newBranch, prTitle, prBody string, edit func() error) erro
 		newBranch, newBranch); err != nil {
 		return err
 	}
-	_, err = githubClnt.CreatePullRequest(
+	pr, err := githubClnt.CreatePullRequest(
 		prTitle, prBody, newBranch, *baseBranch, istioRepo)
-	return err
+	if err != nil {
+		return err
+	}
+	return githubClnt.AddlabelsToPR(istioRepo, pr, "release-note-none")
 }
 
 // UpdateIstioVersionAfterReleaseTagsMadeOnDeps runs updateVersion.sh to update


toolbox/githubctl/main.go
```diff
@@ -173,9 +173,12 @@ func cloneIstioMakePR(newBranch, prTitle, prBody string, edit func() error) erro
...
 		return err
 	}
-	_, err = githubClnt.CreatePullRequest(
+	pr, err := githubClnt.CreatePullRequest(
 		prTitle, prBody, newBranch, *baseBranch, istioRepo)
-	return err
+	if err != nil {
+		return err
+	}
+	return githubClnt.AddlabelsToPR(istioRepo, pr, "release-note-none")
```
ldemailly, line 181: can't the label be put first / vs added later ?
or put  '''release-note NONE in the text ?

ldemailly (approved):

toolbox/githubctl/main.go
chxchx: The PR has to be created before we could add a label, right?
And putting "release-note NONE" in PR description essentially lets the bot add "release-note-none" label to the PR and we could just do it directly here.

ldemailly: ah interesting, I was expecting github api to have the option to create a pr with labels

so you add all the lgtm, cla etc labels 1 by 1 on the auto prs ?

chxchx: The github api takes a list of labels to be applied to PR. I structured the client so that you can do `githubClnt.AddlabelsToPR(istioRepo, pr, "labelOne", "labelTwo", "labelThree")`. Maybe one could create a PR with labels directly, but perhaps having two separated util functions makes it more flexible.

PR #534, opened 2017-10-05
Make sure release-note is none

yutongz:
**Release note**:

```release-note
none
```

https://github.com/istio/pilot/pull/1402

### DIFF
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -34,8 +34,7 @@ var (
 const (
 	istioDepsFile         = "istio.deps"
 	prTitlePrefix         = "[DO NOT MERGE] Auto PR to update dependencies of "
-	prBody                = "This PR will be merged automatically once checks are successful." +
-		"\r\n```release-note\r\nNONE\r\n```"
+	prBody                = "This PR will be merged automatically once checks are successful."
 	dependencyUpdateLabel = "dependency-update"
 
 	// CI Artifacts URLs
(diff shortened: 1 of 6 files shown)


toolbox/deps_update/main.go
```diff
@@ -34,7 +34,8 @@ var (
 const (
 	istioDepsFile         = "istio.deps"
 	prTitlePrefix         = "[DO NOT MERGE] Auto PR to update dependencies of "
-	prBody                = "This PR will be merged automatically once checks are successful."
+	prBody                = "This PR will be merged automatically once checks are successful." +
+		"\r\n```release-note\r\nNONE\r\n```"
```
chxchx: Perhaps define the release-note none as a constant string in CommonUtil.go

yutongz: You are right. I add a param in CreatePullRequest() to handle release-note, leaving it empty will make it "none"

toolbox/util/githubClient.go
```diff
@@ -126,7 +126,11 @@ func (g GithubClient) Remote(repo string) string {
 
 // CreatePullRequest within :repo from :branch to :baseBranch
 func (g GithubClient) CreatePullRequest(
-	title, body, branch, baseBranch, repo string) (*github.PullRequest, error) {
+	title, body, releaseNote, branch, baseBranch, repo string) (*github.PullRequest, error) {
```
chxchx: I don't think changing the function signature is a good idea since this function has been used at multiple places and for each occurrence you need to update the params. A better approach would be function overloading, which defines another CreatePullRequest() with an additional param `releaseNote`. From line 130, you could then call the overloaded CreatePullRequest() with `releaseNote = none`

chxchx: It is also not clear what empty string means when reading the code. I hope my comment makes sense.

yutongz: 1. Do you know someone are using this function outside test-infra? I think I changed every usages in test-infra. For us, most of auto PR needs to have "release-note" part. Actually now in istio, all PR has to have that part so I believe it's a good time to upgrade this function. 
2. Right, I am adding comment to make it clearer.

PR #547, opened 2017-10-12
Update all jobs to 0.3.1 ie bazel 0.5.4

sebastienvas:
People have been testing with bazel 0.5.4 for quite sometime so this is safe.
Updating to 0.6.0 will likely fail and we need to move one repo at a time.

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -88,7 +88,7 @@ presubmits:
     - release-0.2
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


sebastienvas:
@ayj Please watch for some issue in the build. We don't expect any but this is a warning.

ayj (approved):

PR #553, opened 2017-10-17
Remove release-0.1 and create jobs for release-0.2

sebastienvas:
Removing support for 0.1 as we won't be created patch release for it.

Created specific jobs for 0.2 as we updated bazel in master which is incompatible with 0.2

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -27,6 +27,7 @@ presubmits:
   #                  Otherwise, run a Jenkins job.
 
   istio/api:
+
   - name: api-presubmit
     agent: kubernetes
     context: prow/api-presubmit.sh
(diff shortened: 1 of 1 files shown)


yutongz (approved):
/lgtm

prow/config.yaml
```diff
@@ -136,7 +137,67 @@ presubmits:
         hostPath:
           path: /mnt/disks/ssd0
 
+  - name: auth-presubmit
```
ayj, line 140: Does `name` need to be unique? How do we differentiate jobs on prow.istio.io?

sebastienvas: job name and script name are related. You can differentiate them in prow.istio.io

PR #564, opened 2017-10-25
Release-note collector uses tag creation time as the cut off time

### LINKED ISSUE #563: Release notes are missing due to the wrong cut off time
We are using the creation time of the commit SHA of a release tag as the cut off time, which is wrong. A PR is always merged after the creation of its last SHA, which means this specific PR is opt out from release-note candidate.

An example is:
If we try to generate release-note from 0.2.7 to 0.2.9

* The commit of auth 0.2.9 is `0a8c616` https://github.com/istio/auth/releases
* The corresponding  [PR](https://github.com/istio/istio/pull/1137) has release-note
* If we use the time of the commit as the upper bound (0.2.9), we are missing release-note from this PR because it was merged afterwards

yutongz:
**Release note**:

```release-note
none
```

fixes https://github.com/istio/test-infra/issues/563

Instead of commit time, use creation time of release tag(e.g. 0.1.6, 0.2.7) as the cut off time

### DIFF
--- a/toolbox/metrics/buildFreshness/buildFreshness.go
+++ b/toolbox/metrics/buildFreshness/buildFreshness.go
@@ -35,19 +35,19 @@ type DepFreshness struct {
 	Age time.Duration
 }
 
-func getBranchHeadTime(githubClnt *u.GithubClient, repo, branch string) (*time.Time, error) {
+func getBranchHeadTime(githubClnt *u.GithubClient, repo, branch string) (time.Time, error) {
 	sha, err := githubClnt.GetHeadCommitSHA(repo, branch)
 	if err != nil {
-		return nil, err
+		return time.Time{}, err
 	}
 	t, err := githubClnt.GetCommitCreationTime(repo, sha)
 	if err != nil {
-		return nil, err
+		return time.Time{}, err
 	}
 	return t, nil
 }
 
-func getCommitCreationTimeByRef(githubClnt *u.GithubClient, repo, ref string) (*time.Time, error) {
+func getCommitCreationTimeByRef(githubClnt *u.GithubClient, repo, ref string) (time.Time, error) {
 	if u.SHARegex.MatchString(ref) {
 		return githubClnt.GetCommitCreationTime(repo, ref)
 	} else if u.ReleaseTagRegex.MatchString(ref) {
@@ -55,7 +55,7 @@ func getCommitCreationTimeByRef(githubClnt *u.GithubClient, repo, ref string) (*
 	}
 	err := fmt.Errorf(
 		"reference must be a SHA or release tag to get creation time, but was instead %s", ref)
-	return nil, err
+	return time.Time{}, err
 }
 
 func getAgeMetric(githubClnt *u.GithubClient, dep *u.Dependency) (*DepFreshness, error) {
@@ -73,7 +73,7 @@ func getAgeMetric(githubClnt *u.GithubClient, dep *u.Dependency) (*DepFreshness,
 			dep.ProdBranch, dep.RepoName, err)
 		return nil, e
 	}
-	lag := latestTime.Sub(*stableTime)
+	lag := latestTime.Sub(stableTime)
 	return &DepFreshness{*dep, lag}, nil
 }
 
--- a/toolbox/util/githubClient.go
+++ b/toolbox/util/githubClient.go
@@ -252,7 +252,7 @@ func (g GithubClient) CloseIdlePullRequests(prTitlePrefix, repo, baseBranch stri
 	return multiErr
 }
 
-// GetReferenceSHAAndType finds the SHA of the commit to which the HEAD of branch points
+// GetHeadCommitSHA finds the SHA of the commit to which the HEAD of branch points
 func (g GithubClient) GetHeadCommitSHA(repo, branch string) (string, error) {
 	sha, _, err := g.GetReferenceSHAAndType(repo, "refs/heads/"+branch)
 	return sha, err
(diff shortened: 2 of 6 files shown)


toolbox/util/githubClient.go
```diff
@@ -284,26 +284,69 @@ func (g GithubClient) GetTagCommitSHA(repo, tag string) (string, error) {
...
+	if err != nil {
+		return time.Time{}, fmt.Errorf("failed to to get release tag: %s", err)
+	}
+	return release.GetCreatedAt().Time, nil
+}
+
+// getannotatedTagCreationTime gets the creation time of an annotated Tag
+// Release tags except those from istio/istio are annotated tag.
+func (g GithubClient) getannotatedTagCreationTime(repo, tag string) (time.Time, error) {
+	sha, ty, err := g.GetReference(repo, "refs/tags/"+tag)
```
chxchx: This is tangent to this PR but it would be nice to rename `GetReference` to `GetReferenceSHAAndType` (and the function comments) so there is no ambiguous in what it returns (ref oibject vs. sha and ty)

toolbox/util/githubClient.go
```diff
@@ -284,26 +284,69 @@ func (g GithubClient) GetTagCommitSHA(repo, tag string) (string, error) {
...
 	commitSHA, err := g.GetTagCommitSHA(repo, tag)
 	if err != nil {
-		return nil, err
+		return time.Time{}, err
 	}
 	return g.GetCommitCreationTime(repo, commitSHA)
 }
 
+// GetReleaseCreationTime gets the creation time of a release tag (0.1.6, 0.2.7)
+func (g GithubClient) GetReleaseCreationTime(repo, tag string) (createTime time.Time, err error) {
```
chxchx: You put `(createTime time.Time, err error) ` as return types. Does that declare those vars implicitly for you? cause I also notice createTime and err are assigned values directly as in `createTime, err = g.getReleaseTagCreationTime(repo, tag)`

chxchx: Expose g.getReleaseTagCreationTime(repo, tag) and g.getannotatedTagCreationTime(repo, tag) directly and move this function to your script. It depends on the current release logic (likely changed in Q4) and is not consider general purpose and thus should not be put in utils.

toolbox/util/githubClient.go
```diff
@@ -284,26 +284,69 @@ func (g GithubClient) GetTagCommitSHA(repo, tag string) (string, error) {
...
+}
+
+// getannotatedTagCreationTime gets the creation time of an annotated Tag
+// Release tags except those from istio/istio are annotated tag.
+func (g GithubClient) getannotatedTagCreationTime(repo, tag string) (time.Time, error) {
+	sha, ty, err := g.GetReference(repo, "refs/tags/"+tag)
+	if err != nil {
+		return time.Time{}, err
+	}
+
```
chxchx: remove new line

toolbox/util/githubClient.go
```diff
@@ -284,26 +284,69 @@ func (g GithubClient) GetTagCommitSHA(repo, tag string) (string, error) {
...
+	if ty != "tag" {
+		return time.Time{}, fmt.Errorf("refs/tags/%s is not pointing to a tag object", tag)
+	}
+
```
chxchx, line 328: remove new line

toolbox/metrics/buildFreshness/buildFreshness.go
```diff
@@ -35,27 +35,27 @@ type DepFreshness struct {
 	Age time.Duration
 }
 
-func getBranchHeadTime(githubClnt *u.GithubClient, repo, branch string) (*time.Time, error) {
+func getBranchHeadTime(githubClnt *u.GithubClient, repo, branch string) (time.Time, error) {
 	sha, err := githubClnt.GetHeadCommitSHA(repo, branch)
 	if err != nil {
-		return nil, err
+		return time.Time{}, err
```
chxchx, line 41: Is there a reason why we want to switch to returning stucts from returning pointers?
Returning structs basically copies the entire struct from the stack frame of the current function to its caller. Performance aside, finding every occurrence of this functions (various of others also being refactored) can be challenging and prone to bugs. Perhaps running `bazel build //...` helps catching some of them.
I am not sure about this one. Perhaps @sebastienvas may have a better take at it.

toolbox/util/githubClient.go
yutongz: Yes, we can do that.

toolbox/metrics/buildFreshness/buildFreshness.go
yutongz: The idea comes from using method `Getxxx()` instead of directly access a member. I think that's the safer way get value. But I am kind of curious about why they still make those member public though.
I think it's ok for us to change api, we don't have code depending on this outside of test-infra right?

toolbox/util/githubClient.go
yutongz: Make sense!

toolbox/util/githubClient.go
yutongz: Actually I want to keep those new lines as separators to keep it clean.

toolbox/metrics/buildFreshness/buildFreshness.go
sebastienvas: who are they?

sebastienvas: I don t think performance is an issue here. You must return a value when you don't want it to be modified. When I was working with this github library I had to redo my function signature multiple time. I think using value is fine

yutongz: They: github golang api

yutongz:
PTAL

Issue #556, opened 2017-10-20
Updating version in downloadIstioCandidate.sh during release should be configurable 


yutongz:
As we discussed in https://github.com/istio/istio/pull/1191. The release tool should have a flag to decide if we need to update version in download-script in master branch.

yutongz:
/assign @chxchx 
Do you want to work on this fix?

chxchx:
The release process is being revamped and we are going to start pre-release on 0.3.* soon. It guess it is a P2 for now.

PR #565, opened 2017-10-30
Add flag to enable update download version in extra branches when finalize release

yutongz:
**Release note**:

```release-note
none
```

With extra flag `--extra_branches=master`, in step 5 in [release instruction](https://github.com/istio/istio/tree/master/release#semi-automated-release-since-02), helper will update version in [release/downloadIstioCandidate.sh](https://github.com/istio/istio/blob/release-0.2/release/downloadIstioCandidate.sh) in master branch. It's optional and decided by engineers making the specific release.

Will update release instruction in istio/istio after repo re-organized process.

Fixes https://github.com/istio/test-infra/issues/556
Related PRs: 
https://github.com/istio/istio/pull/1191
https://github.com/istio/istio/pull/1272


yutongz:
cc @ayj @linsun

toolbox/githubctl/main.go
```diff
@@ -147,30 +149,7 @@ func TagIstioDepsForRelease() error {
 
 func cloneIstioMakePR(newBranch, prTitle, prBody string, edit func() error) error {
```
chxchx: Maybe just remove this function entirely and replace with `githubClnt.CreatePRUpdateRepo()`?

toolbox/githubctl/main.go
```diff
@@ -24,14 +24,16 @@ import (
...
-	nextRelease = flag.String("next_release", "", "Tag of the next release")
-	githubClnt  *u.GithubClient
+	owner                              = flag.String("owner", "istio", "Github owner or org")
+	tokenFile                          = flag.String("token_file", "", "File containing Github API Access Token")
+	op                                 = flag.String("op", "", "Operation to be performed")
+	repo                               = flag.String("repo", "", "Repository to which op is applied")
+	baseBranch                         = flag.String("base_branch", "", "Branch to which op is applied")
+	refSHA                             = flag.String("ref_sha", "", "Reference commit SHA used to update base branch")
+	nextRelease                        = flag.String("next_release", "", "Tag of the next release")
+	extraBranchesUpdateDownloadVersion = flag.String("extra_branches", "",
```
chxchx: Consider rename `extra_branches` to `update_rel_branches`

chxchx (commented):
Thank you yutong for making this change!

ldemailly:
thx ! I think we only ever need to update the dowloader in master (at least as of today) as that's where the vanity URL points to (so maybe instead of extra flag you can always make 2 PR when not on master and 1 PR when on master ?)

yutongz:
@ldemailly Yeah, I think it makes sense to not require extra changes for releasers. Well I am not sure if it's a better idea or necessary to push every patch release to master to reach customers?

ldemailly:
I sort of take it back, let's say we have a 0.3.2 and then someone makes a patch 0.2.11 maybe they shouldn't overwrite the "latest" that should stay 0.3.2 and maybe after some testing only the 0.2.11 can be updated in the stable one

yutongz:
That's a great example. So, let's keep it optional for now and let releaser decide if it's necessary to update master.

ldemailly:
sounds good ! ship it ! (with maybe an update to the release doc though)

yutongz:
/approve
I am not sure if this is a good time to make a PR in istio/istio, they are pretty against PRs right now, because of all the repo re-organize process.

PR #567, opened 2017-10-31
Skip auto dependency update until migration to mono repo is done

chxchx:
**Release note**:

```release-note
none
```


yutongz:
Let's keep it now.

PR #576, opened 2017-11-03
Increase build timeout limit after mono repo reorg

chxchx:
Previous timeout limit was set before mono repo structure, which is no longer enough and causes builds to abort prematurely as seen in https://github.com/istio/istio/pull/1405

**Release note**:

```release-note
none
```


mandarjog (approved):
lgtm

PR #595, opened 2017-11-10
Move go vendors to their own files

sebastienvas:
golang deps is pretty cool to find the right version of deps to use. I am moving the go vendor to their own file, such that I can work on a way to generate it from golang deps output.

```release-note
none
```


sebastienvas:
@costinm you might be interested by this. I created a script that generates the go vendor for bazel based on dep Gopkg.lock.

PR #596, opened 2017-11-14
Release qualification trigger script

chxchx:
The daily repo is created here https://github.com/istio-releases/daily-release. Will set up gating test jobs and prow. 

**Release note**:

```release-note
none
```


sebastienvas:
This is coming big enough, that I would like to see unit test for that code.

chxchx:
@sebastienvas Is there a unit test example in test-infra that I can look at? This code depends heavily on GitHub API and I was not sure if unit tests are a good fit in this case.

sebastienvas:
@chxchx There should be ways to abstract github api calls such that can verify your code behavior. Let's not block this PR on it, but let's think about how to do that for next check in.

yutongz:
/lgtm

Gone just merge it now so that we can have a official document about this part.

yutongz:
Created a issue to track unit test effort:  https://github.com/istio/test-infra/issues/623

yutongz (approved):

PR #597, opened 2017-11-14
Set up daily release qualification jobs

chxchx:
Jobs are triggered when PRs are made on this repo https://github.com/istio-releases/daily-release

**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -883,7 +883,7 @@ presubmits:
         - name: service
           mountPath: /etc/service-account
           readOnly: true
-        - name: istio-e2e-rbac-kubeconfig
+        - name: daily-release-e2e-rbac-kubeconfig
           mountPath: /home/bootstrap/.kube
         - name: cache-ssd
           mountPath: /home/bootstrap/.cache
@@ -903,9 +903,9 @@ presubmits:
       - name: service
         secret:
           secretName: service-account
-      - name: istio-e2e-rbac-kubeconfig
+      - name: daily-release-e2e-rbac-kubeconfig
         secret:
-          secretName: istio-e2e-rbac-kubeconfig
+          secretName: daily-release-e2e-rbac-kubeconfig
       - name: cache-ssd
         hostPath:
           path: /mnt/disks/ssd0
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -856,6 +856,59 @@ presubmits:
         hostPath:
           path: /mnt/disks/ssd0
 
+  istio-releases/daily-release:
+
+  - name: daily-release-qualification
+    agent: kubernetes
+    context: prow/qualify.sh
+    always_run: true
+    rerun_command: "/test daily-release-qualification"
+    trigger: "(?m)^/(retest|test (all|daily-release-qualification))?,?(\\s+|$)"
+    branches:
+    - master
+    spec:
+      containers:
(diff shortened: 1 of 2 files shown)


sebastienvas (approved):

prow/config.yaml
```diff
@@ -856,6 +856,59 @@ presubmits:
         hostPath:
           path: /mnt/disks/ssd0
 
+  istio-releases/daily-release:
+
+  - name: daily-release-qualification
+    agent: kubernetes
+    context: prow/qualify.sh
```
sebastienvas: currently the name of job needs to match the prow script

sebastienvas (changes requested):
Name of the job needs to match the prow script

prow/config.yaml
chxchx: Got it

prow/plugins.yaml
```diff
@@ -12,6 +12,7 @@ triggers:
   - istio/old_mixer_repo
   - istio/old_pilot_repo
   - istio/test-infra
+  - istio-releases/daily-release
   trusted_org: istio
```
chxchx, line 16: @sebastienvas Should I add istio-releases to the trusted_org here?

chxchx:
PTAL

prow/plugins.yaml
```diff
@@ -82,3 +83,12 @@ plugins:
   - lgtm
   - hold
   - yuks
+
+  istio-releases/daily-release:
+  - assign
```
yutongz: If we are not using auto system in this repo, then only trigger would be enough, we don't want human interaction around.

prow/config.yaml
```diff
@@ -856,6 +856,59 @@ presubmits:
...
+          limits:
+            memory: "20Gi"
+            cpu: "5000m"
+      nodeSelector:
+        cloud.google.com/gke-nodepool: build-pool
+      volumes:
+      - name: service
+        secret:
+          secretName: service-account
+      - name: istio-e2e-rbac-kubeconfig
```
yutongz: Maybe create a new cluster for release qualification.
You can use this script https://github.com/istio/test-infra/tree/master/scripts#update-e2e-clustersh to create a new cluster and create a secret into prow.
Tip: You need to remove this line when you run https://github.com/istio/test-infra/blob/master/scripts/update-e2e-cluster.sh#L96 since there is not such secret in Prow before.

prow/plugins.yaml
sebastienvas: trusted org, means we trust the user from this org. Since we are part of both orgs, this should not be required.

prow/plugins.yaml
sebastienvas: Yeah assign, release-note, hold should not be required. I think we would still be merging on success right ?

chxchx: I talked to yutong and we believe calling github API to merge a successful PR is a better approach in this case

prow/config.yaml
sebastienvas: Yes we should do that.

chxchx: Done. Please help double check my code here as well as https://pantheon.corp.google.com/kubernetes/list?project=istio-testing&organizationId=433637338589 to make sure we did not miss anything. Thanks!

PR #599, opened 2017-11-16
Adds an after success run for e2e pilot

sebastienvas:
-->
```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -254,6 +254,51 @@ presubmits:
         hostPath:
           path: /mnt/disks/ssd0
     run_after_success:
+    - name: istio-pilot-e2e
+      agent: kubernetes
+      context: e2e-smoke.sh
+      rerun_command: "/test istio-pilot-e2e"
+      trigger: "(?m)^/test (istio-pilot-e2e)?,?(\\s+|$)"
+      skip_report: true
+      max_concurrency: 8
+      spec:
+        containers:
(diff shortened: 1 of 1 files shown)


guptasu (approved):

prow/config.yaml
```diff
@@ -254,6 +254,51 @@ presubmits:
         hostPath:
           path: /mnt/disks/ssd0
     run_after_success:
+    - name: istio-pilot-e2e
+      agent: kubernetes
+      context: e2e-smoke.sh
```
yutongz: Change this?

prow/config.yaml
```diff
@@ -254,6 +254,51 @@ presubmits:
...
+            hostPort: 9998
+          resources:
+            requests:
+              memory: "512Mi"
+              cpu: "500m"
+            limits:
+              memory: "20Gi"
+              cpu: "5000m"
+        nodeSelector:
+          cloud.google.com/gke-nodepool: build-pool
```
yutongz, line 294: You need to use "new-cluster-pool", which doesn't have kubeconfig mounted

yutongz (changes requested):

prow/config.yaml
sebastienvas: Done

prow/config.yaml
sebastienvas: The kubeconfig is mounted on the pod not on the pool.

yutongz (approved):

PR #608, opened 2017-11-17
Oncall notification and block auto-merge if postsubmit failing

yutongz:
**Release note**:

```release-note
none
```

There is already a running server with this tool. We can review this PR, make improvements and update server at same time.

This tool scans latest status of each Post Submit job (istio-postsubmit and three e2e tests) in a periodical circle. First get latest run number in `latest-build.txt` and based on `finished.json` we can figure if test passed or failed.
[artifacts](https://pantheon.corp.google.com/storage/browser/istio-prow/istio-postsubmit/?project=istio-testing)

* If any **new** failures appeared in this circle, a email with detailed information will be sent to "istio-oncall" group mailing list.

* If latest status any test is failed, try to block auto-merge by adding "do-not-merge/post-submit" to PRs again protected banch(master) in protected repo(istio/istio), otherwise, remove all "do-not-merge/post-submit" labels from PRs.

Simple design doc: https://goo.gl/vmYhVt

PR #612, opened 2017-11-17
Upgrade Mungegithub

yutongz:
**Release note**:

```release-note
none
```

* Upgrade mungegithub
* Add Makefile
* Add "do-not-merge/post-submit" label in config to match requirement in https://github.com/istio/test-infra/pull/608


yutongz:
cc @chxchx  FYI

PR #614, opened 2017-11-21
Upgrade Golang version. New changes in Mixer needs latest Golang

guptasu:
Currently it works because we are using bazel build and that installs 1.9 version of go,
However, for other steps (doing go build and go test outside bazel) we need latest version of go installed on the test machine.

**Release note**:

```release-note
none
```

### DIFF
--- a/scripts/tools/linux-install-golang
+++ b/scripts/tools/linux-install-golang
@@ -8,7 +8,7 @@ fi
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 . ${DIR}/all-utilities || { echo "Cannot load Bash utilities" ; exit 1 ; }
 
-GO_VERSION='1.8'
+GO_VERSION='1.9'
 GO_BASE_URL="https://storage.googleapis.com/golang"
 GO_ARCHIVE="go${GO_VERSION}.linux-amd64.tar.gz"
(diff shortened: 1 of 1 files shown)


ldemailly:
which mixer change required 1.9 and how was that switch decided ? did we compare/test 1.8 and 1.9 to make sure there are no issues or perf regression ?

guptasu:
@ldemailly We moved to latest version of bazel (I think few months back), which has Go 1.9 as part of it. We have been using that for some time IMO. However, prow machines were built with old version of `go` and only with the latest change in Mixer, this stale prow machine version issue came up.

guptasu:
@sebastienvas @yutongz when will the test infra machine pick the latest go version.. I still see old version of Go which causes the error in prow logs for mixer build : ```W1121 00:01:20.620] mixer/pkg/cache/ttlCache.go:33: undefined: sync.Map```

logs: https://prow.istio.io/log?job=istio-presubmit&id=410

ldemailly:
Bazel version doesn't by itself force a go version afaik
We should probably set it rather than get a random default

sebastienvas:
bazel and go should be usging the same version. Like @ldemailly said, we should update https://github.com/istio/istio/blob/master/bin/use_bazel_go.sh to pick the version that we expect and fail otherwise and use the same version for golang build.

yutongz:
Change takes effect after https://github.com/istio/test-infra/pull/639

PR #617, opened 2017-11-22
Enable prow for mixerclient and proxy

rkpagadala:
**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -30,6 +30,114 @@ presubmits:
   #   spec:          If this exists then run a kubernetes pod with this spec.
   #                  Otherwise, run a Jenkins job.
 
+  istio/mixerclient:
+
+  - name: mixerclient-presubmit
+    agent: kubernetes
+    context: prow/mixerclient-presubmit.sh
+    always_run: true
+    rerun_command: "/test mixerclient-presubmit"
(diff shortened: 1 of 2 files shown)


prow/config.yaml
```diff
@@ -30,6 +30,114 @@ presubmits:
...
+          limits:
+            memory: "20Gi"
+            cpu: "5000m"
+      nodeSelector:
+        cloud.google.com/gke-nodepool: build-pool
+      volumes:
+      - name: service
+        secret:
+          secretName: service-account
+      - name: istio-e2e-rbac-kubeconfig
```
sebastienvas: You can remove this volume and associated volume moun as there is not e2e test.

prow/config.yaml
```diff
@@ -30,6 +30,114 @@ presubmits:
...
```
sebastienvas: Same comment here.

prow/config.yaml
```diff
@@ -1344,6 +1452,60 @@ postsubmits:
...
```
sebastienvas: Same comment here.

sebastienvas (changes requested):
Thanks. A few minor changes.

sebastienvas (approved):

PR #622, opened 2017-11-28
add no auth release qualification job with default proxy

chxchx:
Will add more tests with different combination of auth, clusterwide, default_proxy in subsequent PRs but would like to do so incrementally.

**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -999,12 +999,12 @@ presubmits:
 
   istio-releases/daily-release:
 
-  - name: daily-release-qualification
+  - name: daily-e2e-rbac-no_auth
     agent: kubernetes
-    context: prow/daily-release-qualification.sh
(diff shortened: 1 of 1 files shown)


sebastienvas (approved):

prow/config.yaml
```diff
@@ -1050,6 +1050,57 @@ presubmits:
       - name: cache-ssd
         hostPath:
           path: /mnt/disks/ssd0
+  - name: daily-e2e-rbac-no_auth-default
```
yutongz: I don't think we can put two presubmits like this, we need to use run after succeeded.

Also change the name "daily-release-qualification" maybe.

yutongz: I take it back, we can do this: https://github.com/kubernetes/test-infra/blob/master/prow/config.yaml#L235

PR #624, opened 2017-11-29
Add more release qualification tests

chxchx:
Job scripts can be found in https://github.com/istio-releases/daily-release/tree/master/prow

**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1205,6 +1205,11 @@ presubmits:
           path: /mnt/disks/ssd0
   - name: daily-e2e-cluster_wide-auth
     agent: kubernetes
+    context: prow/daily-e2e-cluster_wide-auth.sh
+    always_run: true
+    trigger: "(?m)^/(retest|test (all|daily-e2e-cluster_wide-auth))?,?(\\s+|$)"
+    branches:
+    - master
     spec:
       containers:
       - image: gcr.io/istio-testing/prowbazel:0.3.2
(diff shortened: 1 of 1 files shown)


sebastienvas (approved):

prow/config.yaml
```diff
@@ -1101,6 +1101,188 @@ presubmits:
...
+          secretName: service-account
+      - name: daily-release-e2e-rbac-kubeconfig
+        secret:
+          secretName: daily-release-e2e-rbac-kubeconfig
+      - name: cache-ssd
+        hostPath:
+          path: /mnt/disks/ssd0
+  - name: daily-e2e-cluster_wide-auth
+    agent: kubernetes
+    spec:
```
yutongz: I think we need to add "always_run: true" here. This is part of presubmit.
Also "context" which will show on github and "branch" master

prow/config.yaml
```diff
@@ -1101,6 +1101,188 @@ presubmits:
...
+      volumes:
+      - name: service
+        secret:
+          secretName: service-account
+      - name: cache-ssd
+        hostPath:
+          path: /mnt/disks/ssd0
+  - name: daily-e2e-cluster_wide-auth-default
+    agent: kubernetes
+    spec:
```
yutongz: Same here.

prow/config.yaml
yutongz: And also add "rerun_command" and "trigger" like jobs before, although we probably won't use them in release qualification. Just for consistency.

chxchx:
PTAL

PR #625, opened 2017-12-01
enable update bot for istio

rkpagadala:
**Release note**:

```release-note
none
```

### DIFF
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -82,12 +82,12 @@ func generateArtifactURL(repo, ref, suffix string) string {
 
 // Updates the list of dependencies in repo to the latest stable references
 func updateDeps(repo string, deps *[]u.Dependency, depChangeList *[]u.Dependency) error {
-	if repo != "istio" {
-		for _, dep := range *deps {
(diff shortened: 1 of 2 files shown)


toolbox/deps_update/main.go
```diff
@@ -82,12 +82,12 @@ func generateArtifactURL(repo, ref, suffix string) string {
...
-		for _, dep := range *deps {
-			if err := u.UpdateKeyValueInFile(dep.File, dep.Name, dep.LastStableSHA); err != nil {
-				return err
-			}
+	for _, dep := range *deps {
+		if err := u.UpdateKeyValueInFile(dep.File, dep.Name, dep.LastStableSHA); err != nil {
+			return err
 		}
+	}
+	if repo != "istio" || len(*hub) == 0 {
```
chxchx, line 90: For istio repo, it is important to call `./install/updateVersion.sh` which updates the yaml templates and debian urls that include the proxy tag. Perhaps we could removed the switch statements on mixer, pilot and auth given the mono repo change? @sebastienvas

chxchx: Quoting Krishna, "I kept the statements just in case we need to do 0.2 release (unlikely). We can cleanup later. Update version also does not need to be called with the new release process."

PR #637, opened 2017-12-19
Shorter gaps between dep propagation

chxchx:
**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1920,7 +1920,7 @@ periodics:
       secret:
         secretName: service-account
 
-- interval: 30m
+- interval: 10m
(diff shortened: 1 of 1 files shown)


yutongz:
Pre Submit(including e2e-smoke and pilot-e2e) usually takes 15+12=27 min. 30 min seems to be too ambitious.

chxchx:
As we have spoken offline, the dependency update script avoids creating duplicate PRs by checking an MD5 digest of all dependencies. So the consensus was to do this even more often.

yutongz:
/lgtm
That's pretty good.

PR #641, opened 2017-12-19
Update config.yaml

yutongz:
**Release note**:

```release-note
none
```

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -1920,7 +1920,7 @@ periodics:
       secret:
         secretName: service-account
 
-- interval: 10m
+- interval: 12h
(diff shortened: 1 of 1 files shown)


yutongz:
@rkpagadala approve this.

cxuu:
I thought we wanted to do this with shorter wait period? What is the reason behind such revert?

yutongz:
IIRC, the code didn't successfully skip the same diff, instead it created multiple PRs with the same changes. 

@rkpagadala : Correct me if I am wrong or if you want to add more info :)

rkpagadala:
Yes. For the same diff we got multiple PR's

PR #645, opened 2017-12-23
Auto-dep-update: cleanup before check

yutongz:
**Release note**:

```release-note
none
```

https://github.com/istio/proxy/pull/769

When doing dep-update, when the formal PR is not old enough to be close, the tool doesn't simply return, instead it force commits to that branch again and make Prow removes the "lgtm" label and go to a dead loop.

Thanks @rkpagadala pointing it out!

/cc @chxchx


yutongz:
hmm, here it works.

PR #647, opened 2017-12-27
remove too many owners file

rkpagadala:
Do we need so many OWNERS file, or is the top level enough?


yutongz:
Please keep the prow/OWNERS. I want to keep Nick's power in Prow folder :)

rkpagadala:
ok

PR #648, opened 2018-01-02
enable bot for proxy and istio

rkpagadala:
istio and proxy repo are fixed now. we can enable the bots


mattdelco:
This seems alright to me though I didn't know it was disabled in the first place.  It seems like the OWNERS file should be deleted.  There's only 1 name in the file--I'm guessing that file was added under the assumption it'd be additive to the parent's OWNERS file but afaik once an OWNERS file is found no further ancestry is consulted in the directory tree.

PR #650, opened 2018-01-03
add protoc 3.5.0 to prowbazel

### LINKED ISSUE #649: Need protoc version 3.5.0 on prow machine
Now that we are not using bazel for proto compiling, we need protoc on the prow build machines. We require v3.5.0 (this will match version in WORKSPACEs).

cc: @costinm

rkpagadala:
fixes #649

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -42,7 +42,7 @@ presubmits:
     - master
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.3.2
+      - image: gcr.io/istio-testing/prowbazel:0.3.4
         args:
         - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
         - "--clean"
@@ -141,7 +141,7 @@ presubmits:
     - release-0.2
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.3.2
+      - image: gcr.io/istio-testing/prowbazel:0.3.4
         args:
         - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
         - "--clean"
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -91,7 +91,7 @@ presubmits:
     - master
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.3.3
+      - image: gcr.io/istio-testing/prowbazel:0.3.4
         args:
         - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
         - "--clean"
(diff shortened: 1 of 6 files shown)


scripts/tools/linux-install-protoc
```diff
@@ -0,0 +1,35 @@
...
+
+PROTOC_VERSION="3.5.0"
+PROTOC_ARCHIVE="protoc-${PROTOC_VERSION}-linux-x86_64.zip"
+PROTOC_BASE_URL="https://github.com/google/protobuf/releases/download/v${PROTOC_VERSION}"
+PROTOC_URL="${PROTOC_BASE_URL}/${PROTOC_ARCHIVE}"
+PROTOC_BINARY='protoc'
+PROTOC_DIRECTORY="${TOOLS_DIR}/protoc"
+
+function update_protoc() {
+  local protoc_current="$(protoc --version | grep -oh '[0-9]*\.[0-9]*\.[0-9]*')" \
```
mattdelco, line 19: The protoc_current doesn't seem to be used anywhere.  I'm guessing this was derived from the repo script, which doesn't have a version.  Other scripts would wrap much of the functionality that follows within a:

if [[ "${PROTOC_VERSION}" != "${protoc_current}" ]]; then

prow/config.yaml
```diff
@@ -91,7 +91,7 @@ presubmits:
     - master
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.3.3
+      - image: gcr.io/istio-testing/prowbazel:0.3.4
```
mattdelco, line 94: gitub doesn't seem to let me comment on an unchanged line.  There's two instance of 0.3.2 in this file that I'm guessing aren't at that version for a particular reason so they could also be updated to 0.3.4 (to keep things consistent).

mattdelco (commented):
I know this is being made in response to the issue request, but I have my doubts about the suitability of the request.  I don't see why the build doesn't just pull this in for itself.  I'm guessing protoc is being treated differently because it's not a go library pulled in by dep, but dep itself is also being pulled in by the top-level Makefile so it seems like protoc could be handled similarly.  This would make the build more self sufficient and hermetic while also avoiding the need for these images (or other people) to update protoc every time there's a desire to update.

scripts/tools/linux-install-protoc
rkpagadala: even golang install script is the same

prow/config.yaml
rkpagadala: fixed

scripts/tools/linux-install-protoc
mattdelco: I don't quite understand the comment.  The go install script has a:

if [[ "${GO_VERSION}" != "${go_current}" ]]; then

and if I run "grep _current *" in test-infra's scripts/tool/ then the only file that's currently not using the result of XXX_current is the one for repo.  I don't know how important this optimization is for the update function to no-op when the version is already the desired one, but in any case it seems like this protoc_current should either have something that reads/uses it or not exist at all.

rkpagadala: fixed

mattdelco (commented):
/lgtm
/approve

PR #651, opened 2018-01-04
update bot for non bazel

rkpagadala:
adjust to the new toml files

### DIFF
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -113,7 +113,7 @@ func updateDeps(repo string, deps *[]u.Dependency, depChangeList *[]u.Dependency
 		}
 	}
 	cmd := fmt.Sprintf("./install/updateVersion.sh %s", args)
-	_, err := u.Shell(nil, cmd)
+	_, err := u.Shell(cmd)
 	return err
 }
 
@@ -163,7 +163,7 @@ func updateDependenciesOf(repo string) error {
 		return nil
 	}
 
-	if _, err = u.Shell(nil, "git checkout -b "+branch); err != nil {
+	if _, err = u.Shell("git checkout -b " + branch); err != nil {
 		return err
 	}
 	if err = updateDeps(repo, &deps, &depChangeList); err != nil {
@@ -174,12 +174,12 @@ func updateDependenciesOf(repo string) error {
 	}
 	if repo == istioRepo {
 		goPath := path.Join(repoDir, "../../..")
-		env := []string{"GOPATH=" + goPath}
-		if _, err = u.Shell(env, "make depend"); err != nil {
+		env := "GOPATH=" + goPath
+		if _, err = u.Shell(env + " make depend"); err != nil {
 			return err
 		}
 	}
-	if _, err = u.Shell(nil, "git diff --quiet HEAD"); err == nil {
+	if _, err = u.Shell("git diff --quiet HEAD"); err == nil {
 		// it exited without error, nothing to commit
 		log.Printf("%s is up to date. No commits are made.", repo)
 		return nil
--- a/toolbox/deps_update/main.go
+++ b/toolbox/deps_update/main.go
@@ -155,7 +155,7 @@ func updateDependenciesOf(repo string) error {
 		return nil
 	}
 
-	if _, err = u.Shell(nil, "git checkout -b " + branch); err != nil {
+	if _, err = u.Shell(nil, "git checkout -b "+branch); err != nil {
 		return err
 	}
 	if err = updateDeps(repo, &deps, &depChangeList); err != nil {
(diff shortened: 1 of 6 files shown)


toolbox/util/commonUtils.go
```diff
@@ -86,9 +87,29 @@ func ContainsString(slice []string, target string) bool {
...
+	//  name = "istio.io/api"
+	//  revision = "b08011c721e03edd61c721e4943607c97b7a9879"
+
+	found := false
+	keySearch := fmt.Sprintf("name = %q", key)
+	for i, line := range lines {
+		if strings.HasPrefix(line, "#") || strings.HasPrefix(line, "//") {
+			continue
+		}
+		if strings.Contains(line, keySearch) {
```
mattdelco, line 101: Is it worthwhile to also verify that lines[i+1] has a prefix of "  revision ="?  It looks like other attributes could be specified like "version" or "source".

toolbox/util/commonUtils.go
```diff
@@ -116,8 +132,24 @@ func UpdateKeyValueInFile(file, key, value string) error {
...
+// UpdateKeyValueInFile updates in the file all occurrences of key to
+// a new value, except comments started with # or //
+func UpdateKeyValueInFile(file, key, value string) error {
+	input, err := ReadFile(file)
+	if err != nil {
+		return err
+	}
+	var found, foundToml bool
+	lines := strings.Split(input, "\n")
+	lines, found = updateKeyValueInLines(lines, key, value)
```
mattdelco, line 150: should this line be an "else" to the "if file == "Gopkg.toml" ?

toolbox/util/githubUtils.go
```diff
@@ -124,22 +125,30 @@ func GetAPITokenFromFile(tokenFile string) (string, error) {
...
-	if _, err := ShellSilent(
-		"git clone " + gclient.Remote(repo)); err != nil {
+	repoPath := path.Join(pathPrefix, repo)
+	if err := os.RemoveAll(repoPath); err != nil {
+		return "", err
+	}
+	if err := os.Chdir(pathPrefix); err != nil {
+		return "", err
+	}
+	var additionalEnv []string //no additional environment variables
```
mattdelco: Is there a significance with using using "additionalEnv" here but "nil" elsewhere?  I'd be inclined to just use nil, or have kept Shell() as-in but added a ShellWithEnv() that's used by the one instance I saw that cares to pass in an environment variable.

toolbox/util/commonUtils.go
rkpagadala: no, i wanted both to run on toml

toolbox/util/commonUtils.go
rkpagadala: fixed, by crashing if revision not found

mattdelco:
I didn't notice any remaining issues and looks alright to me, though I know Charles is taking a look and he's in the owners file so I'll let him do the formal approval tags.

toolbox/util/commonUtils.go
```diff
@@ -129,25 +165,28 @@ func GetMD5Hash(text string) string {
 	return hex.EncodeToString(hash[:])
 }
 
-// Shell runs command on shell and get back output and error if get one
-func Shell(format string, args ...interface{}) (string, error) {
-	return sh(format, false, args...)
+// Shell runs command on shell and get back output and error if get one,
+// it takes a set of environment vaiables that are appended to existing environment
+func Shell(env []string, format string, args ...interface{}) (string, error) {
```
chxchx: Maybe declaring a new function such as `ShellEnv()` that takes the extra `env []string` is a better approach since  changing function signature often create bugs where old function is still used somewhere but was forgotten to be updated. I have only seen one use case in this PR where `env` is not nil. Having to say nil every time we call Shell is also a bit overhead.

chxchx: With further thought, I would even say neither new function like `ShellEnv` or signature change is necessary. You can just define env vars before issuing command, say `Shell("GOPATH=/some/path; command")`

toolbox/deps_update/main.go
```diff
@@ -119,12 +122,19 @@ func updateDeps(repo string, deps *[]u.Dependency, depChangeList *[]u.Dependency
...
+	repoDir, err := u.CloneRepoCheckoutBranch(githubClnt, repo, *baseBranch, "", "src/istio.io")
 	if err != nil {
 		return err
 	}
 	defer func() {
-		if err = u.RemoveLocalRepo(repoDir); err != nil {
+		if err = os.Chdir(saveDir); err != nil {
+			log.Fatalf("Error during chdir: %v\n", err)
+		}
+		if err = u.RemoveLocalRepo("src"); err != nil {
```
chxchx, line 137: I think `RemoveLocalRepo` takes an absolute path

toolbox/util/commonUtils.go
chxchx: Okay that depends on a slightly different impl of `sh()`. Just copy over https://github.com/istio/istio/blob/master/tests/util/commonUtils.go#L134

toolbox/deps_update/main.go
rkpagadala: It just takes a path not absolute.

rkpagadala:
@chxchx PTAL

PR #656, opened 2018-01-06
change path for checkout

rkpagadala:


toolbox/deps_update/main.go
```diff
@@ -126,15 +126,15 @@ func updateDependenciesOf(repo string) error {
 	if err != nil {
 		return err
 	}
-	repoDir, err := u.CloneRepoCheckoutBranch(githubClnt, repo, *baseBranch, "", "src/istio.io")
+	repoDir, err := u.CloneRepoCheckoutBranch(githubClnt, repo, *baseBranch, "", "code/src/istio.io")
```
mattdelco: Based on what I have in my tree and what's used in green-builds manifest, I think this should you should be adding "go" rather than "code".  You'll also probably find that the other stuff you've added for creating the bin directory won't be necessary (I haven't needed to do this, and if it is necessary is seemed like a mkdir -p should be in the Makefile's rule for ${GOPATH}/bin/dep

rkpagadala: weird. it was not working for me before. But not it does. Changed the path to go as suggested.

toolbox/deps_update/main.go
```diff
@@ -175,7 +175,7 @@ func updateDependenciesOf(repo string) error {
 	if repo == istioRepo {
 		goPath := path.Join(repoDir, "../../..")
 		env := "GOPATH=" + goPath
-		if _, err = u.Shell(env + " make depend"); err != nil {
+		if _, err = u.Shell(env + " make depend.update"); err != nil {
```
mattdelco, line 178: I can see how "depend.update" is apparently needed for force updates, but I'm not sure if "depend" is still needed for first-time init (I'm using the word "init" generically rather than making an reference to how "depend" is just an alias for "init").  Offhand also looks like init.sh is the only thing that pulls in a newer envoy tar.  So, you might have to do "make depend depend.update" until this situation with the makefiles is improved.

rkpagadala: depend is not needed, i just tested it the code we need to commit is the same, in both cases

PR #664, opened 2018-01-20
allow envoy dependency for proxy

rkpagadala:


toolbox/deps_update/main.go
```diff
@@ -64,7 +71,19 @@ func updateDepSHAGetFingerPrint(repo string, deps *[]u.Dependency) (string, []u.
 	}
 	digest += *baseBranch + *hub
 	for i, dep := range *deps {
-		commitSHA, err := githubClnt.GetHeadCommitSHA(dep.RepoName, dep.ProdBranch)
+		var commitSHA string
+		if dep.RepoName == envoyRepoPath {
```
chxchx, line 75: I would move this specific logic to the shell script that is invoking this binary. Then we could schedule the prow job and pass in different flags accordingly. If we were to updated the schedule we do not need to change this binary anymore.

chxchx (commented):
Just one thing and otherwise lgtm

PR #672, opened 2018-01-25
Increase cleanup script frequency

sebastienvas:


prow/config.yaml
```diff
@@ -1123,7 +1123,7 @@ periodics:
       secret:
         secretName: auth-e2e-rbac-kubeconfig
```
yutongz: Actually, I should mention it in your latest PR. If we are not supporting 0.2, we can delete these clusters for old auth/pilot/mixer repo and of course, we don't need to cleanup them.

prow/config.yaml
```diff
@@ -1155,7 +1155,7 @@ periodics:
       secret:
         secretName: pilot-e2e-rbac-kubeconfig
```
yutongz: Same here

prow/config.yaml
```diff
@@ -1187,7 +1187,7 @@ periodics:
       secret:
         secretName: mixer-e2e-rbac-kubeconfig
```
yutongz: Same here.

prow/test-infra-cleanup-GKE.sh
```diff
@@ -27,4 +27,4 @@ set -u
 set -x
 
 echo "=== Cleaning up GKE clusters ==="
-./scripts/cleanup-GKE "$@"
+./scripts/cleanup-GKE -h 1
```
yutongz, line 30: I thought we agree 2 hours but I guess 1 is fine :)

sebastienvas: the problem is that the e2e-smoke also creates cluster during presubmit, and they run more ofter so I rather have 1 h

prow/config.yaml
sebastienvas: done

prow/config.yaml
sebastienvas: done

prow/config.yaml
sebastienvas: done

yutongz (approved):
/lgtm

PR #678, opened 2018-02-02
Separate prow unit test job

chxchx:


prow/config.yaml
```diff
@@ -198,6 +198,38 @@ presubmits:
...
+          securityContext:
+            privileged: true
+          env:
+          - name: GOOGLE_APPLICATION_CREDENTIALS
+            value: /etc/service-account/service-account.json
+          volumeMounts:
+          - name: service
+            mountPath: /etc/service-account
+            readOnly: true
+          - name: kubeconfig
```
sebastienvas: Do we still need a kubeconfig for unit tests ?

rkpagadala:
@chxchx there is no target go-test in istio Makefile, please check prow/istio-unit-test.sh

prow/config.yaml
yutongz: Yes, afaik pilot, security unit tests require them.

And I would suggest you create a new cluster dedicated for unit tests, a small one might just work.

prow/config.yaml
```diff
@@ -198,6 +198,38 @@ presubmits:
         secret:
           secretName: istio-e2e-rbac-kubeconfig
     run_after_success:
+    - name: istio-unit-test
```
yutongz: IIRC, istio-unit-test should not be a "run_after_success" of presubmit. Instead, those two are supposed to run in parallel, right?

prow/config.yaml
chxchx: You are right. I kept it since `setup_and_export_git_sha` in `lib.sh` uses it but then it is just exporting the directory and no unit tests access it so I have removed it in latest commit

chxchx:
@rkpagadala Thanks for the good catch! I was not aware of the makefile change. I have updated prow/istio-unit-test.sh in this PR https://github.com/istio/istio/pull/3127. Please review.

prow/config.yaml
chxchx: Done

chxchx:
PTAL

prow/config.yaml
```diff
@@ -197,34 +197,39 @@ presubmits:
...
+          mountPath: /home/bootstrap/.kube
+      volumes:
+      - name: service
+        secret:
+          secretName: service-account
+      - name: kubeconfig
+        secret:
+          secretName: istio-e2e-rbac-kubeconfig
+    run_after_success:
     - name: istio-pilot-e2e
```
yutongz: These e2e-kind tests (e2e-smoke, pilot-e2e) should run after istio-persubmit, in your change, these e2e tests will be triggered after istio-unit-test which is the same as we have now. 

So in my idea, it should be like this:
```
-istio-presubmit (build)
  -run_after_success
    -pilot-e2e
    -e2e-smoke
-istio-unit-test (build and unit test)
```

yutongz (changes requested):

prow/config.yaml
chxchx: Done

chxchx:
PTAL

PR #680, opened 2018-02-06
Update config.yaml

rkpagadala:
branch is 0.5  instead of 0.5.1

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -172,6 +172,11 @@ presubmits:
     branches:
     - master
     - release-0.5
+    - release-0.6
+    - release-0.7
+    - release-0.8
+    - release-0.9
+    - release-1.0
     spec:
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -171,7 +171,12 @@ presubmits:
     trigger: "(?m)^/(retest|test (all|istio-presubmit))?,?(\\s+|$)"
     branches:
     - master
-    - release-0.5.1
+    - release-0.5
+    - release-0.6
```
chxchx, line 175: I was not sure whether this would resort to no op if branch does not exist. Maybe add those branches in after we make them?

rkpagadala: it should no-op, I looked at the code. Trying to add more so that we will need less manual intervention when creating patch release.

PR #684, opened 2018-02-08
Install fpm in prowbazel and istio-builder images

chxchx:
Trying to fix https://github.com/istio/istio/pull/3279 which eventually will fix https://github.com/istio/istio/pull/2444

### DIFF
--- a/scripts/linux-install-software
+++ b/scripts/linux-install-software
@@ -39,6 +39,9 @@ function install_packages() {
     python \
     python-yaml \
     realpath \
+    ruby \
+    rubygems \
+    ruby-dev \
     time \
     unzip \
     uuid-dev \
@@ -46,6 +49,7 @@ function install_packages() {
     wget \
     zip \
     zlib1g-dev \
+    && gem install --no-ri --no-rdoc fpm \
     && return 0
   return 1
 }


mattdelco:
The change appears to be adequate to get fpm added, but it's probably also appropriate to add:

https://github.com/istio/test-infra/blob/master/docker/istio_builders/Makefile

and maybe this one (but perhaps it's basically unused now and if anything should be deleted):

https://github.com/istio/test-infra/blob/master/docker/prowbazel/Makefile

I'm not entirely clear if anyone still uses these Makefiles anymore to generate the images or if instead people basically just run "docker build" and push it manually.

rkpagadala:
/hold
matt / charles you can remove the hold once your concerns are addressed.
I did not want the bot to merge too fast.

chxchx:
@mattdelco @rkpagadala 
Those two make files are in use (manual invocation of `make image; make push` is needed to create new image with fpm) and in fact their docker files call the linux-install-software script to install the prereqs.
See https://github.com/istio/test-infra/blob/master/docker/istio_builders/istio_builder/Dockerfile#L17
and https://github.com/istio/test-infra/blob/master/docker/prowbazel/Dockerfile#L17
If no other concern, please comment `hold cancel`

PR #685, opened 2018-02-08
Upgrade prowbazel image to 0.3.6 to include fpm

chxchx:
I have run `make image && make push` already from prowbazel directory


rkpagadala:
This looks good to me, can you also look at merging istio-builder and prowbazel and just use one docker image everywhere. I think istio-builder is supposed to supercede

chxchx:
I was not sure why there are two separate images to begin with. I have just upgraded istio-builder image as well and pushed.

PR #688, opened 2018-02-10
Job-level flakyness detection by rerun

chxchx:
This PR starts triggering of reruns on failed jobs and reports flakiness data. More PRs to come. See the [design doc](https://docs.google.com/document/d/1VgLU_ELy6Ohd8XA5ap_DDH0Vsecf2iclNREjBtz6tRg/edit#heading=h.5k59xksatw2r) for more details.


chxchx:
PTAL @yutongz @sebastienvas @rkpagadala

toolbox/oncall_alert/main.go
```diff
@@ -24,120 +24,84 @@ import (
...
-	for _, t := range protectedPostsubmits {
-		postSubmitJobs = append(postSubmitJobs,
-			&postSubmitJob{
-				name: t,
-				// init to be true to avoid false negative
-				latestRunPass: true,
-			})
+	// Ensure all reruns commands are issued to Prow
+	if _, err := u.Shell(`gcloud container clusters get-credentials prow \
+		--project=istio-testing --zone=us-west1-a`); err != nil {
```
sebastienvas: Avoid hardcoding prow cluster info

chxchx: It is necessary to have this check tho. Do you think defining them in flags as default value is a good idea? Or is there some configuration file I could access to get this info at runtime?

toolbox/oncall_alert/config.go
```diff
@@ -0,0 +1,59 @@
...
+	gmailAppPassFileDocker = "/etc/gmail/gmail-app-pass"
+)
+
+var (
+	gcsBucket        = flag.String("bucket", "istio-prow", "Prow artifact GCS bucket name.")
+	interval         = flag.Int("interval", 60, "Check and report interval(seconds)")
+	numRerun         = flag.Int("num_rerun", 1, "Number of reruns to detect flakyness")
+	owner            = flag.String("owner", "istio", "Github owner or org")
+	tokenFile        = flag.String("github_token", tokenFileDocker, "Path to github token")
+	gmailAppPassFile = flag.String("gmail__app_password", gmailAppPassFileDocker, "Path to gmail application password")
```
sebastienvas: gmail__ to _

sebastienvas: Did you create a token or are you sharing the actual password ?

toolbox/oncall_alert/config.go
```diff
@@ -0,0 +1,59 @@
...
+	protectedRepo    = flag.String("protected_repo", "istio", "Protected repo")
+	protectedBranch  = flag.String("protected_branch", "master", "Protected branch")
+	guardMaster      = flag.Bool("guard_master", false, "Suspend merge bot if postsubmit fails")
```
sebastienvas: It should not be limited to master.

toolbox/util/gcsClient.go
```diff
@@ -44,12 +45,10 @@ func NewGCSClient() *GCSClient {
 func (gcs *GCSClient) GetFileFromGCSReader(bucket, obj string) (*storage.Reader, error) {
```
sebastienvas: rename to GetFile since its GCSClient it is implied.

toolbox/util/gcsClient.go
```diff
@@ -61,15 +60,23 @@ func (gcs *GCSClient) GetFileFromGCSString(bucket, obj string) (string, error) {
...
 	if _, err = buf.ReadFrom(r); err != nil {
-		log.Printf("Failed to read from gcs reader, %v", err)
+		log.Printf("Failed to read from gcs reader, %v\n", err)
 		return "", err
 	}
-
 	return buf.String(), nil
 }
+
+// WriteTextToFileOnGCS writes text to file on gcs
```
sebastienvas: rename to WriteTestToFile

toolbox/util/commonUtils.go
```diff
@@ -40,6 +40,27 @@ var (
 	}
 )
 
+// IsLongerThan compares two duration
+func IsLongerThan(durA time.Duration, durB time.Duration) bool {
```
sebastienvas: Do you really need a function for this ?

toolbox/util/githubUtils.go
```diff
@@ -180,3 +188,23 @@ func CreateCommitPushToRemote(branch, commitMsg string) error {
 	_, err := Shell("git push -f --set-upstream origin " + branch)
 	return err
 }
+
+// BlockMergingOnBranch adds "do-not-merge/post-submit" labels to all PRs
+// in :branch on the repo which the :githubClnt connects to
+func BlockMergingOnBranch(githubClnt *GithubClient, repo, branch string) error {
```
sebastienvas, line 194: Maybe change this to write /hold on the PR ? that way people know what s going on and it is not implementation specific

toolbox/util/gcsClient.go
```diff
@@ -61,15 +60,23 @@ func (gcs *GCSClient) GetFileFromGCSString(bucket, obj string) (string, error) {
...
+func (gcs *GCSClient) WriteTextToFileOnGCS(bucket, gcsObj, txt string) error {
```
sebastienvas: Where is this used ?

toolbox/oncall_alert/main.go
sebastienvas: just define it as a constant at least.

toolbox/oncall_alert/main.go
```diff
@@ -265,31 +286,64 @@ func getProwResult(target string) (*ProwResult, error) {
...
 	}
-
-	if err := githubClnt.RemoveLabelFromPRs(options, *protectedRepo, doNotMergeLabel); err != nil {
-		log.Printf("Failed to remove label to PRs: %v", err)
-		return
+	flakeStats, err := DeserializeFlakeStats(flakeStatsString)
+	if err != nil {
+		return err
+	}
+	flakeStats = append(flakeStats, newFlakeStat)
```
sebastienvas: You cannot use GCS storage for this, as you cannot guarantee consistenty. You want to use storage that supports transaction. Maybe bigquery ? Have you looked at the flaky detection that k8s is using ? I think they already do all this anyways.

sebastienvas (changes requested):
PLease see comments

toolbox/oncall_alert/main.go
chxchx: Done

toolbox/oncall_alert/config.go
chxchx: Done

toolbox/util/gcsClient.go
chxchx: I believe you mean `WriteTextToFile`. and done.

toolbox/util/gcsClient.go
chxchx: I rename it to Read() for simplicity. In my defense Yutong commited this code lol

toolbox/util/commonUtils.go
chxchx: It is just that calling Nanoseconds() repetitively is not very clean. I removed this function tho.

toolbox/oncall_alert/config.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2017 Istio Authors
```
rkpagadala: 2018

toolbox/util/githubUtils.go
rkpagadala: it hold/post-submit-failure supported? that way it won't conflict with other uses of /hold

chxchx: Unfortunately hold/post-submit-failure is not supported. Using /hold cancel directly might unblock PRs that are intended to be held off. I will sync up with Yutong to understand if we could define our own label on mungegithub and how that is going to affect the migration to tide. For now lets keep it as it.

toolbox/oncall_alert/main.go
chxchx: Understand. After talking with @hklai I think it is a good idea to work on the storage in a separate PR, so that we could have the rerun/retest running first, and prow shows these rerun too. In the meantime, I will update the design doc to have a design on the durable transactional storage, and we can carry on the discussion there.

toolbox/oncall_alert/config.go
chxchx: Done

chxchx:
PTAL

toolbox/oncall_alert/config.go
```diff
@@ -0,0 +1,60 @@
...
+	sender          = "istio.testing@gmail.com"
+	oncallMaillist  = "istio-oncall@googlegroups.com"
+	messageSubject  = "ATTENTION - Istio Post-Submit Test Failed"
+	messagePrologue = "Hi istio-oncall,\n\n" +
+		"Post-Submit is failing in istio/istio, please take a look at following failure(s) and fix ASAP\n\n"
+	messageEnding = "\nIf you have any questions about this message or notice inaccuracy, please contact istio-engprod@google.com."
+	losAngeles    = "America/Los_Angeles"
+
+	// Gmail setting
+	gmailSMTPSERVER = "smtp.gmail.com"
```
sebastienvas: gmailSMTPServer

toolbox/oncall_alert/config.go
```diff
@@ -0,0 +1,60 @@
...
+	gmailSMTPPORT   = 587
```
sebastienvas: gmailSMTPPort

toolbox/oncall_alert/main.go
```diff
@@ -172,91 +175,108 @@ func formatMessage(failures map[*postSubmitJob]bool) (mess string) {
...
-	// Total count of jobs which is in failing status right now.
-	// Including jobs failed in circles before.
-	// If 0 --> unblock auto-merge
-	failedCount := 0
-
-	for _, job := range postSubmitJobs {
-		latestRunNo, err := getLatestRun(job.name)
+// Returns an array of postSubmitJob since the same job might fail at multiple runs
+func checkOnJobsWatched() []failure {
+	newFailures := []failure{}
```
sebastienvas: prefer var newFailures []failure. That way user can check if the value return by the function is nil as opposed to being empty.

toolbox/oncall_alert/main.go
```diff
@@ -146,24 +110,63 @@ func main() {
...
+			return e
+		}); err != nil {
+			log.Printf("Unable to trigger the %d-th rerun of job %v", i, job.name)
+		}
+	}
+	return nil
+}
+
+func getProwJobConfig(job *jobStatus, runNo int) (ProwJobConfig, error) {
+	cfg := ProwJobConfig{}
```
sebastienvas: var cfg ProwJobConfig since it is going to be used by jason marshall, that way you can check if it remains nil

toolbox/oncall_alert/main.go
```diff
@@ -172,91 +175,108 @@ func formatMessage(failures map[*postSubmitJob]bool) (mess string) {
...
-				continue
-			}
-			job.latestRun = latestRunNo
-			job.latestRunPass = prowResult.Passed
-			if *debug {
-				log.Printf("Job: %s -- Latest Run No: %d -- Passed? %t", job.name, latestRunNo, prowResult.Passed)
+		// Clone a slice of keys in map to avoid editing while iterating
+		// Then check previously pending runs to see if results are ready
+		pendingRuns := make([]int, 0, len(job.pendingFirstRuns))
+		for runNo := range job.pendingFirstRuns {
```
sebastienvas: you meant to use an array ?
```golang
var pendingRuns [10]int
for i:=0; i<10; i++ {
  pendingRuns[i] = i
}
```

sebastienvas (changes requested):
I would like to see some tests. You should be able to create an interface for ProwJob info and change your code  accept that interface.

The code is big enough that you should move all the logic into a pkg and create a main that sets the flags to use.

toolbox/oncall_alert/config.go
chxchx: Done

toolbox/oncall_alert/config.go
chxchx: Done

toolbox/oncall_alert/main.go
chxchx: Done

toolbox/oncall_alert/main.go
chxchx: Done

toolbox/oncall_alert/main.go
chxchx: I tried your suggestion but no luck. The reason is that `pendingFirstRuns` has type `map[int]time.Time`. The size of the key set is unknown at compile time so I cannot do things like `var pendingRuns [10]int`. Using make to claim the size that fits exactly our need and then calling append() will not resize the array.

chxchx:
@sebastienvas @rkpagadala 
I have finished package encapsulation and addressed other comments as well. Please take a look. In the meantime, I will continue to work on adding tests for this.

chxchx:
PTAL

sisyphus/sisyphusd_test.go
```diff
@@ -0,0 +1,216 @@
...
+		prowZoneMock, gubernatorURLMock, gcsBucketMock, &Config{
+			CatchFlakesByRun: true,
+			PollGapDuration:  100 * time.Millisecond,
+		})
+	prowAccessorMock := NewProwAccessorMock(gubernatorURLMock)
+	prowAccessorMock.exitSignalToSisyphus = make(chan bool, 1)
+	sisyphusd.prowAccessor = prowAccessorMock
+	sisyphusd.SetExitSignal(prowAccessorMock.exitSignalToSisyphus)
+	sisyphusd.storage = NewStorageMock(t)
+	sisyphusd.Start()
```
sebastienvas: you should probably run start in a goroutine just in case it times out.

sisyphus/prowAccessor.go
```diff
@@ -0,0 +1,161 @@
...
+	gcsClient     *u.GCSClient
+}
+
+// NewProwAccessor creates a new ProwAccessor
+func NewProwAccessor(prowProject, prowZone, gubernatorURL, gcsBucket string) *ProwAccessor {
+	return &ProwAccessor{
+		prowProject:   prowProject,
+		prowZone:      prowZone,
+		gubernatorURL: gubernatorURL,
+		gcsClient:     u.NewGCSClient(gcsBucket),
```
sebastienvas: Instead of using GCSClient, you should use an interface that just Read(). That way by mocking the storage, and you can actually test the ProwAccessor code.

sisyphus/sisyphusd_test.go
```diff
@@ -0,0 +1,216 @@
...
+		NumRerun:         DefaultNumRerun,
+	}
+	sisyphusd := NewDaemon(protectedJobsMock, prowProjectMock,
+		prowZoneMock, gubernatorURLMock, gcsBucketMock, cfg)
+	if !reflect.DeepEqual(sisyphusd.GetConfig(), cfgExpected) {
+		t.Error("setting catchFlakesByRun failed")
+	}
+}
+
+func TestProwResultsMock(t *testing.T) {
```
sebastienvas, line 178: funny that you have a test for you Mock

sisyphus/prowAccessor.go
```diff
@@ -0,0 +1,161 @@
...
+	return &cfg, nil
+}
+
+func (p *ProwAccessor) triggerConcurrentReruns(jobName, node string, numRerun int) error {
+	log.Printf("Rerunning %s\n", jobName)
+	recess := 1 * time.Minute
+	maxRetry := 3
+	for i := 0; i < numRerun; i++ {
+		if err := u.Retry(recess, maxRetry, func() error {
+			_, e := u.Shell("kubectl create -f \"https://prow.istio.io/rerun?prowjob=%s\"", node)
```
sebastienvas: Same comment here, you might want to abstract this call to ease testing of the rest of the code

sisyphus/prowAccessor.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	lastBuildTXT = "latest-build.txt"
+	finishedJSON = "finished.json"
+	startedJSON  = "started.json"
+)
+
+// IProwAccessor defines the Prow interface
+type IProwAccessor interface {
```
sebastienvas: you can even call this CIinfo or something as those are not very Prow Specific. 
```goland
GetLatestRun(jobName string) (int, error)
GetResult(jobName string, runNo int) (*ProwResult, error)
Rerun(jobName string, runNo, numRerun int) error
GetRunURL() string
```

sisyphus/sisyphusd.go
```diff
@@ -0,0 +1,285 @@
...
+			} else {
+				log.Printf("No new tests failed in last circle.")
+			}
+		}
+		if d.branchProtector != nil {
+			d.branchProtector.process(newFailures)
+		}
+		log.Printf("Sleeping for %d", d.pollGapDuration)
+		select {
+		case <-d.exitSignal:
```
sebastienvas: probably pass a context to start instead. It will also ease testing as you can pass a timeout context.
Also when doing a for loop like that it might be good to call a private startOne() method that does one iteration only, that way you check the expected output of one iteration in a test.

sebastienvas (commented):
PTAL

sisyphus/sisyphusd_test.go
chxchx: My experience has been that a test function will return and be marked as pass once execution reaches the end, even if there are unfinished go routines. To ensure that `sisyphusd` has looped through all the mock test runs, we have to block the test function from exiting until `sisyphusd` is done, but that is equivalent to what I have here, which is to call `sisyphusd.Start()` directly. We have a 300 seconds timeout limit on testinfra presubmit so I think we should be fine.

sisyphus/sisyphusd_test.go
chxchx: I know \*laugh_cry\* The mock has become rather complected, and it becomes necessary to ensure the mock behaviors conform to its specification before we use the mock to test even more complicated rerun logic.

sisyphus/prowAccessor.go
chxchx: Done

sisyphus/prowAccessor.go
chxchx: Done

sisyphus/prowAccessor.go
chxchx: Done

sisyphus/sisyphusd.go
chxchx: Done

chxchx:
PTAL

sisyphus/ci.go
```diff
@@ -0,0 +1,175 @@
...
+type ProwMetadata struct {
+	Repo       string `json:"repo"`
+	RepoCommit string `json:"repo-commit"`
+}
+
+// ProwJobConfig matches the structure published in started.json
+type ProwJobConfig struct {
+	Node        string `json:"node"`
+	JenkinsNode string `json:"jenkins-node"`
+	Version     string `json:"version"`
```
sebastienvas, line 68: Are you using the JenkinsNode ? if not you can remove it, json parsing will do the right thing.

sebastienvas:
thanks for all the changes. 
/lgtm

sebastienvas (approved):

sebastienvas:
Please check if you need to get rid of the JenkinsNodes, because I think k8s might remove that eventually so, that way you won't have to update your code if they do.

sebastienvas:
@chxchx please make sure to run scripts/update_deps.sh. It seems that the test you wrote were not actually run by the presubmit since they were not added in the BUILD. I tried fixing it, but then the test fail with 
2018/03/28 06:52:01 Failed to create a gcs client, dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.

sebastienvas:
Nevermind I fixed it in https://github.com/istio/test-infra/pull/756. However running gcsclient in a unitest is not recommended as you can't predict the user env(not everybody uses gcloud). In my desktop and laptop those tests are failing, but they are passing in Prow.

chxchx:
The gcs client was created as a default value in prowAccessor constructor. Although the unit tests substitute the gcs client with a mock, this constructor was called at first so we still have the credential issue Seb mentioned. This PR https://github.com/istio/test-infra/pull/758 should fix it.

PR #691, opened 2018-02-15
Update bazel to 0.10.0

sebastienvas:


rkpagadala:
@sebastienvas how about istio-builder?

sebastienvas:
Updated istio-builders as well.

PR #698, opened 2018-02-16
Install go-junit-report in prowbazel image

chxchx:


sebastienvas:
Don't you need to update the prow config files as well ?

chxchx:
@sebastienvas 

Yes Seb, I was creating this PR for review before using the image directly. To address your comment I have built and pushed the image and also include update to prow configmap in this PR as well. PTAL.

rkpagadala:
lgtm letting seb let this go through

sebastienvas (approved):

PR #705, opened 2018-02-21
Use prowbazel 0.4.2 in bookinfo test

chxchx:


chxchx:
Ping @sebastienvas @hklai

rkpagadala:
@chxchx why do need this change, what does the 0.4.2 version of prowbazel have? Can you add some info in the description

sebastienvas (approved):

chxchx:
@rkpagadala 0.4.2 contains go-junit-report. This upgrade is not necessary for bookinfo since this job does not use go-junit-report just yet. Nevertheless, this upgrade is beneficial to have all e2e jobs use the same base image, such that in future upgrades we will always remember to upgrade all and not just left a single one out.

I encountered this bug when I was testing the new image containing go-junit-report only to be told command not found. It took a while for me to test the docker image on a different host and to ssh into the Prow cluster to check the deployed configmap before Seb realized the bookinfo e2e job was still using old image. Since I did a search and replace 0.4.1 -> 0.4.2 it still was not updated. 

We also keep a change log of what each version of images has. I love this idea and it is also a good reference to have.

PR #707, opened 2018-02-26
Enabling mason for master and next releases

sebastienvas:
This PR duplicates istio presubmit and postsubmit entries such that master and future releases do not need to mount a kubeconfig and older release like 0.5 and 0.6 still use the kubeconfig.

It is also enables mason on dailies


rkpagadala:
lgtm

chxchx:
/lgtm

if only we had templates. Could make changes like this much easier.

sebastienvas:
ah shoot I needed  https://github.com/istio/istio/pull/3790 to be approved. Now most test will fail until the PR is merged

PR #710, opened 2018-02-27
Updating prow after 5 months.

sebastienvas:


sebastienvas:
I just force merged as I updated some config manually when doing the upgrade and did not want those changes to be undone by a another merge

Issue #712, opened 2018-02-28
Upgrade prow to go 1.10


mandarjog:
We need to upgrade prow to 1.10.

gofmt 1.10 makes subtle changes in white space formatting. This can throw off checks.
We should update to go 1.10 as soon as possible.

chxchx:
Fixed by https://github.com/istio/test-infra/pull/716

PR #711, opened 2018-02-28
Update bazel to 0.11

lizan:
Needed by https://github.com/istio/proxy/pull/1176

PR #714, opened 2018-02-28
Bump up golang to 1.10

sebastienvas:
This does nothing until I create the new image and update the images that Prow uses. I''ll do this second part once release 0.6 is out


sebastienvas:
Fixes part of #714 
Note that I will push both bazel 0.11 and golang 1.10 in the same image. cf #711

mandarjog:
partially addresses #712 
Thanks.

PR #723, opened 2018-03-05
Bind tester service account to clusteradmin.

myidpt:
Fixes https://github.com/istio/istio/issues/3919.

### DIFF
--- a/scripts/update-e2e-cluster.sh
+++ b/scripts/update-e2e-cluster.sh
@@ -34,6 +34,8 @@ PROW_ZONE='us-west1-a'
 PROW_PROJECT='istio-testing'
 PROW_TEST_NS='test-pods'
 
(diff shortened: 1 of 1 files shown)


ldemailly:
how is this different from
https://github.com/istio/istio/blob/d9a5caea8443b058ab7b689f2a802c9d9afa00d6/prow/cluster_lib.sh#L46-L50

we already had clusteradmin role otherwise nothing at all would have worked (not just CA stuff)

sebastienvas:
That s for the shared cluster. cluster lib is used for provisioned cluster

ldemailly:
eventually we shall split unit tests from tests that require an api server and more
but thanks for this current fix/workaround

sebastienvas:
Yes that's what I want to do to

PR #724, opened 2018-03-05
creating cluster admin role on right cluster

sebastienvas:


ldemailly:
:)

myidpt:
Oh, thanks for the fix.

PR #728, opened 2018-03-07
Adds flags go pkg_check to set build id and jobName

sebastienvas:
To better integrate with circle CI
And update logging to use glog


yutongz (approved):
/lgtm

PR #729, opened 2018-03-08
Install clang-5.0 for proxy build

lizan:
For https://github.com/istio/proxy/issues/1124


lizan:
/assign @sebastienvas 
can you take a look?

rkpagadala (commented):
lgtm but I think we might need to update istio_builders also

lizan:
@rkpagadala I don't think we are using istio_builders to build proxy (envoy) binaries, no? right now clang is only needed for ASAN and TSAN tests, so only prowbazel will be good.

sebastienvas (approved):

PR #739, opened 2018-03-19
Update prow

sebastienvas:


prow/cluster/deck_deployment.yaml
```diff
@@ -19,14 +19,14 @@ spec:
...
-        image: gcr.io/k8s-prow/deck:v20180223-28212fe21
+        image: gcr.io/k8s-prow/deck:v20180316-93ade3390
         imagePullPolicy: Always
         ports:
           - name: http
             containerPort: 8080
         args:
         - --hook-url=http://hook:8888/plugin-help
-        - --redirect-http-to=prow.istio.io
+        #- --redirect-http-to=prow.istio.io
```
chxchx, line 29: Why removing this?

sebastienvas: I think there was a bug in the code, the server was receiving to many redirect, so I just disabled it until they fix it.

PR #741, opened 2018-03-20
Update coverage metric to get data directly from GCS

sebastienvas:


sebastienvas:
FYI, working at http://35.199.148.121:9103/metrics

toolbox/metrics/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+image:
+	bazel run //toolbox/metrics/cmd:metrics
+
+deploy: 
+	kubectl apply -f cmd/deployment.yaml 
+	kubectl apply -f cmd/service.yaml
+
+config:
+
```
chxchx: Is the newline intentional?

toolbox/metrics/cmd/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+var (
+	port               = flag.Int("listen_port", 9103, "Port on which to expose metrics and web interface.")
+	gcsBucket          = flag.String("bucket", "istio-code-coverage", "GCS bucket name.")
+	codeCovTrackJob    = flag.String("coverage_job", "codecov_master", "In which job we are tracking code coverage.")
+	githubRepo         = flag.String("repo", "istio", "Repo for which we are capturing code coverage")
+	serviceAccountJSON = flag.String("service_account_json", "", "Path to the service account JSON")
+
+	ms *metrics.Publisher
+	// Cannot set storage in init since we need flags.
```
chxchx: If you move line 65-73 into init() you will be able to set storage when creating it.

In fact, it is a better pattern to combine `NewGCSStorage()` and `storage.Set(param...)` as `NewGCSStorage(param...)` so users of this library will never miss the second step of setting params. 

Is it possible to merge the GCSStorage in the coverage pkg with GCSClient in toolbox/util/gcsClient.go? This way this library is available to all applications, instead of each application implements its own.

toolbox/metrics/cmd/main.go
```diff
@@ -0,0 +1,91 @@
...
+func newMetricPublisher() *metrics.Publisher {
+	suite := metrics.Suite{
+		"codecov": coverage.NewMetric(storage),
+	}
+	return metrics.NewPublisher(suite, defaultUpdateInterval, defaultUpdateTimeout)
+}
+
+func init() {
+	ms = newMetricPublisher()
+	ms.RegisterMetrics()
```
chxchx: Call `RegisterMetrics` inside `metrics.NewPublisher()` as part of processing `suite`.

toolbox/metrics/metrics.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Metric is used to collect data
+type Metric interface {
+	Update(ctx context.Context) error
+	GetCollector() prometheus.Collector
+}
+
+// RegisterMetrics should be called in init
+func (p *Publisher) RegisterMetrics() {
+	for _, m := range p.suite {
```
chxchx, line 51: Call `RegisterMetrics` inside `metrics.NewPublisher()` as part of processing `suite`. AKAICT, all usage of `Publisher` in this PR requires `RegisterMetrics()`.

toolbox/metrics/cmd/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    istio: metrics-fetcher
+  name: metrics-fetcher
+  namespace: metrics
+spec:
+  ports:
+  - port: 9103
```
chxchx, line 10: This is not an ask but I see multiple hardcoded ports 9103 in the yamls and in go files and it would be nice to have a single copy of config.

toolbox/metrics/cmd/main.go
sebastienvas: Yeah that's the way I had it before, did not know you could call flag.Parse in init.

toolbox/metrics/Makefile
sebastienvas: no

toolbox/metrics/metrics.go
sebastienvas: RegisterMetrics needs to be called in init per prometheus requirement so I would rather keep separate to keep it consistent with promotheus

toolbox/metrics/cmd/service.yaml
sebastienvas: in go this is a flag. I am not I understand what you want.

sebastienvas:
PTAL

PR #747, opened 2018-03-21
Enabled tide and remove test infra in mungegithub

sebastienvas:
I removed stop as I accidentally called update-configs yesterday which brought down prow. We should never do that, and if we do I think it should be manual.


yutongz (approved):
/lgtm

hklai (approved):
/lgtm

rkpagadala (approved):

PR #751, opened 2018-03-22
Using preset in prow config.yaml to remove duplicate code

sebastienvas:
presets can be used for env, volumes and volumes mounts. Get rid of 500 lines of code, which is not few.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -46,6 +46,15 @@ presets:
   - name: service
     secret:
       secretName: service-account
+- labels:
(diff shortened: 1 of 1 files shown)


sebastienvas:
@chxchx this is what I was talking about. With presets you can remove duplicate code. I wish we could also use the container image but that's not supported right now

rkpagadala (approved):
/lgtm

chxchx:
Thank you so much! I too wish they will soon support this in container spec.

PR #761, opened 2018-03-29
Disabling github approval as it is confusing

sebastienvas:
note for that last we need both github approval and /approve. After this is merged then we won't need to.


rkpagadala (approved):
/lgtm

PR #766, opened 2018-04-02
Add a new cluster in Mason config

sebastienvas:
@john-a-joyce 

Needed by https://github.com/istio/istio/pull/4240

### DIFF
--- a/boskos/configs.yaml
+++ b/boskos/configs.yaml
@@ -13,6 +13,11 @@ configs:
           version: 1.9
           zone: us-central1-f
           enablekubernetesalpha: true
+        - machinetype: n1-standard-2
+          numnodes: 4
+          version: 1.9
+          zone: us-west1-a
+          enablekubernetesalpha: true
         vms:
         - machinetype: n1-standard-4
           sourceimage: projects/debian-cloud/global/images/debian-9-stretch-v20180206


boskos/configs.yaml
```diff
@@ -13,6 +13,11 @@ configs:
           version: 1.9
           zone: us-central1-f
           enablekubernetesalpha: true
+        - machinetype: n1-standard-2
+          numnodes: 4
+          version: 1.9
+          zone: us-west1-a
```
tiswanso, line 19: for multicluster, I think we need the 2 gke projects to be in the same zone.  The reason is that we need pod to pod reachability.

boskos/configs.yaml
```diff
@@ -13,6 +13,11 @@ configs:
           version: 1.9
           zone: us-central1-f
           enablekubernetesalpha: true
+        - machinetype: n1-standard-2
```
tiswanso, line 16: Is modifying the current boskos resource type the short term change?  I was thinking we'd add a new element to the list for a resource type for use by multicluster tests.

boskos/configs.yaml
sebastienvas: They will be in the same project just in a different region.

boskos/configs.yaml
sebastienvas: It is up to you. I was affraid we would run out of resources if we add another one. Maybe not.

tiswanso: I'm not sure of the cost of new projects v. the cost of making all projs in current boskos pool get 2 gke clusters allocated all the time.  I think the 2 gke cluster allocation would be more expensive.

One other thing I noticed is that if we did make the 2nd cluster different (smaller), I think the only way we could tell which cluster was which type is to use gcloud commands.  I think if we add a name attribute here in the boskos/config.yaml we could just use `kubectl config *-context` to determine the more powerful cluster.  But if other tests need multiple identical clusters the the name attribute isn't required.

sebastienvas:
Merging this for now, we can revisit after

PR #772, opened 2018-04-04
pilot-multicluster-e2e use presets, enable reporting

tiswanso:
original pilot-multicluster-e2e job diffs were created prior to
presets merged. Missed the presets prior to merging the multicluster
job config.  Additionally, removed the skip_report flag.

Also, introduce some common macros for yaml sections that are
common for multiple job instances. Only use in pilot-multicluster
for now.

Actual e2e test implementation: https://github.com/istio/istio/pull/4240

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -81,6 +81,29 @@ presets:
     secret:
       secretName: daily-release-e2e-rbac-kubeconfig
 
+common_e2e_branches: &common_e2e_branches
+  - master
+  - release-0.7
+  - release-0.8
+  - release-0.9
+  - release-1.0
+
+common_e2e_spec: &common_e2e_spec
+  containers:
+  - image: gcr.io/istio-testing/prowbazel:0.4.7
+    args:
+    - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
+    - "--clean"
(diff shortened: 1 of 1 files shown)


tiswanso:
@sebastienvas -- tried trigger after merge of #767 and #766, I see evidence of the resulting job [here](http://gcsweb.k8s.io/gcs/istio-prow/directory/istio-pilot-multicluster-e2e/) but I'm not sure I can see any results without the `skip_report: true` removed.  So, removing that.  

Also, I noticed I missed the presets changes so fixing that as well.  Also I am proposing some additional yaml niceness to this file, in case folks like it :-)

chxchx:
/approve
But would like to let Seb to review before lgtm

sebastienvas:
@tiswanso are you sure yaml templating works in kuberenetes ?

tiswanso:
@sebastienvas -- not 100% sure.  It's in yaml spec 1.1.  But I see the "<<:" merge key operator is an parser optional supported capability in the spec.  All the python yaml modules I've used support it.  Do you know of a yaml linter that I can run that would give more confidence that prow likes it?

sebastienvas:
Apparently this should work
/lgtm

tiswanso:
thanks, @sebastienvas & @chxchx !

PR #776, opened 2018-04-05
add boskos config for perf tests

rkpagadala:


rkpagadala:
@sebastienvas not sure if I did the right thing. PTAL and correct me

boskos/configs.yaml
```diff
@@ -28,3 +28,35 @@ configs:
...
+        - machinetype: n1-highcpu-2
+          sourceimage: projects/ubuntu-os-cloud/global/images/ubuntu-1604-xenial-v20180323
+          zone: us-central1-f
+          tags:
+          - http-server
+          - https-server
+          scopes:
+          - https://www.googleapis.com/auth/cloud-platform
+
+- name: gke-perf-empty
```
sebastienvas: you don't need that. You can just ask for a project in the resources.yaml. However we might want to create a new a project pool type for this.

PR #777, opened 2018-04-05
Move ci_to_gubernator main pkg into sub directory

chxchx:
This change is required to enable go get on this main pkg. Currently, both main and ci_to_gubernator pkgs reside in the same directory, which seems to confuses go get.


chxchx:
PTAL

PR #785, opened 2018-04-11
Make mason construct faster by creating cluster and vms in parallels

sebastienvas:
This should speed up recycling time and therefore solves some of the tests failing because not resources are available. 

This should help https://github.com/istio/istio/issues/4862

I added changed the config format, which means that I would have to redeploy mason and reset all the resources. The new config format is easier to use and validate.

I also added validation based on the mistakes that @rkpagadala made, such that we can catch those when the PR is made.


boskos/gcp/config.go
```diff
@@ -93,6 +99,11 @@ func (rc *resourcesConfig) Construct(res *common.Resource, types common.TypeToRe
 		return r
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), operationTimeout)
+	defer cancel()
+	errGroup, derivedCtx := errgroup.WithContext(ctx)
+	var mutex sync.RWMutex
```
rkpagadala: Can this name be more meaningful. Change the name to show what it is responsible for locking

rkpagadala (dismissed):
otherwise lgtm

PR #788, opened 2018-04-11
Sisyphus on release qualification as presubmit

chxchx:
Previous implementation of sisyphus only works with postsubmit due to GCS path structure. This PR aims to extend sisyphus to generate the GCS path differently depending if the jobs are presubmits or postsubmits.

The objective is to deploy sisyphus (test rerun) on release qualifications first. This change is required before the deployment.


sisyphus/ci.go
```diff
@@ -77,8 +77,10 @@ type ProwAccessor struct {
 	prowProject   string
 	prowZone      string
 	gubernatorURL string
+	gcsBucket     string
 	gcsClient     u.IGCSClient
 	rerunCmd      func(node string) error
+	presubmitJobs []string
```
rkpagadala: use a map for this

chxchx: Do you mean merge protectedJobs and presubmitJobs into one map such as map[string]bool where the bool indicates if a job is in presubmit, or do you mean replacing the highlighted line as `presubmitJobs map[string]bool` where we only care about the keySet, and we can do constant time constainsKey look up instead of iterating through all keys?

rkpagadala: I was thinking about lookup vs iterating through all keys. If you can merge presubmit and postsubmit that sounds better

chxchx: Done

chxchx:
PTAL

sisyphus/ci.go
```diff
@@ -80,7 +80,7 @@ type ProwAccessor struct {
 	gcsBucket     string
 	gcsClient     u.IGCSClient
 	rerunCmd      func(node string) error
-	presubmitJobs []string
+	presubmitJobs map[string]bool
```
rkpagadala: minor comment: since you are not using value, it can be empty
```presubmitJobs map[string]struct{}```

rkpagadala (commented):
otherwise lgtm

chxchx:
PTAL

PR #790, opened 2018-04-12
Adding v1alpha3-v2 as a required test

rshriram:


chxchx:
I re-deployed mungegithub for you on istio to take this config.

PR #794, opened 2018-04-13
Fix githubctl

hklai:
My previous change broke githubctl as it would exit the loop after just one cycle of wait.
The fix has been tested locally, and pushed to the airflow gcs bucket last night. We have a daily release today. =)

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -285,10 +285,11 @@ func DailyReleaseQualification() error {
 
 	verbose := true
 	ci := u.NewCIState()
-	retryDelay := 5 * time.Minute
+	retryDelay := 5 * time.Second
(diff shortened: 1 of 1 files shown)


toolbox/githubctl/main.go
```diff
@@ -289,18 +289,20 @@ func DailyReleaseQualification() error {
 	totalRetries := 240 // Max 20 hours wait before we automatically close the PR and fail release qualification.
```
rkpagadala: Change code to
 	retryDelay := 5 * time.Minute
        maxWaitTime := 20 * time.Hour // Max 20 hours wait before we automatically close the PR and fail release qualification.
 	totalRetries := int( maxWaitTime / retryDelay)

hklai: Done

hklai: Done. PTAL

rkpagadala (approved):
/lgtm

Issue #795, opened 2018-04-13
Daily tests should not test in one namespace


Tahler:
istio/test-infra's [config.yaml](https://github.com/istio/test-infra/blob/master/prow/config.yaml) points to files in [istio-releases/daily-release](https://github.com/istio-releases/daily-release/tree/master/prow). Some of these files test in a single namespace but they should be testing cluster wide. This will allow us to remove istio-one-namespace-auth.yaml and istio-one-namespace.yaml from the release archive, as described in istio/istio#4859.

This issue is in conjunction with istio-releases/daily-release#541.

Tahler:
Closing. The needed changes for this repo were merged in.

PR #796, opened 2018-04-13
Daily e2e tests always run cluster-wide (i.e. not one-namespace)

Tahler:
Addresses #795. Will only work against istio-releases/daily-release#532.


prow/config.yaml
```diff
@@ -556,83 +556,15 @@ presubmits:
...
-    agent: kubernetes
-    context: prow/daily-e2e-rbac-auth.sh
-    always_run: true
-    rerun_command: "/test daily-e2e-rbac-auth"
-    trigger: "(?m)^/(retest|test (all|daily-e2e-rbac-auth))?,?(\\s+|$)"
+    rerun_command: "/test daily-e2e"
+    trigger: "(?m)^/(retest|test (all|daily-e2e))?,?(\\s+|$)"
     branches:
     - master
+    - release-0.6
```
rkpagadala: why is this required?
what happens when we run for  branch release-0.8 here

prow/config.yaml
```diff
@@ -648,14 +580,15 @@ presubmits:
...
-      context: prow/daily-e2e-rbac-auth-skew.sh
+      context: prow/daily-e2e-skew.sh
       always_run: true
-      rerun_command: "/test daily-e2e-rbac-auth-skew"
-      trigger: "(?m)^/(retest|test (all|daily-e2e-rbac-auth-skew))?,?(\\s+|$)"
+      rerun_command: "/test daily-e2e-skew"
+      trigger: "(?m)^/(retest|test (all|daily-e2e-skew))?,?(\\s+|$)"
       branches:
       - master
+      - release-0.6
```
rkpagadala: why is this required?
what happens when we run for  branch release-0.8 here

prow/config.yaml
```diff
@@ -671,14 +604,15 @@ presubmits:
...
-    context: prow/daily-e2e-rbac-auth-default.sh
+    context: prow/daily-e2e-default.sh
     always_run: true
-    rerun_command: "/test daily-e2e-rbac-auth-default"
-    trigger: "(?m)^/(retest|test (all|daily-e2e-rbac-auth-default))?,?(\\s+|$)"
+    rerun_command: "/test daily-e2e-default"
+    trigger: "(?m)^/(retest|test (all|daily-e2e-default))?,?(\\s+|$)"
     branches:
     - master
+    - release-0.6
```
rkpagadala: same here

sebastienvas (changes requested):
as I commented in https://github.com/istio-releases/daily-release/pull/532, the PR does not make much sense, we should not be renaming anything, but instead removing the cluster wide one and make the cluster wide default since that s the intent of #795.

prow/config.yaml
Tahler: I'm unsure if or why they are needed. That line was present in the configurations for the daily-e2e-cluster_wide tests and I think I miscommunicated with Seb, thinking we should keep them.

sebastienvas (approved):

PR #800, opened 2018-04-14
Update Mason to check for cluster

sebastienvas:
Fixes https://github.com/istio/istio/issues/4860

### DIFF
--- a/boskos/cmd/mason/Dockerfile
+++ b/boskos/cmd/mason/Dockerfile
@@ -5,8 +5,6 @@ RUN apk update \
   && apk add ca-certificates && rm -rf /var/cache/apk/* \
   && gcloud components install kubectl -q --no-user-output-enabled
 
-RUN gcloud components install kubectl -q
-
 ADD mason /usr/bin/mason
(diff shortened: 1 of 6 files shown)


boskos/cmd/mason/Dockerfile
```diff
@@ -1,6 +1,11 @@
-FROM alpine
+FROM google/cloud-sdk:alpine
 
-RUN apk update && apk upgrade && apk add ca-certificates && rm -rf /var/cache/apk/*
+RUN apk update \
+  && apk upgrade \
+  && apk add ca-certificates && rm -rf /var/cache/apk/* \
+  && gcloud components install kubectl -q --no-user-output-enabled
+
+RUN gcloud components install kubectl -q
```
chxchx: Is the repeatition intentional? The same command is included in the previous RUN clause

sebastienvas: you are right thanks!

sebastienvas:
@chxchx  PTAL

PR #804, opened 2018-04-17
Sisyphus triggers reruns sequentially and reliable PR commit SHA retrieval

chxchx:
This is required as Prow does not support concurrent reruns on the same failed jobs. 

Previously Sisyphus seemed to be stuck in a loop and keeps rerunning the same job at the same commit. The issue is that the `repo-commit` field reported by Prow is bogus and changes after every rerun. This PR also adds a reliable routine to get the actual PR commit SHA from which the rerun is made so we only do the rerun for a pre-specified number of times.

### DIFF
--- a/sisyphus/ci.go
+++ b/sisyphus/ci.go
@@ -166,10 +166,22 @@ func (p *ProwAccessor) GetResult(jobName string, runNo int) (*Result, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Presubmit
-	// "master:241d69701ebf3c57c0d86d016937d23615dae390,554:cc02e59987163c8deb80db8bcc203c318c1b752e"
-	// Postsubmit
-	// "master:241d69701ebf3c57c0d86d016937d23615dae390"
+	/*	The .repos field in started.json is a string to string map that contains exactly
+		one kv pair. For example, a pre-submit run may look like
+			"repos": {
+				"repo": "master:1920a01a8df75baf71ae5fa38d68e4b055bad65c,554:cc02e59987163c8deb80db8bcc203c318c1b752e"
+			}
+		which records the PR number and sha, as well as the base branch and sha.
+		A post-submit run looks like
+			"repos": {
+				"repo": "master:241d69701ebf3c57c0d86d016937d23615dae390"
+			}
+	*/
+	if len(cfg.Repos) != 1 {
+		return nil, fmt.Errorf(
+			"the .repos field in started.json contains %d kv pairs and we expect only one",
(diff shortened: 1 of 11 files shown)


rkpagadala:
lgtm however I want Seb to weigh in

sisyphus/istio_deployment/main.go
```diff
@@ -53,14 +53,12 @@ var (
 	catchFlakesByRun     = flag.Bool("catch_flakes_by_rerun", true, "whether to rerun failed jobs to detect flakyness")
 
 	protectedJobs = []string{
```
sebastienvas: you might able to query those from the github api directly instead of hard coding them

chxchx: Done

chxchx:
PTAL

sisyphus/ci.go
```diff
@@ -161,24 +162,38 @@ func (p *ProwAccessor) GetResult(jobName string, runNo int) (*Result, error) {
...
-	}, nil
+	cfg, err := p.getProwJobConfig(jobName, runNo)
+	if err != nil {
+		return nil, err
+	}
+	// Presubmit
+	// "master:241d69701ebf3c57c0d86d016937d23615dae390,554:cc02e59987163c8deb80db8bcc203c318c1b752e"
+	// Postsubmit
+	// "master:241d69701ebf3c57c0d86d016937d23615dae390"
+	for _, baseSHAprSHA := range cfg.Repos {
```
sebastienvas: Can you just check the length of cfg.Repos and take the first element to make future review easier ?

sebastienvas (changes requested):
minor changes, looks good otherwise.

sisyphus/ci.go
chxchx: I totally agree with you but it seems impossible to access a key-value pair in go map using an index. For example https://groups.google.com/forum/#!topic/golang-nuts/zBcqMsDNt7Q. The return statement in the for loop ensures we do one iteration, and the return statement after the loop generates unfound exception.

sebastienvas: I did not realized it was a map. That brings another issue I guess, how do you know the SHA is the right one ? Is it because in our case the map length is always one ? if that the case please make it as an assumption and fail otherwise.

chxchx: Yes, the map length is always one. I added an explicit check for this and fail otherwise as you suggested. Also added more comments explaining the structure of started.json so people looking at my code have better context of what it is doing. PTAL

sebastienvas (approved):

PR #808, opened 2018-04-19
Rename daily prow jobs to be globally unique

chxchx:
This is required by sisyphus since the GCS path used by Prow makes the assumption that job names are globally unique. As such, sisyphus actually also triggers rerun on istio jobs even though we only enabled it on the daily releases.

Sister PR https://github.com/istio-releases/daily-release/pull/555


rkpagadala (approved):
/lgtm

PR #816, opened 2018-04-26
send pr on branch for daily tests

rkpagadala:
PR for testing daily releases only happens on master branch, with this change we can send pr on any branch.

### DIFF
--- a/toolbox/githubctl/README.md
+++ b/toolbox/githubctl/README.md
@@ -9,16 +9,17 @@ You will need a ```<github token file>``` text file containing the github perona
 $ git clone https://github.com/istio/test-infra.git
 ```
 
-and build it using (might need bazel 0.5.4)
+and build it using
 
 ```
 $ bazel build //toolbox/githubctl
 ```
 
-The binary output is located in bazel-bin/toolbox/githubctl/githubctl.
+The binary output is located in bazel-bin/toolbox/githubctl/linux_amd64_stripped/githubctl.
+
 
 ```
-$ alias githubctl="${PWD}/bazel-bin/toolbox/githubctl/githubctl"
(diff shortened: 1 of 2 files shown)


toolbox/githubctl/main.go
```diff
@@ -245,17 +245,23 @@ func CreateIstioReleaseUploadArtifacts() error {
 // DailyReleaseQualification triggers test jobs buy creating a PR that generates
 // a GitHub notification. It blocks until PR status is known and returns nonzero
 // value if failure. Links to test logs will also be logged to console.
-func DailyReleaseQualification() error {
+func DailyReleaseQualification(baseBranch *string) error {
```
chxchx, line 248: There should be a base_branch flag defined so may be we could call AssertNotEmpty on it and then used its value directly? and let its default value be the master branch so we don't need the if clause checking if baseBranch is nil.

toolbox/githubctl/README.md
```diff
@@ -9,16 +9,17 @@ You will need a ```<github token file>``` text file containing the github perona
 $ git clone https://github.com/istio/test-infra.git
 ```
 
-and build it using (might need bazel 0.5.4)
+and build it using
```
chxchx, line 12: Thank you for updating the README!

toolbox/githubctl/main.go
rkpagadala: there are other places in the code where baseBranch is used, I did not want to go through them. Say What, lets look at it tomorrow.

chxchx: sure but they are not in conflict with each other since only one of the functions will be called, which is selected using `--op=` flag. Flags like `--base_branch` is just a way for users to pass parameters through command line.

PR #819, opened 2018-04-28
Get Latest Green SHA to Kick off Release

chxchx:
The overall overhead is about 2 min despite caching and parallel go routines, which I think still saves us quite some time. It is possible that a green SHA cannot be found given the constraints on `max_commit_depth` and `max_run_depth`. We could set them as large as possible or still keep green build repo around if this happens. 

EDIT: I took another approach to implement this, which caches all prow results to local before any checks on SHA. It now takes 50 seconds to run.

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -80,6 +80,11 @@ func fastForward(repo, baseBranch, refSHA *string) error {
 	return githubClnt.FastForward(*repo, *baseBranch, *refSHA)
 }
 
+type task struct {
+	job       string
+	runNumber int
+}
+
 // preprocessProwResults downloads the most recent prow results up to maxRunDepth
 // then returns a two-level map job -> sha -> passed (true) or failed (false)
 func preprocessProwResults() map[string]map[string]bool {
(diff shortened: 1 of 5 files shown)


toolbox/githubctl/main.go
```diff
@@ -79,167 +80,77 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
+				}
+				if !passed {
+					log.Printf("[%s] failed %s", sha, job)
+					// the race on allChecksPassed is benign
+					allChecksPassed = false
+				} else {
+					log.Printf("[%s] passed %s", sha, job)
+				}
+				wg.Done()
+			}(job)
```
rkpagadala: pass the results map also, that way you avoid races

rkpagadala (changes requested):
not convinced about concurrent access, lets talk on Monday

chxchx:
@rkpagadala PTAL. I took a different approach to implement this. See my updates on the PR description.

rkpagadala:
@chxchx lets talk about this monday

/hold

chxchx:
PTAL @rkpagadala

toolbox/githubctl/main.go
```diff
@@ -79,167 +80,97 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
+func preprocessProwResults() map[string]map[string]bool {
+	log.Printf("Start preprocessing prow results")
+	prowAccessor := s.NewProwAccessor(
+		prowProject,
+		prowZone,
+		gubernatorURL,
+		gcsBucket,
+		u.NewGCSClient(gcsBucket))
+	cache := make(map[string]map[string]bool)
+	tasks := make(chan task)
```
rkpagadala: Add size, make the task a pointer, name the channel as tasksCh

tasksCh := make(chan *task, *maxConcurrentRequests)

toolbox/githubctl/main.go
```diff
@@ -79,167 +80,97 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
-}
-
-// UpdateIstioVersionAfterReleaseTagsMadeOnDeps runs updateVersion.sh to update
-// istio.VERSION and then create a PR on istio
-func UpdateIstioVersionAfterReleaseTagsMadeOnDeps() error {
-	u.AssertNotEmpty("base_branch", baseBranch)
-	releaseTag, err := getReleaseTag()
-	if err != nil {
-		return err
+			wg.Done()
```
rkpagadala: move this to just before the for loop start and make it

defer wg.Done()

toolbox/githubctl/main.go
```diff
@@ -79,167 +80,97 @@ func fastForward(repo, baseBranch, refSHA *string) error {
...
-			istioRepo, releaseTag, *refSHA, archiveDir); err != nil {
-			return err
+	for i := 0; i < *maxCommitDepth; i++ {
+		log.Printf("Checking if [%s] passed all checks. %d commits before HEAD", sha, i)
+		allChecksPassed := true
+		for _, job := range postSubmitJobs {
+			passed, keyExists := results[job][sha]
+			if !keyExists {
+				log.Printf("Results unknown in local cache for [%s] at [%s], treat the test as failed", job, sha)
+				passed = false
```
rkpagadala: you dont need this line, when key does not exist you get default value which is false in this case

rkpagadala (commented):
otherwise lgtm. Thanks for doing this.

toolbox/githubctl/main.go
chxchx: done

toolbox/githubctl/main.go
chxchx: done

toolbox/githubctl/main.go
chxchx: done.

Passing pointers instead reduces the total execution time by half. Awesome.

PR #822, opened 2018-05-01
add git_branches for green build and update bot

rkpagadala:
add git_branches for green build and update bot, this allows for less effort every new release


rkpagadala:
related change https://github.com/istio/global-build/pull/11

PR #830, opened 2018-05-04
Longer e2e timeout to avoid premature abortion

chxchx:
Nowadays the e2e test suite takes about an hour to finish, and the timeout is set to be an hour. Hence we see quite a few timeout termination on otherwise healthy tests, which in turns failed the daily releases (https://k8s-testgrid.appspot.com/istio#daily-e2e-rbac-no_auth). This PR intends to extend the timeout to accommodate lengthy tests.


hklai:
Shall we try 90 mins first?

chxchx:
aye aye

PR #835, opened 2018-05-09
Revert "Remove kubeconfig for unit tests since its not being used anyway"

sebastienvas:
Reverts istio/test-infra#834


sebastienvas:
this is apparently creating issues

PR #836, opened 2018-05-10
Push presubmit results to different GCS path so testgrid could multiplex

chxchx:
The goal is to create multiple testgrid panels for istio based on attributes such as presubmit vs postsubmits, and each panel could use different gcs prefix to multiplex the test results.

### DIFF
--- a/toolbox/ci2gubernator/README.md
+++ b/toolbox/ci2gubernator/README.md
@@ -45,7 +45,10 @@ The binary `ci_to_gubernator` constructs these files and uploads these artifacts
 
 ## Usage
 
-At the start of a build, execute the following command to create and upload `started.json` on GCS.
+At the start of a build, execute the following command to create and upload `started.json` on GCS. If the job is running as part of the presubmit, one should
+also specify the `--presubmit` flag. Presubmits are uploaded to a different
+locations on GCS so we could have different panels on testgrid that makes
+multiplexing results attainable.
 
 ```bash
 $ ci_to_gubernator --job_starts \
@@ -62,7 +65,7 @@ At the end of a build, execute the following command to
 * upload `finished.json`, `artifacts/junit_*.xml`, and `build-log.txt` to GCS
 * update `latest-build.txt` of this job with the current build number
 
-Caveat: this command exits with the same exit code as supplied by --exit_code.
+The same rules on `--presubmit` usage apply in here.
 
 ```bash
 $ ci_to_gubernator \
--- a/toolbox/ci2gubernator/README.md
+++ b/toolbox/ci2gubernator/README.md
@@ -45,8 +45,9 @@ The binary `ci_to_gubernator` constructs these files and uploads these artifacts
 
 ## Usage
 
-At the start of a build, execute the following command to create and upload `started.json` on GCS. If the job is running as part of the presubmit, one should
-also specify the `--presubmit` flag. Presubmits are uploaded to a different
+At the start of a build, execute the following command to create and upload `started.json` on GCS.
+If the job is running as part of the presubmit, one should
+also specify the `--stage=presubmit` flag. Presubmits are uploaded to a different
 locations on GCS so we could have different panels on testgrid that makes
 multiplexing results attainable.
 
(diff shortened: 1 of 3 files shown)


toolbox/ci2gubernator/README.md
```diff
@@ -45,7 +45,10 @@ The binary `ci_to_gubernator` constructs these files and uploads these artifacts
 
 ## Usage
 
-At the start of a build, execute the following command to create and upload `started.json` on GCS.
+At the start of a build, execute the following command to create and upload `started.json` on GCS. If the job is running as part of the presubmit, one should
```
hklai: perhaps fix the line breaks to make the paragraph consistent?

toolbox/ci2gubernator/lib/ci2gubernator.go
```diff
@@ -67,6 +68,10 @@ func NewConverter(bucket, org, repo, job string, build int) *Converter {
 		job:           job,
 		build:         build,
 	}
+	if presubmit {
```
hklai: instead of a boolean, how about using a suffix string so it becomes reusable for other cases as well? And then you can just put everything in the same sprintf line, assuming the default suffix is just an empty string.

toolbox/ci2gubernator/README.md
chxchx: Done

toolbox/ci2gubernator/lib/ci2gubernator.go
chxchx: Done

toolbox/ci2gubernator/README.md
```diff
@@ -62,7 +66,7 @@ At the end of a build, execute the following command to
 * upload `finished.json`, `artifacts/junit_*.xml`, and `build-log.txt` to GCS
 * update `latest-build.txt` of this job with the current build number
 
-Caveat: this command exits with the same exit code as supplied by --exit_code.
+The same rules on `--presubmit` usage apply in here.
```
hklai: this is out of date.. In fact I am not sure if we need this sentence at all.

toolbox/ci2gubernator/lib/ci2gubernator.go
```diff
@@ -57,10 +57,10 @@ type Converter struct {
 }
 
 // NewConverter creates a Converter
-func NewConverter(bucket, org, repo, job string, build int) *Converter {
+func NewConverter(bucket, org, repo, job, mux string, build int) *Converter {
```
hklai: why is "mux"?

toolbox/ci2gubernator/lib/ci2gubernator.go
```diff
@@ -57,10 +57,10 @@ type Converter struct {
...
 	return &Converter{
 		gcsClient:     u.NewGCSClient(bucket),
-		gcsPathPrefix: fmt.Sprintf("%s/%d", job, build),
+		gcsPathPrefix: fmt.Sprintf("%s/%s/%d", mux, job, build),
```
hklai: if mux is empty the resulting string will have two slashes. will that path work?

toolbox/ci2gubernator/README.md
chxchx: Good catch. Fixed it in latest commit, but I would prefer to keep this line just to remind the users to not miss it.

toolbox/ci2gubernator/lib/ci2gubernator.go
chxchx: MUX is short for multiplexer but I changed to `stage` to be consistent with the main file and to be more self-explanatory.

toolbox/ci2gubernator/lib/ci2gubernator.go
chxchx: Yes because mux is the first input to the format string so if it is empty there will only be a leading slash, and filepath.join will handle this. However, I changed it to use filepath.join directly instead of sprintf to make this case more evident

toolbox/ci2gubernator/lib/ci2gubernator.go
hklai: Have you not pushed the commit for this change?

hklai (approved):
/lgtm

PR #840, opened 2018-05-17
Fix workspace

sebastienvas:


rkpagadala (approved):
/lgtm
the order matters?

chxchx:
@rkpagadala I guess the linter enforces such order and if not fixing it, it breaks everyone's presubmit.

PR #844, opened 2018-05-17
Enable testing version of tide on istio and proxy

sebastienvas:


prow/config.yaml
```diff
@@ -21,20 +21,46 @@ branch-protection:
         contexts:
         - cla/google
       repos:
-        test-infra:
+        istio:
           protect: true
+          required_status_check:
+            contexts:
+            - "ci/circleci: build"
```
hklai, line 28: why do we need to hardcode them here?

PR #849, opened 2018-05-22
Adds 1.10 cluster for testing

sebastienvas:


sebastienvas:
cc @ldemailly

hklai (approved):
/lgtm

ldemailly:
awesome, where/how does it run/get used?
are the reports that v2 isn't working with 1.10 true?

sebastienvas:
still need to create a postsubmit. Which job should I duplicate ?

ldemailly:
rbac-auth and some v2 pilot test(s) too ?

PR #854, opened 2018-05-25
disable update bot 

rkpagadala:


qiwzhang:
You may want to wait for this change.   I am pretty sure the current proxy will break many mixer client tests in istio.

PR #857, opened 2018-05-26
Block merge to istio/istio:release-0.8, and open up master.

hklai:
release-0.8 will be moved to the common block when the branch is also blocked in istio.github,io.


sebastienvas:
close in favor of https://github.com/istio/test-infra/pull/858

PR #861, opened 2018-05-30
Fix issue where bad cluster would still make it

sebastienvas:
Also added a logging statement to understand which project are encountering the issue:
Unable to connect to the server: dial tcp 35.233.173.199:443: i/o timeout

Updated Vendors for k8s.io/test-infra


hklai (approved):
/lgtm

PR #862, opened 2018-05-31
 ci2gubernator to force unlock on timeout, and create latest-build.txt if nonexistent 

chxchx:


hklai:
I think there are other underlying problems....

1. if latest-build.txt  does not exist, the job will fail.
2. test job should not fail even if upload/update fails.


Can we address those as well?

hklai:
As for the timeout, we prob want to figure out some load time variances. I don't want to block the tests for 30 mins just waiting for the upload. We may as well just timeout and lose one data point...

sebastienvas:
Fixed the failure, I accidently rebased my personal kubernetes/test-infra:boskos. I am now protecting that branch until I get my PR in the kubernetes repo. https://github.com/kubernetes/test-infra/pull/8182

chxchx:
@hklai You are right about the timeout not being the issue. As I explained in here https://github.com/istio/istio/pull/5961#issuecomment-393959015 This was caused by a persisting lock file that renders all subsequent lock acquire requests to block until timeout. The fix is to force cleaning the lock file after 5 minute timeout, since usually the lock is held for seconds. 

This PR also addressed your first concern by creating latest-build.txt if it does not exist. 

Your second concern is already fixed by https://github.com/istio/istio/pull/5961.

@sebastienvas Thanks for fixing the presubmit and making your branch protected!

toolbox/ci2gubernator/lib/ci2gubernator.go
```diff
@@ -136,14 +136,28 @@ func (c *Converter) UpdateLastBuildTXT() error {
 		return err
 	}
 	if err = m.ContextLock(ctx); err != nil {
+		// force resetting lock state on timeout
+		m.Unlock()
```
hklai: Is this supposed to try acquire the lock again after force unlocking?
If not, multiple jobs can write at the same time.

However, if you do that, what happens when multiple jobs are waiting and they all try to unlcok and then lock? Are they going to interfere with each other?

sebastienvas: I am just thinking that the lock is not very necessary in this case. It is ok if the last writer wins, it is eventually consistent. That s how bootstrap.py is doing it.

chxchx: I update the code to retry acquiring the lock after force removal of lock after initial lock() request timeout. I take a look on the implementation of ContextLock() and it is doing retries with exponential backoffs until the given context timeout so once the first failing client remove the lock file, other pending clients who have not return from ContextLock() will retry and get the lock sequentially. 

I would agree with Seb that it is okay for latest-build.txt to be a bit off, since testgrid does not depend on this file it turns out. Only concern is that any future tools that we use might have different assumptions and debugging then would be hard. I guess if the locking scheme becomes an issue again we could go ahead and remove the code.

chxchx: Okay it seems bootstrap.py used by k8s is also assuming last writer wins forever, which means their tooling has to make the same assumption. So I guess it is up to us whether we want to keep the lock.

hklai:
Let's not add the locking logic until we know we need it. =)

chxchx:
@hklai PTAL

toolbox/ci2gubernator/lib/ci2gubernator.go
```diff
@@ -129,21 +122,18 @@ func (c *Converter) UploadBuildLog(logFile string) error {
...
+	gcsPath := filepath.Join(filepath.Dir(c.gcsPathPrefix), lastBuildTXT)
+	update := func() error {
+		log.Printf("Updating gs://%s to be %d", gcsPath, c.build)
+		return c.gcsClient.Write(gcsPath, fmt.Sprintf("%d", c.build))
+	}
+	exists, err := c.gcsClient.Exists(gcsPath)
 	if err != nil {
 		return err
 	}
-	if err = m.ContextLock(ctx); err != nil {
```
hklai: seems easier to read with a more linear flow:


if exists && needUpdate() {
  log.Printf("Updating gs://%s to be %d", gcsPath, c.build)
  return c.gcsClient.Write(gcsPath, fmt.Sprintf("%d", c.build))
}

and then implement needUpdate() as a helper method...

toolbox/util/gcsClient.go
```diff
@@ -48,6 +50,23 @@ func NewGCSClient(bucket string) *GCSClient {
 	}
 }
 
+// Exists finds if an object already exists on GCS bucket
+func (gcs *GCSClient) Exists(obj string) (bool, error) {
```
hklai, line 54: do we need this or can we derive this from the error received while reading?

toolbox/ci2gubernator/lib/ci2gubernator.go
chxchx: Done

toolbox/util/gcsClient.go
chxchx: Go does not distinguish or allow extending `error` so to do it that way we have to hack it with string matching, which to me seems kind of sketchy since the error message is meant for human consumption and could change any minute from server side.

hklai (approved):
/lgtm

PR #863, opened 2018-05-31
Remove v1 tests from required context.

hklai:
@sebastienvas  Is this the right way to make test not required? And how come the prow tests are not on the list?


sebastienvas:
closed in favor of https://github.com/istio/test-infra/pull/865

PR #864, opened 2018-06-01
Mason config

sebastienvas:
All the cluster and vms were being created on the same zone. Updating the config such that it will query the zones available for a given projects and round robing around the zones for clusters and vms creation.

Note that when we query the zone per project, we shuffle the zones, such that each cluster and vm creation on different project don't use the same zone. If this problem persists we could consider querying GCP to get zones where it is less likely to get a quota or failure.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -487,7 +487,7 @@
     "prow/slack",
     "testgrid/util/gcs"
   ]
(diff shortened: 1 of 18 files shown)


hklai:
Could you add more information about why the zone was the problem? Otherwise the change lgtm,

hklai (approved):
/lgtm

PR #871, opened 2018-06-04
Not running e2e v1 tests

sebastienvas:


rkpagadala (commented):
what happens when "branches" has "master" and skip_branches also has master? which one takes precendence?

prow/config.yaml
```diff
@@ -421,7 +421,8 @@ presubmits:
     - name: istio-pilot-e2e
       agent: kubernetes
       context: prow/istio-pilot-e2e.sh
-      optional: true
```
hklai, line 424: is optional supposed to be branch specific?
Can we make it optional in one branch but not the other? 

And now since optional: true is removed, does it mean it is required in branches like release-0.8?

sebastienvas: optional is not branch specific. So yes it will be considered required on all branches in that case.

sebastienvas: which was the case for 0.8

sebastienvas:
@hklai ping

hklai (approved):
/lgtm

PR #876, opened 2018-06-06
Using vendored deps for bazel

sebastienvas:
Bazel has pretty good support for vendors now.


kyessenov:
Nice :) I hear docker sandbox support is also native now (overheard grievances in envoy community).

PR #877, opened 2018-06-07
Githubctl uses V-style logging and Less verbose by default

chxchx:


sebastienvas:
why not switching everything the same logging library ?

chxchx:
@sebastienvas That is just an obvious point that I missed. Did just that.

PR #884, opened 2018-06-12
Mark ci/circleci: e2e-pilot-noauth-v1alpha3-v2 as required.

incfly:
For https://github.com/istio/istio/issues/6196

We could remove this from required in case the test is not stable enough. But given it's been in release-0.8 for quite a while, it should be fine to be required.


incfly:
PTAL @hklai  thanks!

hklai (approved):
/lgtm

PR #886, opened 2018-06-14
Reduce syncs to every 10 minutes to ease Boskos Server

sebastienvas:
Simple workaround until we implement a better boskos client that should do that for us.


hklai (approved):
/lgtm

PR #888, opened 2018-06-15
 Making mixer test not required since it is flaky.

hklai:


kyessenov:
May I suggest putting `t.Skip` for the two flaky tests? The other 8 in the suite are not flaky.

rshriram (approved):

rshriram:
t.skip sounds good. but it will still not be required until we see the tests pass reliably.

kyessenov:
They are a pre-requisite for the daily build. So it's the highest priority to fix them.

hklai:
So far I have not seen failures other than the two.
Let's skip them for now and re-evaluate again later today.
https://github.com/istio/istio/pull/6326

PR #894, opened 2018-06-19
Add noauth pilot e2e test as required again.

incfly:
It's been stable for a while, https://k8s-testgrid.appspot.com/istio-presubmits#circleci-e2e-pilot-noauth-v1alpha3-v2-presubmit.

In case of anyone complaining about flakiness, we can skip the flaky test rather than skip the whole test suite.

Fixe https://github.com/istio/istio/issues/6196


incfly:
@hklai

PR #898, opened 2018-06-21
Adding dashboard e2e tests to presubmit.

hklai:
Also removing e2e-suites from postsubmit. Instead run individual e2e tests.


rkpagadala (approved):
/lgtm

PR #900, opened 2018-06-22
Green sha script reads post submit list from prow config

chxchx:
There were new tests added as postsubmits and this is just to make sure things are consistent. 

I tested locally and there is no green sha yet, since recent commits consistently failed on `istio-pilot-e2e-envoyv2-v1alpha3-k8s-1.10` and `e2e-dashboard`. I will mark this PR as on hold and will merge after tests become stable.

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -46,7 +46,6 @@ var (
 	maxConcurrentRequests = flag.Int("max_concurrent_reqs", 50, "Max number of concurrent requests permitted")
 	githubClnt            *u.GithubClient
 	ghClntRel             *u.GithubClient
-	postSubmitJobs        = []string{}
 )
 
 const (
@@ -84,7 +83,7 @@ type task struct {
 
 // preprocessProwResults downloads the most recent prow results up to maxRunDepth
 // then returns a two-level map job -> sha -> passed (true) or failed (false)
-func preprocessProwResults() map[string]map[string]bool {
+func preprocessProwResults(postSubmitJobs []string) map[string]map[string]bool {
 	glog.Infof("Start preprocessing prow results")
 	prowAccessor := s.NewProwAccessor(
 		prowProject,
@@ -116,7 +115,7 @@ func preprocessProwResults() map[string]map[string]bool {
 			}
 		}()
 	}
-	// note: if postSubmitJobs was not found in map, the for loop exits immediately
+	// if postSubmitJobs is empty, the for loop exits immediately
 	for _, job := range postSubmitJobs {
 		cache[job] = make(map[string]bool)
 		runNumber, err := prowAccessor.GetLatestRun(job)
@@ -139,8 +138,8 @@ func getLatestGreenSHA() (string, error) {
 	u.AssertNotEmpty("base_branch", baseBranch)
 	u.AssertPositive("max_commit_depth", maxCommitDepth)
 	u.AssertPositive("max_run_depth", maxRunDepth)
-	postSubmitJobs = readPostsubmitListFromProwConfig(*owner, *repo, *baseBranch)
-	results := preprocessProwResults()
+	postSubmitJobs := readPostsubmitListFromProwConfig(*owner, *repo, *baseBranch)
+	results := preprocessProwResults(postSubmitJobs)
 	sha, err := githubClnt.GetHeadCommitSHA(*repo, *baseBranch)
 	if err != nil {
 		glog.Fatalf("failed to get the head commit sha of %s/%s: %v", *repo, *baseBranch, err)
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -116,7 +116,6 @@ func preprocessProwResults() map[string]map[string]bool {
 			}
 		}()
 	}
-	postSubmitJobs := postSubmitJobsMap[*baseBranch]
 	// note: if postSubmitJobs was not found in map, the for loop exits immediately
 	for _, job := range postSubmitJobs {
 		cache[job] = make(map[string]bool)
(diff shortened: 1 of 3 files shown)


toolbox/githubctl/main.go
```diff
@@ -47,9 +47,12 @@ var (
 	// needs to be consistent with prow config map
 	postSubmitJobs = []string{
```
sebastienvas: Why not read from the config map direclty ?

sebastienvas: https://github.com/kubernetes/test-infra/blob/master/prow/config/config.go#L263

chxchx: Thanks for the pointer and that is exactly what I plan to do too. This PR tries to unblock release while the work on reading config is pending. @rkpagadala has another PR this does this as well and he takes care of branches too so lets proceeds with his.

chxchx:
PTAL

toolbox/githubctl/main.go
```diff
@@ -303,6 +277,39 @@ func DailyReleaseQualification(baseBranch *string) error {
 	return fmt.Errorf("release qualification failed")
 }
 
+func traverseJobTree(job config.Postsubmit, postsubmitJobs *[]string, targetBranch string) {
+	for _, branch := range job.Brancher.Branches {
```
sebastienvas: you want to use ```RunsAgainstBranch``` instead 
https://github.com/kubernetes/test-infra/blob/master/prow/config/jobs.go#L194

toolbox/githubctl/main.go
```diff
@@ -303,6 +277,39 @@ func DailyReleaseQualification(baseBranch *string) error {
...
+}
+
+func readPostsubmitListFromProwConfig(org, repo, branch string) []string {
+	var postsubmitJobs []string
+	repoRootDir, err := u.Shell("git rev-parse --show-toplevel")
+	repoRootDir = strings.TrimSuffix(repoRootDir, "\n")
+	if err != nil {
+		glog.Fatalf("cannot find repo root directory path")
+	}
+	prowConfigYaml := filepath.Join(repoRootDir, "prow/config.yaml")
```
sebastienvas, line 302: I think in golang you can give relative path. Just add it as a const at the top of the file.

rkpagadala:
One of the problems I see with this approach is the inability to override. For e.g. istio-pilot-e2e-envoyv2-v1alpha3-k8s-1.10 is not expected to pass in post submit right now

https://k8s-gubernator.appspot.com/build/istio-prow/istio-pilot-e2e-envoyv2-v1alpha3-k8s-1.10/275/

sebastienvas:
Where is this thing running by the way ?

rkpagadala:
it is run as part of green build right now, it would be good to have flexibility to run it as part of Release Process (Airflow)

toolbox/githubctl/main.go

toolbox/githubctl/main.go
chxchx: Yes relative path works but I think absolute path is better because it is less susceptible to directory restructuring. If this go file is moved elsewhere it would still work.

chxchx:
@rkpagadala I added a new `--skip` flag for the purpose of override. Also update the documentation on its use. PTAL

toolbox/githubctl/main.go
```diff
@@ -38,37 +41,13 @@ var (
...
-			"istio-postsubmit",
-		},
-		"release-0.8": {
-			"istio-postsubmit",
-			"e2e-suite-rbac-no_auth",
-			"e2e-suite-rbac-auth",
-			"e2e-cluster_wide-auth",
-		},
-	}
+	postSubmitJobs        = []string{}
```
rkpagadala: remove this global and pass it around. otherwise lgtm

rkpagadala (approved):

toolbox/githubctl/main.go

toolbox/githubctl/main.go
```diff
@@ -303,6 +281,36 @@ func DailyReleaseQualification(baseBranch *string) error {
 	return fmt.Errorf("release qualification failed")
 }
 
+func traverseJobTree(job config.Postsubmit, postsubmitJobs *map[string]struct{}, targetBranch string) {
+	if job.Brancher.RunsAgainstBranch(targetBranch) {
+		(*postsubmitJobs)[job.Name] = struct{}{}
```
rkpagadala: what happens if you rewrite 
(*postsubmitJobs)[job.Name] = struct{}{}
as
postsubmitJobs[job.Name] = struct{}{}

chxchx: Without dereferencing, I got 
`
test_infra/toolbox/githubctl/main.go:286:17: invalid operation: postsubmitJobs[job.Name] (type *map[string]struct {} does not support indexing)
`

sebastienvas: even if you remove * from the function def ? a map is already a pointer.

toolbox/githubctl/main.go
```diff
@@ -303,6 +281,36 @@ func DailyReleaseQualification(baseBranch *string) error {
 	return fmt.Errorf("release qualification failed")
 }
 
+func traverseJobTree(job config.Postsubmit, postsubmitJobs *map[string]struct{}, targetBranch string) {
```
rkpagadala: Thanks Seb.
Charles, the function definition can be changed to 
func traverseJobTree(job config.Postsubmit, postsubmitJobs map[string]struct{}, targetBranch string) {
notice that postsubmitJobs is declared as a map not a pointer to a map.

toolbox/githubctl/main.go
rkpagadala: :) nice catch Seb

toolbox/githubctl/main.go

PR #902, opened 2018-06-23
index postsubmit job by branch

rkpagadala:
index postsubmit job by branch so that we can look for correct green sha based on the tests required for  a particular branch

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -45,11 +45,25 @@ var (
 	ghClntRel             *u.GithubClient
 	// unable to query post-submit jobs as GitHub is unaware of them
 	// needs to be consistent with prow config map
-	postSubmitJobs = []string{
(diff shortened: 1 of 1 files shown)


sebastienvas:
https://github.com/kubernetes/test-infra/blob/master/prow/config/config.go#L263

rkpagadala:
@chxchx please follow this up by reading from config map

toolbox/githubctl/main.go
```diff
@@ -120,6 +138,7 @@ func preprocessProwResults() map[string]map[string]bool {
 			}
 		}()
 	}
+	postSubmitJobs := postSubmitJobsMap[*baseBranch]
```
sebastienvas, line 141: you mgith want check that the key exist, and fail otherwise

rkpagadala: thanks seb. fixed it.

PR #905, opened 2018-06-25
adds CF test back as required

zachgersh:
This rarely fails, unless something is actually broken, and should be a default test that must pass.

Co-authored-by: Utako Ueda <uueda@pivotal.io>


hklai (approved):
/lgtm

PR #908, opened 2018-06-25
Extending prometheus metrics server with flakiness

chxchx:
After https://github.com/kubernetes/test-infra/pull/8424 is merged, data will be available in bucket `k8s-metrics`. Until then, for testing purposes, could use bucket `istio-flakes`.

### DIFF
--- a/toolbox/metrics/flakes/flakes.go
+++ b/toolbox/metrics/flakes/flakes.go
@@ -54,7 +54,7 @@ func NewMetric() *FlakeGauge {
 		gauge: prometheus.NewGaugeVec(
 			prometheus.GaugeOpts{
 				Name: "consistency",
-				Help: "In past 7 days, Number of successful runs over number of total runs",
+				Help: "In yesterday, Number of successful runs over number of total runs",
 			},
 			[]string{"job"},
 		),
--- a/toolbox/metrics/flakes/flakes.go
+++ b/toolbox/metrics/flakes/flakes.go
@@ -88,7 +88,7 @@ func (f *FlakeGauge) update() error {
 
 // Update implements metrics.Metric interface
 func (f *FlakeGauge) Update(ctx context.Context) error {
-	glog.Infof("upading flakes matrics")
+	glog.Infof("updaing flakes matrics")
 	errc := make(chan error)
 	go func() {
 		errc <- f.update()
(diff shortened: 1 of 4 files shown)


toolbox/metrics/flakes/flakes.go
```diff
@@ -0,0 +1,102 @@
...
+
+	u "istio.io/test-infra/toolbox/util"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	bucket             = "k8s-metrics"
+	latestFlakesMetric = "istio-job-flakes-latest.json"
```
sebastienvas, line 30: Just for information, who is creating this? Is there a way to get flakiness per org/repo/branch ?

toolbox/metrics/flakes/flakes.go
```diff
@@ -0,0 +1,102 @@
...
+		} else {
+			f.gauge.WithLabelValues(flake.Job).Set(consistency)
+		}
+	}
+	return nil
+}
+
+// Update implements metrics.Metric interface
+func (f *FlakeGauge) Update(ctx context.Context) error {
+	glog.Infof("upading flakes matrics")
```
sebastienvas: typo: updating

chxchx: Done

toolbox/metrics/flakes/flakes.go
chxchx: The PR linked in this PR description adds a new metric config to https://github.com/kubernetes/test-infra/tree/master/metrics/configs and k8s infra team has already set up a job to build all metrics every 24 hours and upload results to `gs://k8s-metrics`. That PR is still pending, and for the purpose of an example, an output that I created is at `gs://istio-flakes/istio-job-flakes-latest.json`.

I think it is possible to get this metric per repo and/or per branch, as such information is probably tracked in the [big query instance](https://bigquery.cloud.google.com/table/k8s-gubernator:build.all?pli=1). However, this metric that I added is meant for just istio because of the way we use Prow is different from k8s. Existing flakiness metric config contains SQL query that depends on differences between job names for presubmis and postsubmits (`ci-`/`pr-` prefix) and the `version` field must not be "unknown", neither is true for istio. So doing this per org is out of the question in this case.

sebastienvas: this might get resolved when we switch to podutils.

PR #914, opened 2018-06-28
New metrics image that include flakes data

chxchx:


sebastienvas:
is this already deployed ?
/lgtm

chxchx:
Yes it has been deployed and working http://35.199.148.121/metrics

I am having some trouble configuring velodrome frontend. It seems data was missing from the point of view of velodrome. CodeCov seems also missing data, so the problem should be in prometheus mertric server, but it works fine when you ping it with browser. I work with you in person when you have time.

PR #918, opened 2018-07-02
Stop merge bot from merging master and release-1.0 branch

hklai:

PR #919, opened 2018-07-02
Disable auto merge in master and release-1.0, and add release 1.0 test config

rkpagadala:
change config so that we can test 1.0.0 on istio and daily-releases.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -29,7 +29,7 @@ blocked_branches: &blocked_branches
   release-0.7: *blocking_merge
   release-0.8: *blocking_merge
   release-1.0.0-snapshot.0: *blocking_merge
-  release-1.0.0: *blocking_merge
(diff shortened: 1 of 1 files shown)


hklai:
hmm... we are touching the same file and this has similar thing that I have already done in https://github.com/istio/test-infra/pull/918...

prow/config.yaml
```diff
@@ -405,7 +406,7 @@ presubmits:
       trigger: "(?m)^/test (istio-pilot-e2e-envoyv2-v1alpha3)?,?(\\s+|$)"
       branches:
       - release-0.8
-      - release-1.0.0-snapshot.0
```
hklai: Let's add release-1.0 instead. It should be release-1.0 instead of release-1.0.0.

hklai (approved):

PR #920, opened 2018-07-03
Update prowconfig for branch protector

sebastienvas:
There was some inconsistency in the config files. Makefile would create a config file, while the prow config update plugin would update config.yaml. Update makefile to also create a config.yaml and updated branch protector to use that file as well.

### DIFF
--- a/prow/cluster/branchprotector_cronjob.yaml
+++ b/prow/cluster/branchprotector_cronjob.yaml
@@ -16,7 +16,7 @@ spec:
           - name: branchprotector
             image: gcr.io/k8s-prow/branchprotector:v20180620-3b98af6f3
             args:
-            - --config-path=/etc/config/config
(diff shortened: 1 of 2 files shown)


prow/cluster/branchprotector_cronjob.yaml
```diff
@@ -16,7 +16,7 @@ spec:
           - name: branchprotector
             image: gcr.io/k8s-prow/branchprotector:v20180620-3b98af6f3
             args:
-            - --config-path=/etc/config/config
+            - --config-path=/etc/config/config.yaml
```
hklai, line 19: and this is copied from prow/config.yaml?

hklai (commented):
/lgtm

hklai (approved):
/lgtm

Issue #916, opened 2018-06-30
Evaluate the need for alpha cluster for running Istio unit tests


hklai:
In the past Istio tests relied on some alpha features, but it should not be the case anymore.

If so, we should move away from using alpha clusters which get deleted every 30 days. Because of the automatic deletion, Istio test had been failing for almost 20 hours and many developers were blocked.

![7gdnhwgfvbc](https://user-images.githubusercontent.com/31747472/42119850-b774cfe2-7bc5-11e8-8b6d-58dacd06bb13.png)

And in case alpha clusters are needed, they should be renewed automatically to minimize outages.

sebastienvas:
Also recreated the main clusters to not be alpha.

PR #923, opened 2018-07-06
Remove alpha features from cluster

sebastienvas:
Removing unnecessary alpha features.

Might help with https://github.com/istio/istio/issues/6838

For boskos, we would need to over the resources to have non alpha clusters, so it might take a couple days to take effect.


rkpagadala (approved):
/lgtm

chxchx:
fixes https://github.com/istio/test-infra/issues/916

PR #924, opened 2018-07-09
run proxy postsubmit on sds_api branch

rkpagadala:
run proxy postsubmit on sds_api branch as per request from security team

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -967,7 +967,8 @@ postsubmits:
     always_run: true
     rerun_command: "/test proxy-postsubmit"
     trigger: "(?m)^/(retest|test (all|proxy-postsubmit))?,?(\\s+|$)"
-    branches: sds_api
+    branches:
+    - sds_api
     labels:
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -961,6 +961,40 @@ postsubmits:
       nodeSelector:
         cloud.google.com/gke-nodepool: build-pool
 
+  - name: proxy-postsubmit
+    agent: kubernetes
+    context: prow/proxy-postsubmit.sh
+    always_run: true
+    rerun_command: "/test proxy-postsubmit"
+    trigger: "(?m)^/(retest|test (all|proxy-postsubmit))?,?(\\s+|$)"
```
qiwzhang, line 969: I am not clear how the test will be triggered.  Does it always need a PR?  How about if we merge the changes directly without PR(we can do that since it is a private branch).

rkpagadala: I think it will run on commit, this is not related to PR

PR #928, opened 2018-07-13
Remove the "needs-ok-to-test" check in Tide config for istio.github.io.

hklai:
There are a lot of outside contributor fixing docs so we should not
block merges using that label.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -130,7 +130,6 @@ tide:
   - repos:
     - istio/api
     - istio/test-infra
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -143,14 +143,24 @@ tide:
     - do-not-merge/work-in-progress
```
sebastienvas, line 142: you need to remove istio/istio.github.io from the previous query

hklai: Done

PR #941, opened 2018-08-03
Update tide to working version

sebastienvas:
Previous version had an issue assuming job name and context would be the same.


rkpagadala (approved):
/lgtm
/approve

PR #944, opened 2018-08-07
Assorted fixes to test-infra

rshriram:
Do not self approve PRs
Make codecov optional
Make multi-cluster optional (i.e. not required)
Do not allow repo-admins to dismiss pull-request reviews

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -105,6 +105,7 @@ presubmits:
       branches: *branch_spec
       agent: kubernetes
       context: prow/istio-pilot-multicluster-e2e.sh
+      optional: true
       rerun_command: "/test istio-pilot-multicluster-e2e"
       trigger: "(?m)^/test (istio-pilot-multicluster-e2e)?,?(\\s+|$)"
       labels:
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -51,9 +51,6 @@ branch-protection:
         - cla/google
       required_pull_request_reviews:
         required_approving_review_count: 1
-        dismissal_restrictions:
-          teams:
-          - repo-admins
       restrictions:
         teams:
         - repo-admins
@@ -69,7 +66,6 @@ branch-protection:
           required_status_checks:
             contexts:
             - "ci/circleci: build"
-            - "ci/circleci: codecov"
             - "ci/circleci: e2e-simple"
             - "ci/circleci: lint"
             - "ci/circleci: test"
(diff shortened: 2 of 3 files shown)


prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -105,6 +105,7 @@ presubmits:
       branches: *branch_spec
       agent: kubernetes
       context: prow/istio-pilot-multicluster-e2e.sh
+      optional: true
```
tiswanso, line 108: Multicluster is important to my org.  I'd like a process to get this on the required test list / merge gate.

sdake: @costinm - Isn't it accurate that promotions of e2e tests to required require TOC approval?  I think you mentioned this in relationship to coverage test cases which were moved to approved without TOC oversight.  In this case, optional should have been where this test started out - however, I think @tiswanso struggles to find the prow documentation so this PR could have started out optional (as it should have).

rshriram: We can change this to required only after the test has been proven to be stable for a week. Otherwise, we risk a flood of PRs stuck on this. Flipping the required bit is a small change that should not require PRs to rebase.

While you may have tested this a couple of times, we need to see how this holds up when we have 80+ PRs, and several of them doing /retest. prow e2e tests are barely holding up for single cluster tests, and take a very long time.

IF you want this test to gate a release, then please add it to the daily release tasks. Those run in the background and can have as many required tests as we wish.

tiswanso: @rshriram -- Yep.. agree with the points.  Admittedly, I didn't expect the test-infra PR to merge so quickly after submitting so unfortunately this discussion didn't happen in that context.  My appologies for the noise.  This test has been pending merge for months and I've been triggering it manually basically multiple times a day throughout rel1.0 so it's not "new" in that sense, but yes, it's not been in the presubmit PR queue prior so the load hasn't been applied.

Thanks for clarifying the process.

tiswanso: /lgtm

tiswanso (approved):

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
costinm: Changes to test infrastructure and required tests should be broadly announced - since it affects all developers. And if developers raise issues - the TOC is the right forum to decide.

PR #949, opened 2018-08-15
Use clang 5.0 for istio/proxy release-1.0 branch.

PiotrSikora:

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.release-1.0.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.release-1.0.yaml
@@ -1,7 +1,7 @@
 
 bazel_spec: &bazel_spec
   containers:
-  - image: gcr.io/istio-testing/prowbazel:0.4.11
+  - image: gcr.io/istio-testing/prowbazel:0.4.10
     args:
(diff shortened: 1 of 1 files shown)


PiotrSikora:
cc @rkpagadala @lizan

PiotrSikora:
@lizan any thoughts on whether we should keep using clang-5.0 on `release-1.0` or upgrade to clang-6.0?

lizan:
I'm less concerned about clang versions to build, since we're already switching from gcc to clang for 1.0 release, right?

PiotrSikora:
@lizan nope, we switched to clang for release builds on `master` branch only.

lizan:
I meant switching from gcc to clang-5.0 is as risky as switching from gcc to clang-6.0, so either is fine if we are going to do either anyway.

PiotrSikora:
@lizan yeah, but we're not switching to either on the `release-1.0` branch ;)

kyessenov:
Piotr, make sure to start a PR on istio to pick up your changes. We cannot
quality whether the build changes are OK unless we run tests end to end.

kyessenov:
E.g. you are not done until you ship it to istio. I had too much "fun"
trying to absorb last changes into istio.

:shipit:

PR #954, opened 2018-08-27
White list collab branches

sebastienvas:


kyessenov (approved):
Would be better to have regex instead of exact names, but this works for now.

/lgtm
/approve

sebastienvas:
@kyessenov I misread the doc, it actually does not work the way I thought it did, it just whilte lists per branch per user. Example of use:
https://github.com/kubernetes/test-infra/blob/master/prow/plugins.yaml#L159

sebastienvas:
Closing as configuration options do not work as is for istio, and bot is too verbose.

Issue #957, opened 2018-08-29
Root cause why some presubmit jobs are not finishing


hklai:
See https://k8s-gubernator.appspot.com/pr/istio/istio/8347 and https://k8s-gubernator.appspot.com/pr/istio/istio/8292

There are a few presubmit jobs that were never finished, and newer runs were started. 

According to @BenTheElder, it could be due to rebasing/force push, or the test pod gets rescheduled. 
If the nodes are not stable then kubernetes will re-schedule the pod, and it will collect a new build ID. It had happened in k8s when a broken workload caused the nodes to run out of PIDs. 

"if these were not prow jobs, and just some random kubernetes pod, then if the node becomes unhealthy kubernetes restarts or reschedules the pod since prowjobs are usually just kubernetes pods managed by a prow controller, this can happen to them too. if this does happen to them, they will re-vend the build ID on boot IIRC. See https://github.com/kubernetes/test-infra/issues/5700."

"if it's happening something is wrong with the cluster possibly
we haven't had any real issues with this for some time now, and we were able to root cause those to a specific workload that was misbehaving badly and causing issues with the nodes
https://github.com/kubernetes/test-infra/issues/5877
(that was the root cause for most of the problems we saw)"

Since there was one commit in #8347 and there is no force push during today, and it may be useful to figure out if pod rescheduling is our problem and whether we can fix it.

Seb, could you check the prow log?

BenTheElder:
> Since there was one commit in #8347 and there is no force push during today

Not even by individual PR authors to their own PRs?

ldemailly:
nope, no change whatsoever to https://github.com/istio/istio/pull/8347 past initial submission yet it's done presubmit 5 times so far and taken > 3h

ldemailly:
also 3x the green one when only 1 PR got merged in master in the window is a separate but also serious issue/bug

BenTheElder:
This is likely a problem with the Kubernetes cluster then, jobs are only automatically aborted and rescheduled when:

- you push new code to the PR while a job is running, so Prow starts a new one instead and aborts the old one
- Kubernetes has to re-schedule the test pod (possibly evicted)

The latter is not specific to prow and not really debuggable without looking at the cluster's logs.

ldemailly:
@hklai want to file a separate issue for the triplicate green builds:
[...]

PR #959, opened 2018-08-30
Update Prow configuration and Fix issue with branch protection

sebastienvas:
I noticed that branch protection was failing which could also impact tide. The reason was that the branch speficifed in the job config was considered as a regex and therefore assuming that jobs from branch pr_4343_master should run since we have jobs defined for master.

The second issues is that we have jobs that are defined as always_run: false but are not marked as optional: true, which Tide consider as required, and therefore explain why some PR that can be merged do not show up in Tide UI

Should help with https://github.com/istio/test-infra/issues/957

### DIFF
--- a/prow/cluster/branchprotector_cronjob.yaml
+++ b/prow/cluster/branchprotector_cronjob.yaml
@@ -14,7 +14,7 @@ spec:
         spec:
           containers:
           - name: branchprotector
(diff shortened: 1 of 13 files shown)


rkpagadala (approved):
/lgtm
/approve

PR #961, opened 2018-08-31
Update Prow deployments and Jobs to ease Cluster pool maintenance

sebastienvas:
Instead of using the pre-made node label, we are defining the following labels:

- prow: prow: Used to run Prow deployments
- testing: build-pool: Used to run prow jobs that require more resources
- testing: test-pool: Used to run tests that don't need high CPU

This will ease maintenance as we can create a new node pool with the label and delete the one being use.


rkpagadala (approved):
/lgtm

PR #963, opened 2018-09-04
Force all deployments to be in the prow node pool

sebastienvas:
Move all deployments that are not test (Boskos, Metrics) to run a in a given node pool for security and to protect our production deplyements. This will also help in case of maintenance, where we can create a new pool and move only prod jobs to that limiting down time.

Follow up of #961


rkpagadala (approved):
/lgtm
/approve

hklai:
What is supposed to be the net effect of this change?

sebastienvas:
Just updated the PR description

PR #965, opened 2018-09-05
Use regex for branch spec

sebastienvas:
As https://github.com/kubernetes/test-infra/pull/9243 is likely not to be accepted.


hklai (approved):
/lgtm

PR #968, opened 2018-09-06
Fix docker install on container to support overlay2

sebastienvas:
Using overlay should improve performance. This will also allow to use a cache for docker in the container.
Updating bazel as well since it missed the last update.


hklai (approved):
/lgtm

PR #969, opened 2018-09-10
Experiment to replace bootstrap with pod utils

sebastienvas:
Might need multiple PRs to get this working.

### DIFF
--- a/prow/new-test.sh
+++ b/prow/new-test.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Copyright 2017 Istio Authors
+
+#   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


prow/new-test.sh
```diff
@@ -0,0 +1,36 @@
...
+# Print commands
+set -x
+
+echo "=== Building ==="
+bazel build //...
+
+echo "=== Testing ==="
+bazel test --test_output=all //...
+
+echo "=== Linting ==="
```
hklai, line 35: shall we lint before we test?

sebastienvas: it is just a test file, I ll remove it after. I think we should actually run those 2 jobs in parallel, like one running tests and the other running linters.

PR #970, opened 2018-09-10
Update test infra jobs to use pod-utils

sebastienvas:
One caveat is that the gubernator link is invalid. I will fix it when we move istio.


hklai (approved):
/lgtm

PR #974, opened 2018-09-18
enable mcp test variants in prow

ayj:
Enable a parallel set of e2e tests with the Mesh Control Protocol (MCP) enabled. MCP test variants are initially optional but will later become mandatory when MCP is the default. At that time, we can remove the MCP test variants and enable MCP itself in the existing e2e test cases.

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -61,6 +61,22 @@ presubmits:
     spec:
       <<: *test_spec
 
+  - name: test-e2e-mixer-no_auth-mcp
+    agent: kubernetes
+    branches: *branch_spec
+    always_run: true
+    optional: true
+    skip_report: true
+    context: prow/test-e2e-mixer-no_auth-mcp.sh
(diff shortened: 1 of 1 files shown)


prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -121,6 +163,19 @@ presubmits:
       max_concurrency: 5
       spec:
         <<: *test_spec
+    - name: e2e-bookInfoTests-envoyv2-v1alpha3-mcp
+      branches: *branch_spec
+      always_run: true
+      agent: kubernetes
+      context: prow/e2e-bookInfoTests-v1alpha3-mcp.sh
```
ayj: Does this name need to match the script in istio/istio, e.g. [prow/e2e-bookInfoTests-envoyv2-v1alpha3-mcp.sh](https://github.com/istio/istio/pull/8808/files#diff-215b0269746658a9594d8701f69fb24b)?

ayj:
/hold

I need to move a few of the presubmit tests to postsubmit per @hklai's request

ayj:
/hold cancel

cc @hklai - ready for review.

hklai (approved):
/lgtm

PR #979, opened 2018-09-25
Fix configuration issues

sebastienvas:
PATH was not set correctly in the image, and gcloud was not loading the service account credentials.


hklai (approved):
/lgtm

PR #980, opened 2018-09-26
Updating to pod-utils for istio master

sebastienvas:


sebastienvas:
cc @cjwagner, if you want to take a look

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -1,5 +1,12 @@
+job_template: &job_template
+  agent: kubernetes
```
cjwagner: This is the default now and can be omitted.

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -11,62 +18,34 @@ istio_container: &istio_container
...
   istio/istio:
 
   - name: istio-unit-tests
-    agent: kubernetes
+    <<: *job_template
     context: prow/istio-unit-tests.sh
-    branches: *branch_spec
     always_run: true
     rerun_command: "/test istio-unit-tests"
     trigger: "(?m)^/(retest|test (all|istio-unit-tests))?,?(\\s+|$)"
```
cjwagner: `rerun_command` and `trigger` now default reasonably based on the `name` so you can drop these fields.

Trigger regexps should never match `/retest`. That will force `/retest` to behave improperly and retrigger all tests, not just failed ones. (This is something we should probably enforce on config load, I just hadn't thought of anyone trying to do this.)

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -11,62 +18,34 @@ istio_container: &istio_container
...
     labels:
-      preset-service-account: "true"
       preset-istio-kubeconfig: "true"
     spec:
-      <<: *build_spec
+      containers:
+      - <<: *istio_container
+        command:
+        - entrypoint
```
cjwagner, line 34: You shouldn't need to call entrypoint directly. The pod utils will automatically have the entrypoint binary call the command you specify with `command` and `args` in the container spec.

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
sebastienvas: Good to know - Will remove them all.

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
sebastienvas: done

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
sebastienvas: I tried 
```yaml
command:
- prow/istio-unit-tests.sh
```
and 
```yaml
args:
- prow/istio-unit-tests.sh
```
and none of them work.

sebastienvas: meaning entrypoint is not being called

cjwagner: Discussed in DMs: This is not the pod utility entrypoint, this is Istio's entrypoint tool so this is actually correct. (Pod utils entrypoint is in the `/tools` directory).

PR #981, opened 2018-09-26
add githubctl to istio-builder

rkpagadala:
add githubctl to istio-builder
githubctl is used for finding green build
githubctl is used for starting tests in daily-releases

having it as part of the docker image makes the builds easier

### DIFF
--- a/scripts/tools/linux-install-githubctl
+++ b/scripts/tools/linux-install-githubctl
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+if [[ "$(uname)" != "Linux" ]]; then
+  echo "Run on Linux only."
+  exit 1
+fi
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+. ${DIR}/all-utilities || { echo "Cannot load Bash utilities" ; exit 1 ; }
+
+
+# Install githubctl
+function install_githubctl() {
+    pushd $DIR/..
+    git clone https://github.com/istio/test-infra.git -b master --depth 1
+      pushd test-infra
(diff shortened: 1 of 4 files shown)


scripts/tools/linux-install-githubctl
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+. ${DIR}/all-utilities || { echo "Cannot load Bash utilities" ; exit 1 ; }
+
+
+# Install githubctl
+function install_githubctl() {
+    pushd $DIR/..
+    git clone https://github.com/istio/test-infra.git -b master --depth 1
```
sebastienvas, line 15: you are already in test-infra, which means that you could build something different from what you have checked out. I guess that would be confusing

sebastienvas:
I am planning to remove the container eventually. Where is it used ?

hklai (changes requested):
Instead of doing this, let's go back to the old way in the green build PR.
i.e. git clone and bazel build it.

The next step is to change it to go get and/or figure out why if that does not work.  This only complicates things.

rkpagadala:
@sebastienvas the container is used for daily and monthly builds

scripts/tools/linux-install-githubctl
rkpagadala: no the docker image just has the scripts directory added. test-infra itself is not checked out

sebastienvas:
in gcb or in prow ?

rkpagadala:
in gcb

PR #984, opened 2018-09-27
One more try to get pod-utils to work for istio master

sebastienvas:
Changes from #980 
- Update docker image to use date-sha instead of version to avoid conflict
- Install python for istio-builder since gsutil needs it
- Install service account if GOOGLE_APPLICATION_CREDENTIALS is set since gsutil needs it
- chown -R was failing where the kubeconfig was mounted since the mount points is owned by root and the container is running as a User. Updated entrypoint to not fail, and also setting KUBECONFIG as an env (See https://github.com/istio/istio/pull/9009) to allow kube config to be mounted outside of container user directory


sebastienvas:
@hklai please review this, I ll try it tonight when there is less traffic
/hold

hklai (approved):
/lgtm

PR #985, opened 2018-09-27
Using a different host path for test infra and proxy

sebastienvas:
This may not work since the data in the cache is already owned by a different UID. So we'll need to clear the existing cache on every node.


hklai (approved):
/lgtm

PR #987, opened 2018-09-28
install fpm on the istio builder image

sebastienvas:
Image with missing fpm. 
We don't access to a cluster anymore to run tests. Removing that dependency.


hklai (approved):
/lgtm

PR #996, opened 2018-10-05
Update k8s version to 1.10.

hklai:
This updates the default jobs to k8s 1.10, and renames gke-e2e-test-1-10 to gke-e2e-test-latest to avoid constant name changes in future.

The jobs that use gke-e2e-test-1-10 will be broken but they run in postsubmit only. Will fix them later once this is deployed.

### DIFF
--- a/boskos/configs.yaml
+++ b/boskos/configs.yaml
@@ -1,4 +1,26 @@
 configs:
+- name: gke-e2e-test-1-11
+  needs:
+    gcp-project: 1
+  config:
+    type: GCPResourceConfig
+    content: |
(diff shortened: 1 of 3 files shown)


sebastienvas:
You also need to update https://github.com/istio/test-infra/blob/master/boskos/resources.yaml

sebastienvas:
and https://github.com/istio/test-infra/blob/master/boskos/metrics-deployment.yaml#L21 this one needs a redeployment.

hklai:
@sebastienvas Thanks. PTAL

PR #999, opened 2018-10-08
add test for istio build

rkpagadala:
the build test for istio has been added with
https://github.com/istio/istio/pull/9201


hklai (approved):
/lgtm

PR #1016, opened 2018-10-30
run proxy postsubmit periodically

rkpagadala:
Proxy postsubmit failed twice in the last week, this is a problem because we depend on it to create artifacts used in istio/istio. Running it periodically overcomes this problem


hklai (commented):
should we run all postsubmit periodically?

rkpagadala:
@hklai
 istio postsubmit is a long list and prone to change, we need a better solution for it.
I added istio/api postsubmit as well

rkpagadala:
@hklai PTAL added postsubmits for istio, api and proxy for master branch

hklai:
are these results used to determine green build in release?

rkpagadala:
Not yet. The main motivation has been to run post submit for proxy which generates artifacts for istio build.

hklai:
I think we should eventually move proxy build to istio automation instead.

rkpagadala:
Closing this new PR at https://github.com/istio/test-infra/pull/1030

Issue #1018, opened 2018-10-31
e2e GKE clusters support for network-plugin=cni


tiswanso:
The prow/boskos e2e GKE clusters need to have the network-policy feature enabled to have the kubernetes installation setup kubelets with the network-plugin=cni option.  This is required to integrate prow e2e tests for the istio/cni feature.

https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#enabling_network_policy_enforcement

2 step process to update existing clusters:
  ```
   $ gcloud container clusters update $cluster --update-addons=NetworkPolicy=ENABLED
   $ gcloud container clusters update $cluster --enable-network-policy
  ```
**NOTE:**  this restarts the VMs in the cluster and takes about 10min to complete from what I've seen.

hklai:
Per @sebastienvas, this may require a change in https://github.com/istio/test-infra/blob/master/boskos/gcp/gke.go

tiswanso:
@hklai -- thanks, @john-a-joyce is taking a look.  His initial question is how to test changes to the boskos/mason code.  I haven't looked at the details in a while but is it possible to run mason stand-alone and point it at our own (private) gcp proj?

howardjohn:
I believe this is fixed since we run the cni tests, but we moved to kind anyways so it isn't really relevant

PR #1020, opened 2018-11-01
Adds a CNI based e2e prow test

john-a-joyce:
This adds a manually triggered e2e_simple
test that installs the optional CNI plugin.
This needs to be combined with istio
PR #9577 and test-infra issue #1018 to be
effective.


hklai (approved):
/lgtm

PR #1021, opened 2018-11-01
change timeout for proxy postsubmit for master to 2 hours

rkpagadala:
Proxy post submit for master needs more time, change timeout to 2 hours for master branch
As per request from @qiwzhang

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
@@ -1,4 +1,24 @@
 
+bazel_postsubmit_spec: &bazel_postsubmit_spec
+  containers:
+  - image: gcr.io/istio-testing/prowbazel:0.4.13
+    args:
+    - "--repo=github.com/$(REPO_OWNER)/$(REPO_NAME)=$(PULL_REFS)"
+    - "--clean"
+    - "--timeout=120"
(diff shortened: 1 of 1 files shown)


prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
```diff
@@ -64,4 +84,4 @@ postsubmits:
     labels:
       preset-service-account: "true"
     spec:
-      <<: *bazel_spec
+      <<: *bazel_postsubmit_spec
```
hklai, line 87: I think you can reuse the same stanza and override only the timeout.

e.g. https://github.com/istio/test-infra/blob/master/prow/config.yaml#L75

rkpagadala: That one does not override anything Andy, it just adds master and release-1.1 in addition to preexisting config

hklai:
master used to be in the stanza above and it got overridden. I think it was
a problem I fixed a while back.

Have you tried it here?

prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
sebastienvas: you can only overwrite with a map, not a list. Krishna is right.

PR #1025, opened 2018-11-06
Enables GKE network policy

john-a-joyce:
This change enables the NetworkPolicy
on GKE clusters and sets the provider
to Calico.  This change is needed to
test with the CNI plugin.


john-a-joyce:
Need some experts to evaluate this PR as I really have no way to properly test this change prior to merging.  I did confirm that the unit tests ran properly in my local environment and modified them to include the new fields.

john-a-joyce:
Resolves  https://github.com/istio/test-infra/pull/1025

PR #1028, opened 2018-11-06
Update sticky lgtm config.

hklai:


hzxuzhonghu:
@hklai What's this for? Now we have a risk, if the author update code with potential bug, and without reviewer review again, as the `lgtm` label still there,  the bot merged it.

linsun:
@hzxuzhonghu the bot won't merge till all required tests have passed.  This is an effort to save folks time to keep enter lgtm

hklai:
@hzxuzhonghu @linsun Correct. And this is restricted to istio hackers members only to minimize the risk,

PR #1030, opened 2018-11-07
run proxy postsubmit, remove green build job, sds api

rkpagadala:
run proxy postsubmit 
remove green build job
remove proxy sds api branch config


hklai (approved):
/lgtm

PR #1031, opened 2018-11-08
Update toolbox to support codecov diff.

hklai:
1. Adding a way to get base SHA from an existing PR in githubctl.

2. Updating pkg_check to generate diff from two coverage files.

### DIFF
--- a/toolbox/githubctl/main.go
+++ b/toolbox/githubctl/main.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	owner                 = flag.String("owner", "istio", "Github owner or org")
-	tokenFile             = flag.String("token_file", "", "File containing Github API Access Token")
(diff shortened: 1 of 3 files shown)


toolbox/githubctl/main.go
```diff
@@ -291,8 +301,10 @@ func readPostsubmitListFromProwConfig(org, repo, branch string) map[string]struc
 
 func init() {
 	flag.Parse()
+	var token string
```
rkpagadala, line 304: Why these changes? Using := to declare is also idiomatic go

rkpagadala (approved):
/lgtm

toolbox/githubctl/main.go
hklai: oh it was due to some previous change when token and error could have been assigned different. I had since removed that code but did not convert this back.

PR #1034, opened 2018-11-09
Update boskos deployment

sebastienvas:
cc @john-a-joyce 

I already updated the deployements. I reset all the cluster and they are now created with a network policy.


hklai (approved):
/lgtm

PR #1038, opened 2018-11-12
rename build-tests to release-test and make it required

rkpagadala:
make the release test required so that daily builds are not broken

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -230,7 +230,7 @@ presubmits:
         - <<: *istio_container
           command:
           - entrypoint
-          - prow/build-tests.sh
+          - prow/release-test.sh
         nodeSelector:
           testing: test-pool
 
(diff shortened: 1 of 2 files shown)


rkpagadala:
/hold
until https://github.com/istio/istio/pull/9893 is merged

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -218,11 +218,10 @@ presubmits:
           - entrypoint
           - prow/e2e-simpleTests-cni.sh
         nodeSelector:
-    - name: build-tests
+    - name: release-test
```
hklai, line 221: move this to postsubmit for now.

failure like the one we had recently is just noise to be in presubmit. also you should set up testgrid to see historical results.

rkpagadala: Our daily builds failed for both 1.1 and master, having this in presubmit avoids that pain.
Will setup testgrid for this job.

hklai (approved):

hklai (approved):
/lgtm

PR #1054, opened 2018-11-21
Adding threshold override to pkg_check

hklai:


hklai:
@rkpagadala PTAL when you get a chance. Merging this to unblock my work for now.

PR #1069, opened 2018-11-28
fix githubctl to use specific paths for monthly/daily

rkpagadala:


toolbox/githubctl/main.go
```diff
@@ -220,6 +221,7 @@ func ReleasePipelineQualification(baseBranch *string) error {
 	u.AssertNotEmpty("hub", hub) // TODO (chx) default value of hub
```
hklai, line 221: there are way too much release specific information in this util.

perhaps this should be updated to create a new PR with some basic input?

PR #1074, opened 2018-11-30
Using cert-manager to renew Prow certs

sebastienvas:


prow/cluster/cert-manager/tls-ing.yaml
```diff
@@ -1,16 +1,22 @@
...
+    certmanager.k8s.io/acme-http01-edit-in-place: "true"
     kubernetes.io/ingress.class: "gce"
 spec:
   tls:
   - secretName: prow-tls
+    hosts:
+    - prow.istio.io
   rules:
-  - http:
+  - host: prow.istio.io
```
hklai, line 18: do we need others like http://velodrome.istio.io?

hklai (approved):
/lgtm

prow/cluster/cert-manager/cluster-issuer.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  namespace: default
+  # Adjust the name here accordingly
+  name: letsencrypt-prod
+spec:
+  acme:
+    # The ACME server URL
+    server: https://acme-v02.api.letsencrypt.org/directory
+    # Email address used for ACME registration
+    email: sebastienvas@google.com
```
rkpagadala, line 14: change this istio-engprod@

prow/cluster/cert-manager/tls-ing.yaml
sebastienvas: not on the prow cluster. This ingress redirecting Prow services. Velodrome is hosted by k8s, so we have to use a proxy.

prow/cluster/cert-manager/cluster-issuer.yaml
sebastienvas: istio-engprod does not exist, and google.com google groups cannot be reached from external. @hklai do we have a better group email for this ?

hklai: istio-oncall@googlegroups.com?

PR #1085, opened 2018-12-06
Adding a bot to /retest approved but failed PRs

hklai:


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -66,6 +66,40 @@ periodics:
...
+      - --template
+      - --ceiling=2
+      - --confirm
+      volumeMounts:
+      - name: token
+        mountPath: /etc/token
+    volumes:
+    - name: oauth
+      secret:
+        secretName: oauth-token
```
rkpagadala, line 101: Does the bot not need a new secret?

hklai: trying to reuse the existing one. not sure why we need it if it works.

(the same issue Cole mentioned exist in other jobs anyway, if that is really an issue)

PR #1086, opened 2018-12-06
Fix retester job typo

hklai:


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -66,7 +66,7 @@ periodics:
     nodeSelector:
       testing: test-pool
 
-- name: periodic-test-infra-retester
+- name: periodic-retester
```
cjwagner, line 69: I would call this `periodic-istio-retester` in case you want to add more of these jobs for other repos. You could add more `repo:foo` tags to the query to get the same functionality, but it is probably better to use different jobs for different repos so that the `ceiling` argument can be used to retest a few PRs at a time in each repo instead of possibly retesting PRs from only one of the repos.

cjwagner (commented):
The `- --token=/etc/token/bot-github-token` also appears to be wrong? I just checked the secret to see if it was added and there is no `bot-github-token` key in the secret only an `oauth` key.

PR #1088, opened 2018-12-06
add istio-releases/pipeline repo for prow

rkpagadala:
add istio-releases/pipeline repo for prow


hklai:
/lgtm

@rkpagadala to figure out if we can maintain only one branch in istio-releases/pipeline.

PR #1102, opened 2018-12-12
Adding config for hklai-test for testing

hklai:


hklai:
adding this file for testing for now.

PR #1148, opened 2019-01-16
set sds test optional

quanjielin:


prow/cluster/jobs/istio/istio/istio.istio.release-1.1.yaml
```diff
@@ -387,6 +388,7 @@ postsubmits:
           testing: test-pool
     - name: istio_auth_sds_e2e
       <<: *job_template
+      optional: true
```
hklai: this is unnecessary for postsubmit

quanjielin: updated.

hklai (approved):
/lgtm

PR #1156, opened 2019-01-22
Update go to 1.11, remove use of deps in favor of go mod.

sebastienvas:
- Update goland to 1.11.6
- Update gazelle, go rules, buildtools, docker
- Update vendors with go mode


hklai:
Is this really needed?

sebastienvas:
Yes, we are using really old version of bazel, deps is deprecated. We were trying to update vendors to be able to upgrade prow, and we failed miserably which is why I started all that.

sebastienvas:
The build failure should go away once we use the image with the newer bazel version.

PR #1143, opened 2019-01-11
add sds e2e test to test-infra

quanjielin:
https://github.com/istio/istio/issues/9035

PR #1158, opened 2019-01-23
enable sds e2e test in master branch

quanjielin:
similar change was made to 1.1 branch in https://github.com/istio/test-infra/pull/1143, add in master presubmit with master reopen


utka:
@sebastienvas. Can it be this change #1156?

hklai:
I suspect so too. Prob need to clean up those lint errors as a follow up.

sebastienvas:
no this passed before. This related to the fact we use the latest v2 version of gometalinter. So bascially the linting guideline changed after a new version of gometalinter was pushed. We need to update the code.

PR #1159, opened 2019-01-24
Move https://github.com/istio/admin-sites/gcsweb.istio.io to https://github.com/istio/test-infra/gcsweb.istio.io

utka:
Moved istio/admin-sites/gcsweb.istio.io under istio/test-infra for better understanding who is owner of the  gcsweb.


sebastienvas:
Please see https://github.com/istio/test-infra/pull/1160.

utka:
Will create another PR with more thing like in #1160.

PR #1163, opened 2019-01-27
Make upgrade tests release blocking

hklai:


utka:
Any reason why we are moving it to blocking? Something changed? And why it was marked optional before?

hklai:
@utka These tests were recently fixed, so I make them required again.

PR #1166, opened 2019-02-07
Adding a new prow file for collab-galley branch

ozevren:


ozevren:
@sebastienvas any chance I can bother you for a review?

sebastienvas:
Could you instead copy from master and remove anything that does not apply.
In master we should be using the "decorate: true" which is the way to go
forward.

sebastienvas (changes requested):
Please copy from the master branch and remove tests that don't make sense, as this one is already updated to use podutils (the decorate: true)

ozevren:
Thanks. Will do soon.

ozevren:
ok, updated.

ozevren:
Not sure what else is needed for this to go in.

PR #1167, opened 2019-02-08
Update Bazel to 0.22 and LLVM/clang to 7.

PiotrSikora:

### DIFF
--- a/docker/istio/shared/tools/install-bazel.sh
+++ b/docker/istio/shared/tools/install-bazel.sh
@@ -2,7 +2,7 @@
 
 set -eux
 
-BAZEL_VERSION='0.21.0'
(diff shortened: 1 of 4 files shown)


sebastienvas:
@utka once this is merged, you would need to [build the new image](https://github.com/istio/test-infra/blob/master/docker/prowbazel/Makefile#L2), and update the proxy jobs in https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml#L4.

The proxy jobs haven't been updated to use podutils, so we should do so, and while we do that, we should create an image for proxy, in this dir https://github.com/istio/test-infra/tree/master/docker/istio

PR #1168, opened 2019-02-08
Enable prow tests for auth-v2 branch

pitlv2109:
We branched off a new branch from **release-1.1** called [authz-v2](https://github.com/istio/istio/tree/authz-v2). We'd like to have same prow tests as **release-1.1**'s triggered when a PR is created for **authz-v2**.

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.authz-v2.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.authz-v2.yaml
@@ -1,0 +1,400 @@
+job_template: &job_template
+  branches:
+  - "^authz-v2$"
+  decorate: true
+  path_alias: istio.io/istio
+
+istio_container: &istio_container
+  image: gcr.io/istio-testing/istio-builder:v20181008-db31a9fd
+  # Docker in Docker
+  securityContext:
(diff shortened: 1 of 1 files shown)


pitlv2109:
@yangminzhu @liminw

utka:
You might need to add more checks to prow/config.yaml later 
/lgtm

yangminzhu:
@utka Is this PR enough to turn on the prow tests on branch `authz-v2`? Would you mind to clarify a little what else changes are needed in `prow/config.yaml`? Thanks.

pitlv2109:
@yangminzhu 
If I understand correctly, it seems like we need to make some tests to be required as in master or release-1.1 (https://github.com/istio/test-infra/blob/7ade5cc4de5471300b673100dc122b5e3882803c/prow/config.yaml#L108). I created a PR for that. Still waiting for some tests to run before cc-ing you.

utka:
@yangminzhu, I was looking at code to figure out if you need to add something else and one thing which i found is prow/config.yaml. I think that should be it.

yangminzhu:
@utka Thanks Maria, do you know when will these change take effect?

PR #1176, opened 2019-02-12
Require certain tests to pass for authz-v2 branch

pitlv2109:
Require same required tests as **release-1.1**'s to pass in order to merge PRs to **authz-v2**.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -105,6 +105,17 @@ branch-protection:
                 - "ci/circleci: e2e-mixer-noauth-v1alpha3-v2"
                 - "ci/circleci: e2e-pilot-cloudfoundry-v1alpha3-v2"
                 - "merges-blocked-needs-admin"
+            authz-v2:
+              protect: true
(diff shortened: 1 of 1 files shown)


pitlv2109:
@yangminzhu This is the PR mentioned in https://github.com/istio/test-infra/pull/1168.
/cc @utka Let us know if we need to do anything else.

prow/config.yaml
```diff
@@ -105,6 +105,17 @@ branch-protection:
                 - "ci/circleci: e2e-mixer-noauth-v1alpha3-v2"
                 - "ci/circleci: e2e-pilot-cloudfoundry-v1alpha3-v2"
                 - "merges-blocked-needs-admin"
+            authz-v2:
+              protect: true
```
yangminzhu: We don't need to make it protect, I think only admin is allowed to merge if it's protected.

pitlv2109: Deleted. Thanks Yangmin!

PR #1179, opened 2019-02-20
Adding more resources to boscos clusters for gke-e2e-tests and perf clusters

utka:


howardjohn:
Not sure if this was the motivation behind this, or other issues, but I think this will fix the upgrade tests that are failing. As far as I can tell, https://github.com/istio/istio/pull/11476 increased the resource requests. other E2E tests use the values-e2e.yaml which uses less resources, but the upgrade test does not, so we will need a bigger cluster to handle this.

utka:
@howardjohn You are correct to fix update tests and some additional work for perf tests

PR #1185, opened 2019-03-06
Adding scope to GKE clusters for boskos

utka:


boskos/gcp/test-configs.yaml
```diff
@@ -20,6 +20,7 @@ configs:
             - https-server
             scopes:
             - https://www.googleapis.com/auth/cloud-platform
+            - https://www.googleapis.com/auth/trace.append
```
sebastienvas, line 26: dont you need to add under cluster ?

boskos/gcp/test-configs.yaml
```diff
@@ -44,3 +45,4 @@ configs:
...
```
sebastienvas, line 55: same comment here

boskos/gcp/config_test.go
```diff
@@ -51,6 +51,7 @@ func TestParseInvalidConfig(t *testing.T) {
 					},
 					Scopes: []string{
 						"https://www.googleapis.com/auth/cloud-platform",
+						"https://www.googleapis.com/auth/trace.append",
```
sebastienvas, line 58: Don't you need to add this under cluster ?

boskos/gcp/config_test.go
```diff
@@ -78,6 +79,7 @@ func TestParseInvalidConfig(t *testing.T) {
...
```
sebastienvas, line 90: Same comment here.

sebastienvas (changes requested):
the code change seems fine, but you added the scopes in the vms. You should add them in the Cluster, and verify that in test.

boskos/gcp/config_test.go
utka: You are right! Good catch

boskos/gcp/config_test.go
utka: Done

boskos/gcp/test-configs.yaml
utka: Done

boskos/gcp/test-configs.yaml
utka: Done

utka:
@sebastienvas ptal

sebastienvas:
You need to release another version of Mason for this to take effect.

sebastienvas:
https://github.com/istio/test-infra/blob/master/boskos/Makefile#L8 and then update https://github.com/istio/test-infra/blob/master/boskos/mason-deployment.yaml#L18, and make mason-depoyment

PR #1188, opened 2019-03-19
Change order of the tests, run after istio_presubmit test is done

utka:


ozevren:
@sebastienvas Can we get this in? It is causing quite a few failures in Istio postsubmit tests.

sebastienvas:
/lgtm
@utka please talk with @fejta about pipeline CRDs. Run after success jobs are not supported anymore.

PR #1197, opened 2019-04-03
Adding the ability to update SHA256 info(for proxy repo) and update SHAs for all istio's repos with "istio.deps" file.

icygalz:
Once this changes checked-in, the script should automatically auto-update SHAs for repos with istio.deps file.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -4,6 +4,7 @@ reviewers:
   - chxchx
   - rkpagadala
   - utka
+  - icygalz
 approvers:
   - sebastienvas
   - yutongz
@@ -12,3 +13,4 @@ approvers:
   - rkpagadala
   - fejta
   - utka
+  - icygalz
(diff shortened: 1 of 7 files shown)


utka:
In reality is it going to be only istio/istio repo?

icygalz:
no, this will also update istio/proxy repo. I have run it locally and it created 2 PRs(1 for istio/proxy and 1 for istio/istio).
This is the created PR example:
https://github.com/icygalz/proxy/pull/10
https://github.com/icygalz/istio/pull/3

utka:
It looks good for updating proxy
for istio/istio, it needs to update Gopkg.lock and vendor files in the same PR otherwise proxy and istio  will not use the same API version.
Example: https://github.com/istio/istio/pull/12199/files
It is updated by running dep update command.

icygalz:
This is the sample PR generated by the scripts:
https://github.com/icygalz/istio/pull/5#event-2257149322
https://github.com/icygalz/proxy/pull/12#event-2257149254

Currently this workflow is being set to run every hour. Do we want to keep updating it every hour? @utka

Issue #1199, opened 2019-04-04
go get mason_client is timing out 


utka:
As example 
https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio/13054/e2e-simpleTests-cni/2508/build-log.txt
We need to add the RUN go get istio.io/test-infra/boskos/cmd/mason_client to the dockerfile that creates the image or use go mod vendor all your dependencies and then bazel build //vendor/istio.io/test-infra/boskos/cmd/mason_client
Example: https://github.com/kubernetes/test-infra/blob/master/hack/update-deps.sh

howardjohn:
Should be fixed by goproxy

PR #1200, opened 2019-04-04
Add go get mason_client to Docker image

utka:
There will be second PR to remove it from istio/prow/mason_lib.sh

### DIFF
--- a/docker/prowbazel/Dockerfile
+++ b/docker/prowbazel/Dockerfile
@@ -45,6 +45,7 @@ RUN sudo chmod +rx /usr/local/bin/entrypoint
 RUN mkdir -p /home/bootstrap/go/src
 ENV GOPATH=/home/bootstrap/go
 ENV PATH ${GOPATH}/bin:${PATH}
+RUN go get istio.io/test-infra/boskos/cmd/mason_client
 
 # Set CI variable which can be checked by test scripts to verify
 # if running in the continuous integration environment.
--- a/boskos/cmd/mason/Dockerfile
+++ b/boskos/cmd/mason/Dockerfile
@@ -1,17 +1,9 @@
 FROM google/cloud-sdk:alpine
 
-RUN apk add --no-cache git make musl-dev go
-ENV GOROOT /usr/lib/go
-ENV GOPATH /go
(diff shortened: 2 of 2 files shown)


sebastienvas:
the dockerfile is the server. The client needs to use the mason client. That won't work.

utka:
Ok, is this one correct?

sebastienvas:
Is prowbazel still used ? (Apparently yes for)
The one for istio is istio-builder I think, so you can update https://github.com/istio/test-infra/blob/master/docker/istio/shared/tools/install-golang.sh#L22
 and https://github.com/istio/test-infra/blob/master/scripts/tools/linux-install-golang#L21 for prowbazel but you ll need to rebuild all the images and update all configs.

sebastienvas:
also what is the reason for doing this ? The reason of calling go get was to always get the latest version as this version should match the deployment. Adding it to the image should be fine, just note that when you'll update the mason client, you also have to update the job configs to roll it out. It is probably safer that way, but also more work for you.

utka:
This is the problem which happens sometimes during tests: https://github.com/istio/test-infra/issues/1199

docker/prowbazel/Dockerfile
```diff
@@ -45,6 +45,7 @@ RUN sudo chmod +rx /usr/local/bin/entrypoint
 RUN mkdir -p /home/bootstrap/go/src
 ENV GOPATH=/home/bootstrap/go
 ENV PATH ${GOPATH}/bin:${PATH}
+RUN go get istio.io/test-infra/boskos/cmd/mason_client
```
fejta, line 48: Each RUN and ENV command creates a new docker layer. Ideally we should try to combine as many of these as possible so we can minimize layers.

fejta (approved):
/lgtm
/hold

`/hold cancel` if you're ready for this to merge

geeknoid:
@sebastienvas @fejta @cjwagner 

Given that Maria's left the project, what should we do with this PR? Do you guys want to take this over the finish line?

geeknoid:
Closing. Please reopen if needed.

PR #1208, opened 2019-04-08
Adding perf testing for daily releases

utka:
get tools was added in https://github.com/istio-releases/pipeline/pull/501
setup perf cluster was added in https://github.com/istio/istio/pull/13159
After this is tested, I will add servicegraph test to run for 10min


mandarjog (approved):
/lgtm

icygalz:
I will lgtm after "setup perf cluster was added in istio/istio#13159" is merged.

PR #1211, opened 2019-04-10
Increase resources for Proxy prow build

utka:


utka:
Updated resources per feedback.

icygalz:
/lgtm

i have added new nodes with bigger resources. we can increase it more if we need it.

sebastienvas (approved):

PR #1215, opened 2019-04-15
Add integration tests that detect race condition to test-infra

lei-tang:
This PR is for #13341 and depends on the PR https://github.com/istio/istio/pull/13342, which has been merged.


utka:
Don't you mind adding it as postsubmit first to check that it is not flaky?

lei-tang:
Done: added it as postsubmit first.

lei-tang:
https://github.com/istio/istio/issues/13341

PR #1216, opened 2019-04-16
Fix deployment manifest links in readme

dpb587-pivotal:
Seems they were renamed at some point a while back.


dpb587-pivotal:
Thought I was already covered by CLA; checking on that. It's an "obvious fix", if you want to merge in the meantime.

dpb587-pivotal:
@googlebot I signed the CLA via corporate @pivotal.

PR #1221, opened 2019-04-18
update istio/proxy to use prowbazel 0.5.2 with gcc-7/g++-7

yangminzhu:
For istio/proxy#2179

/assign @icygalz
I only updated the prowbazel for istio/proxy in `master` and `wasm` branch, not including 1.1 and other periodic jobs. I assume only the istio/proxy in `master` and `wasm` branch will need to compile with the latest istio/proxy code, right?


icygalz:
/lgtm
/approve

thank you again for splitting your PR.

PR #1227, opened 2019-04-24
Removing Integration tests from Release-1.1

ozevren:
The Integration tests in Release-1.1 are not stable. It is not providing any good signals for quality, and causing spurious errors/flake detection in testgrid.


nmittler (approved):
/lgtm
/approve

PR #1229, opened 2019-04-29
Apply plugin config per-repo rather than on all repos.

geeknoid:


utka (changes requested):
You need to add it to istio/test-infra as well

geeknoid:
Oh, OK. Done.

Thanks.

PR #1230, opened 2019-04-29
Revert "Apply plugin config per-repo rather than on all repos."

utka:
Reverts istio/test-infra#1229
Looks like it removed ability for istio-testing to add labels automatically.

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -56,33 +56,7 @@ approve:
   lgtm_acts_as_approve: true
 
 plugins:
-  istio/istio:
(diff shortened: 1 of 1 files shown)


howardjohn:
Do we know this fixes it or a hoping for the best?

howardjohn (approved):

utka:
There are three different PRs in last few days which touched prow configs. I am just going to roll them back one by one until we are back to normal state

utka:
FYI: I will manually add labels, to fix automatically adding labels :)

PR #1231, opened 2019-04-29
Reapply the change to stop prow from messing with istio.io.

geeknoid:


howardjohn:
@utka I think tide didn't pick up the commands

howardjohn (approved):
/lgtm

This will turn it off for istio/installer too (right?), which we want

PR #1233, opened 2019-04-30
Fix cert-manager configuration

sebastienvas:


cjwagner:
LGTM. These changes are already deployed so we can either manually merge now or wait until the new cert is propagated and Prow is able to receive webhooks again.

PR #1235, opened 2019-05-02
Add stackdriver e2e to master daily pipeline

bianpengyuan:
Add stackdriver e2e test to master release test pipeline

/cc @kyessenov

### DIFF
--- a/prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.master.yaml
+++ b/prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.master.yaml
@@ -208,6 +208,24 @@ presubmits:
         nodeSelector:
           testing: test-pool
 
+    - name: release-e2e-stackdriver
+      <<: *job_template
+      always_run: true
+      optional: true
+      context: prow/release-e2e-stackdriver
+      max_concurrency: 5
+      labels:
+        preset-service-account: "true"
+      spec:
+        containers:
+        - <<: *istio_container
+          command:
+          - entrypoint
+          - scripts/run-test.sh
+          - prow/e2e-stackdriver.sh
+        nodeSelector:
+          testing: test-pool
+
 postsubmits:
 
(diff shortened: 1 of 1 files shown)


kyessenov (approved):
/lgtm

prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.master.yaml
```diff
@@ -208,6 +208,24 @@ presubmits:
         nodeSelector:
           testing: test-pool
 
+    - name: release-e2e-stackdriver
```
utka, line 211: Can you add it as postsubmit first?

utka (changes requested):

prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.master.yaml
bianpengyuan: We eventually want to block daily release if this test failed. I marked this test as optional so that it won't block now. Is there some reason that we should put it in postsubmit?

utka: When you add it as presubmit, even as optional, it is going to run for every PR immediately. I would like to see that the test is not flaky on postsubmit, before adding it to presubmit.

bianpengyuan: I thought this is the release pipeline and it should only run with a new daily release?

bianpengyuan: Any update?

utka: Oh :) You are right.

utka (approved):

PR #1243, opened 2019-05-08
Separate master Prow tests from other branches

utka:


ozevren:
Looks ok to me but it looks like you still have a conflict. :)

Also, perhaps it is better to update testgrid first to add the
master-counterparts, so that we don't get disruption in our flow?

*From: *istio-bot <notifications@github.com>
*Date: *Wed, May 8, 2019 at 2:36 PM
*To: *istio/test-infra
*Cc: *Ozben Evren, Review requested

@utka <https://github.com/utka>: The following tests *failed*, say /retest

PR #1244, opened 2019-05-09
Save circleci logs per branch

utka:


sebastienvas:
I am not to familiar with this, but gubernator is being replaced with spyglass.

PR #1246, opened 2019-05-10
Making local & select K8s integration tests required.

ozevren:
The original aggregate gate was required for local tests, and the TOC approved making the (curated) K8s tests required last week.

Now that the gates are up and running and in green mode, this PR makes the all the local tests, and some of the K8s tests (which were in the curated set) required.

https://testgrid.k8s.io/istio-presubmits-master


nmittler (approved):
/lgtm
/approve

PR #1255, opened 2019-05-18
Add KinD to istio builder

howardjohn:
From KinD:
>Building kind now requires go modules (using any upstream supported go version)
or using the new Makefile. Pre-built binaries for all platforms are included with
this release. Please consider using one of the binaries in your CI setups, or
at least pinning to a tagged release.

So using the binary is preferred, and easier than dealing with go modules stuff...

I built the image locally and tried running kind, but for some reason docker says
`* Starting Docker: docker `
but doesn't actual start. Not sure why or how this image is intended used... Presumably if docker works now it will work with this change though

### DIFF
--- a/scripts/tools/linux-install-golang
+++ b/scripts/tools/linux-install-golang
@@ -14,6 +14,8 @@ GO_ARCHIVE="go${GO_VERSION}.linux-amd64.tar.gz"
 GO_URL="${GO_BASE_URL}/${GO_ARCHIVE}"
 GO_DIRECTORY="${TOOLS_DIR}/go"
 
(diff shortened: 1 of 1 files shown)


icygalz:
/lgtm

Thanks!!!

icygalz:
To answer the question:
THe image is used here:
https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio/istio/istio.istio.master.yaml#L8

For the changes to be applied, we need to create and push the image to gcr.io and then change the value on the yaml file above to the newly created one.

howardjohn:
I was confued because I `docker run`ed the image, but then inside of it docker doesn't seem to work, so wasn't sure how it works when we actually use it. Still not sure really, but if it works then its all good. Let me know if you need any other help with kind stuff, me and costin have played around with it a lot

PR #1259, opened 2019-05-20
Adding prow to istio-ecosystem

icygalz:
The hook is returning green now! PR ready for review.

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -11,6 +11,8 @@ triggers:
   - istio/proxy
   - istio-releases/pipeline
   trusted_org: istio
+- repos:
+  - istio-ecosystem/authservice
 
 config_updater:
   maps:
@@ -44,6 +46,8 @@ lgtm:
   - istio/proxy
   - istio-releases/pipeline
   trusted_team_for_sticky_lgtm: "Istio Hackers"
+- repos:
+  - istio-ecosystem/authservice
 
 approve:
 - repos:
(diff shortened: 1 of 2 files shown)


prow/plugins.yaml
```diff
@@ -10,7 +10,9 @@ triggers:
   - istio/test-infra
   - istio/proxy
   - istio-releases/pipeline
+  - istio-ecosystem/authservice
   trusted_org: istio
+  trusted_org: istio-ecosystem
```
sebastienvas: I don t think that s valid YAML:
add another section:
```YAML
- repos
  - istio-ecosystem/authservice
  trusted_org: istio-ecosystem
```

prow/plugins.yaml
```diff
@@ -43,6 +45,7 @@ lgtm:
   - istio/test-infra
   - istio/proxy
   - istio-releases/pipeline
+  - istio-ecosystem/authservice
   trusted_team_for_sticky_lgtm: "Istio Hackers"
```
sebastienvas, line 51: Same comment here you might want to create another section

prow/plugins.yaml
icygalz: fixed

prow/plugins.yaml
icygalz: fixed

icygalz:
Webhook is green now!
/hold cancel

PR #1260, opened 2019-05-21
Create a new istio-builder image with KinD package included

icygalz:
I made sure my git clone is clean with only KinD packages addition before creating and pushing new istio-builder images. I enabled the new image only for e2e-kind-simpleTest postsubmit test.
I will monitor the test and once e2e-kind-simpleTest is passing, i will enable the new image for master's postsubmit tests and monitor it before enabling it for presubmit tests.


icygalz:
@howardjohn  FYI.

PR #1264, opened 2019-05-22
Getting rid run_after_success for postsubmit

icygalz:
Getting rid of run_after_success for postsubmit first. I will monitor it and once they are passing, i will get rid run_after_success for presubmit.
istio-postsubmit.sh contains istio-unit-tests.sh and build&push docker image. Now we don't need stand alone build&push images, i just put in the istio-unit-tests.sh in.


utka:
Exciting!
/lgtm

PR #1271, opened 2019-05-23
Update ninja to 1.9.0 in proxy builder

bianpengyuan:
This is required to build v8 in proxy.

cc @mandarjog @kyessenov


bianpengyuan:
@utka Is it possible to run this builder image locally? It is not efficient to fix just one build error every time based on output from proxy repo. It seems to need lots of set up to run the image properly in local machine though.

bianpengyuan:
@utka @icygalz May I get some review on this change? thanks!

PR #1274, opened 2019-05-24
Turning off upgrade test for release-1.2 pipeline branch

icygalz:
Turning off upgrade test for release-1.2 pipeline branch


howardjohn:
/lgtm

We will replace with 1.1 -> 1.2 upgrade test, but for now lets just get an image out the door so we can start testing

PR #1280, opened 2019-05-29
Change all master and release-1.2 branch workflow to use go1.12.5

icygalz:


ericvn (commented):
Do you want to add a comment to `prowbazel/README.md`?
   * 0.5.5: Update golang to 1.12.5

It seems the version has changed to a date for Istio-builder so maybe an update to `docker/istio_builders/CHANGELOG.md` isn't needed.

ericvn:
Also, not sure how WORKSPACE is used, but it has a Go version as well. I had updated it as such in my un-submitted PR.

`go_register_toolchains(go_version = "1.12.5")`

ericvn (approved):
lgtm

ericvn:
Looks like we might need to move to a newer Bazel build rules in the workspace to get 1.12.5 support. It looks like they dropped support for the 0.16 releases `https://github.com/bazelbuild/rules_go` so maybe 0.17.6. Not sure if that requires changes the other Bazel versions.

ericvn:
I noticed `go.sum` and `go.mod` are updated. Not sure what drives that in this repo. Maybe the Workspace updates caused the changes? Do we need a `go mod vendor` as well to get the _vendor_ directory back in sync with the go modules?

icygalz:
I ran:
angwar@angwar1:~/go/src/istio.io/kind-resources$ go mod verify
all modules verified

PR #1283, opened 2019-05-30
Istio Flakey Test

ChristinaLyu0710-zz:
Contains file that reads past istio-test results of junit.xml located in istio-testing project, sums up the total number of testsuites and cases run and calculates the number of failed tests. The results are written to istio-flakey-test project on Google Cloud Storage.


ChristinaLyu0710-zz:
I signed it!

ChristinaLyu0710-zz:
I signed it!

flakeyTest/jars/a.txt
```diff
@@ -0,0 +1 @@
+
```
utka: do you need this file?

utka (changes requested):
Why do we need all this binary files?

flakeyTest/jars/a.txt
ChristinaLyu0710-zz: I don't need this. Should I delete it? I created this when I created the directory

utka: Yes, I suspect you only need a few files with code. the rest should be deleted/

utka (changes requested):
You just need to have 2 files hese Pair and TotalFlakey

flakeyTest/jars/a.txt
ChristinaLyu0710-zz: Should I also delete all the jar files? But they are needed to compile and run the code

icygalz:
Changes looks to me. Can you also write a README.md file for the folder explaining what your application does and how to use it? We can do it different PR.

/lgtm

PR #1284, opened 2019-05-30
install python requests in istio-builder docker image

utka:


docker/istio/istio/install.sh
```diff
@@ -12,6 +12,7 @@ apt-get -qqy --no-install-recommends install \
   curl \
   lsb-release \
   python \
+  python-requests \
```
ericvn, line 15: More of a repository question, but I notice that the docker/istio_builders/istio_builder/Makefile and CHANGELOG.md seem to keep a running version number and what has changed in those versions. It seems that this change would have incremented that number and highlighted a change. However, based on the next PR, #1285, and a grep of the Istio repos, it seems that maybe this documentation has been abandoned. Or maybe this is a temporary change to do some testing?

If indeed we have changed to procedure, does it make sense to clean up the VERSION in the makable and mark the keeping of versions in the CHANGELOG deprecated.

icygalz:
@utka  you might want to rebase your client for lint-check test.

/lgtm

docker/istio/istio/install.sh
utka: I think it makes sense to deprecate changelog file and remove manual versions update.

icygalz: I think there is confusion here.
docker/istio_builders/istio_builders/ is NOT being used anymore. I heard from our predecessor that it used to be used for release at some point. Currently we are using istio_builder image created from docker/istio/istio/ which does not have changelog file.

PR #1290, opened 2019-05-31
Update test-infra to use go1.12.5

icygalz:
Update test-infra to use go1.12.5


ericvn (commented):
I'm not sure if it's required, and it may break things so maybe try in a new commit that one can back out later if needed, but does the go.mod need to be updated for the github.com/bazelbuild/buildtools change? And maybe add github.com/bazelbuild/rules_go and github.com/bazelbuild/buildtools at the correct levels. It may be updating the versions in the _vendor_ directory will break things.

icygalz:
i dont think it matters. Bazel will look into at WORKSPACE information to get the version it needs to use.

ericvn:
@icygalz OK.

ericvn (approved):

ericvn:
Part of https://github.com/istio/istio/issues/14420

PR #1291, opened 2019-05-31
Add missing integration test gates.

ozevren:


prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -73,6 +73,20 @@ presubmits:
...
+    always_run: true
+    optional: true
+    spec:
+      containers:
+        - <<: *istio_container
+          command:
+            - entrypoint
+            - prow/integ-istioctl-local-presubmit-tests.sh
+      nodeSelector:
+        testing: build-pool
```
icygalz: any reason why you need to use build-pool for this test and not use test-pool?

ozevren: Just following the pattern in the same file. Unit tests are also ran in ta similar way.

icygalz: let's change the tests to use "test-pool" instead of "build-pool". We want to use "build-pool" just for building images if possible..

ozevren: I changed the new tests to use test-pool. We can change the existing ones to use test-pool separately.

icygalz (approved):

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
icygalz: I agree. Thanks!

ozevren:
@icygalz I had to merge with your changes. Could you PTAL and lgtm?

icygalz (approved):

PR #1297, opened 2019-06-03
Update master prow jobs to use new Golang 1.12 and bazel 0.26.0 updates

ericvn:
Will pick up release-1.2 in another PR after verifying `master` works.


ericvn:
FYI https://github.com/istio/istio/issues/14420

PR #1298, opened 2019-06-03
[WIP] Update 1.2 prow jobs to use new Golang 1.12 and bazel 0.26.0 updates

ericvn:
Wait to verify these changes in `master` before merging for `release1.2`

### DIFF
--- a/prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.release-1.2.yaml
+++ b/prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.release-1.2.yaml
@@ -10,7 +10,7 @@ istio_rel_pipeline_spec: &istio_rel_pipeline_spec
     testing: build-pool
 
 istio_rel_pipeline_container: &istio_rel_pipeline_container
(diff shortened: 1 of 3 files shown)


ericvn:
FYI: https://github.com/istio/istio/issues/14420

ericvn:
Need to update to the current images used in master (include kind).

duderino:
This very much needs to be approved by one of the 1.2 release managers.  @fpesce @linsun

ericvn:
@duderino I agree 100% that the release managers have to approve this before even thinking about putting it in. The PR was created to hold a set of changes if they decided to do move release 1.2 builds and prow test to Golang 1.12. The problem with the PR as it sits today, is that the images that are called out don't really point to Golang 1.12 images as we found out in master last week (the image was reverted back to Golang 1.10). My comment earlier was to make this PR at least point to something that is correct, in the case that this PR was used (merged or re-opened (if closed which will likely happen after I make the update) and merged). I will make the updates later today and will then close the PR.

ericvn:
Closing for now. If this PR is used, please verify that the images used updated to match master (the current set matches what was used in master around the second week of June 2019).

PR #1302, opened 2019-06-04
Revert "Update master prow jobs to use new Golang 1.12 and bazel 0.26.0 updates"

ericvn:
Reverts istio/test-infra#1297


howardjohn:
Not sure the problem with the CLA bot, but I know @ericvn has signed as they are a regular contributor

PR #1303, opened 2019-06-04
Add prow config for prow-staging branch

icygalz:
I have pushed new image based on current Dockerfile and use it on prow-staging.yaml's istio_container.


ericvn (approved):
/lgtm

PR #1318, opened 2019-06-10
Making sure only build image job are running on build-pool

icygalz:
Making sure only build image job are running on build-pool and taking out simpleTest-on-KinD from release-1.2 postsubmit tests. This test hasn't succeed yet.


howardjohn (approved):
/lgtm

PR #1326, opened 2019-06-11
Update prow-staging to use prow image with Golang 1.12 and kind

ericvn:
Should I remove the kind container `istio_container_with_kind`, moving the CPU size up to `istio_container` and update the jobs to use `istio_container` as well?


icygalz:
Let's just do this first for prow-staging..

If all tests are passing with KinD packages, we can delete istio_container_with_kind, keep CPU size(this does not prove to help kind test to pass), and every test use istio_container.

PR #1333, opened 2019-06-12
Getting rid istio-presubmit.sh

icygalz:
deleting istio-presubmit job which is responsible on building images.


howardjohn (approved):
/lgtm
on staging so it's safe

PR #1330, opened 2019-06-12
test if flakey-run test runs correctly

ChristinaLyu0710-zz:

PR #1334, opened 2019-06-13
Update prow images for master to Golang 1.12.5 with kind v0.30.0

ericvn:
The base prow image was tested in `prow-staging` in https://github.com/istio/istio/pull/14757.

This PR moves that and associated images to `master`. 

Note: When this is merged, it would be beneficial if someone is available for a bit to possibly revert things in case tests start to fail in `master`. The PR above for `prow-staging` provided feedback that things should work, but want to be safe.

If the approver can check with me on Slack before merging, I can be around to help watch as well.

### DIFF
--- a/prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.master.yaml
+++ b/prow/cluster/jobs/istio-releases/pipeline/istio-releases.pipeline.master.yaml
@@ -10,7 +10,7 @@ istio_rel_pipeline_spec: &istio_rel_pipeline_spec
     testing: build-pool
 
 istio_rel_pipeline_container: &istio_rel_pipeline_container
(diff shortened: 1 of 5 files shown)


ericvn:
Tried to make the `flakeyTest/daily-flakey-test.sh` executable which got the permission failure, but then ran into a javac issue. I suspect that this is being worked in https://github.com/istio/test-infra/pull/1330 (the chmod was in one of them as well). At this point, it looks like a test was added that won't run successfully, so need to wait for the above PR to fix the test (it's passing there).

icygalz:
Don't worry about prow/flakey-run. That's only a trial test.

icygalz:
/lgtm

thank you for the PR :)

PR #1335, opened 2019-06-13
increase resources for proxy

utka:
For master and wasm


icygalz:
this might make other build slower since we only have 10 nodes of that limit size... we can try this out first and we can always increase number of nodes.

/lgtm

howardjohn:
@icygalz could this be related to things like https://k8s-gubernator.appspot.com/build/istio-prow/pr-logs/pull/istio_istio/14822/e2e-dashboard-master/2062 which has been stuck for 10 hours?

icygalz:
it shouldn't be because the test is using test-pool and maria's changes are using build-pool.
test-pool and build-pool are different node pools so they shouldn't starve each other.

icygalz:
btw, that test pass but failed to post finished.json to GCS(i checked the pod).

PR #1336, opened 2019-06-13
Getting rid run after success in presubmit

icygalz:
This has been tested on prow-staging: https://github.com/istio/istio/pull/14772


utka:
Do we have enough resources?

icygalz:
Actually each test in master has been doing BUILD and TEST since a couple weeks ago and don't need to do the istio-presubmit.sh anymore. This is the final plug for getting rid run_after_success in master branch! WOOT WOOT!!

PR #1341, opened 2019-06-14
change periodic yaml

ChristinaLyu:


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -118,6 +144,24 @@ periodics:
     nodeSelector:
       testing: build-pool
 
+  name: istio-flakey-test-periodic
+  context: flakeyTest/daily-flakey-test.sh
+  branches: master
+  labels:
+    preset-service-account: "true"
+  spec:
+    <<: *test_flakey_master_spec
```
fejta, line 153: It looks like this anchor is only used in one location? If so please drop the anchor and put the fields directly into here.

PR #1345, opened 2019-06-17
Add prow linter to master

howardjohn:


howardjohn:
/hold
until https://github.com/istio/istio/pull/14934 merges

howardjohn:
/hold-cancel

@icygalz this is ready now that the test was checked in. Thanks!

icygalz:
Please make circleci:lint to be not required. If not, the failure will
block everybody from merging...

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -47,6 +47,21 @@ presubmits:
       nodeSelector:
         testing: test-pool
 
+  - name: istio-lint-master
```
rshriram, line 50: make circleci lint as non required

howardjohn:
I would prefer to see a few on prow pass then make prow required and remove circle. Otherwise we will block everyone if prow doesn't work and everyone will have to rebase their PRs to pick up the new test which is a pain

icygalz:
The problem is that circleci:lint test after your istio/istio merge will
always be failing, right?

That will affect other people merging because circleci:lint is a required
test. We need to make circleci:lint not required..

howardjohn:
I don't think my change impacts circle test, it has always been flaky and
sometimes gets OOMkilled.

Unless the tiny junit change uses more memory and pushed it over the edge
consistently now?

I

icygalz:
Oo okay, I thought it would always be failing since admin needs to override
the failure to merge #14934..

/lgtm

howardjohn:
it shouldn't, but if it does we should remove it from required sooner

PR #1347, opened 2019-06-18
Delete test-infra config for istio/istio release-0.8

icygalz:
Releaes-0.8 is already end-of-life since long time ago. We don't need to run these tests anymore. Deleting the config.


howardjohn (approved):
/lgtm

PR #1352, opened 2019-06-19
Run e2e-simple tests also for distroless images

kramerul:
This is a PR to activate the tests introduced in https://github.com/istio/istio/pull/14944.

We are not exactly sure, if this is the only place where we need to add this test.

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -433,6 +433,22 @@ presubmits:
         - prow/e2e-simpleTests.sh
       nodeSelector:
         testing: test-pool
+  - name: e2e-simpleTests-distroless-master
+    <<: *job_template
+    optional: true
+    context: prow/e2e-simpleTests-distroless.sh
+    labels:
+      preset-service-account: "true"
+    max_concurrency: 5
+    spec:
+      containers:
+      - <<: *istio_container
+        command:
+        - entrypoint
+        - scripts/build_test_presubmit.sh
+        - prow/e2e-simpleTests-distroless.sh
+      nodeSelector:
+        testing: test-pool
   - name: e2e-simpleTestsMinProfile-master
     <<: *job_template
     optional: true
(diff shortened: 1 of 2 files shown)


howardjohn:
I think we probably want this in `istio.istio.master.yaml` as an optional (non-required) (can be made required if its not flaky for ~a week or so) presubmit test. Right now this is in the release and marked as required, so if a change breaks it it won't fail until we try to release (a daily build most likely, so not that bad) and then block the release.

Also just need to make sure https://github.com/istio/istio/pull/14944 gets merged in first.

kramerul:
We moved the job to `istio.istio.master.yaml` and made it optional.

@howardjohn We are not really sure, if it's OK to call `scripts/build_test_presubmit.sh` within this file.

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -433,6 +433,22 @@ presubmits:
...
+    context: prow/e2e-simpleTests-distroless.sh
+    labels:
+      preset-service-account: "true"
+    max_concurrency: 5
+    spec:
+      containers:
+      - <<: *istio_container
+        command:
+        - entrypoint
+        - scripts/build_test_presubmit.sh
```
howardjohn: Just based on it not being in the other ones, I am a bit suspicious of this. @icygalz is the expert on this so I'll let her take a look, otherwise lgtm

utka: Why do you want to run this script?
https://github.com/istio-releases/pipeline/blob/master/scripts/build_test_presubmit.sh

kramerul: @utka, the distroless docker images are build in `prow/istio-presubmit.sh` called from `build_test_presubmit.sh`. These images are required for this test.

howardjohn (approved):
/lgtm
Thanks!

howardjohn:
@utka @icygalz can you take a look?

utka (approved):

PR #1353, opened 2019-06-19
Make prow lint required, circle optional

howardjohn:
this is just for master for now. We can move to other branches if the release managers want.

Can see test history at: https://prow.istio.io/?job=istio-lint-master

Spot checked the failures and they are all real failures (this is presubmit only)


icygalz:
/lgtm

do we plan to bring prow/lint to release branches?

PR #1362, opened 2019-06-20
istio/istio authz-v2 is not used anymore and deprecated

icygalz:


icygalz:
FYI @yangminzhu

yangminzhu:
Thanks for cleaning this!

PR #1365, opened 2019-06-21
Add conformance post-submit gates.

ozevren:


prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -515,6 +515,18 @@ presubmits:
 postsubmits:
 
   istio/istio:
+    - name: integ-conformance-local-postsubmit-tests-master
```
icygalz: the build failure is because of the indentation. Please remove 2 spaces for the whole test block.

ozevren: Thanks. PTAL.

PR #1366, opened 2019-06-21
Switch circle requirement for prow requirement

howardjohn:
Make sure https://github.com/istio/istio/pull/15068 works fine, then we can turn these tests off as required for master and make the prow ones required.

The e2e-pilot-noauth-v1alpha3-v2 one is already covered by prow and we forgot to remove before


howardjohn:
Removing WIP, looks like the remaining circle one passes on https://github.com/istio/istio/pull/15068.

PR #1368, opened 2019-06-21
Make circleci build nonrequired

howardjohn:
I thought we needed build, but it can actually run fine on its own. See https://circleci.com/workflow-run/8df6e347-e4c0-4438-b559-db10e11d9220 where the one circle test works fine without the build step.

@icygalz can I remove
```
          required_status_checks:	
            contexts:
```
like I did or do I need to set contexts to an empty list?

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -85,9 +85,6 @@ branch-protection:
               protect: true
         istio:
           protect: false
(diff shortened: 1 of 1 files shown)


icygalz:
/lgtm

it works like you did.

PR #1373, opened 2019-06-24
Experimenting with upper resource limit

icygalz:
The flakiness is caused by pods being able to take up higher resources and we couldn't control when the pod takes more resources. By limiting the upper-bound and fit the node's quota, we can avoid the nodes going on CPU contention.

Cons:  the pods will get throttled and test might be slower.


icygalz:
/assign @howardjohn 

what do you think @howardjohn ? 
we can increase the upper bound for memory if needed..

howardjohn (approved):
/lgtm

PR #1376, opened 2019-06-25
Swap prow for circle racetest

howardjohn:
Both will still run, just making circle the required one until prow
racetest is stable.


linsun (approved):
/lgtm

PR #1379, opened 2019-06-25
Upgrade to latest kubectl

howardjohn:
https://github.com/istio/istio/pull/15138 will not work because we use an old kubectl version. This updates the image for prow staging only.

I tested the image correctly installs kubectl.


icygalz:
/lgtm

please double check this work with our tests because kubectl deprecated one of the options(i think it was -a) in version 14 i think.

PR #1380, opened 2019-06-26
Add lint job for operator to prow

sdake:


icygalz:
/lgtm

I will lgtm this for now but you also need to add the configuration in config.yaml for operator repo for prow to protect that repo:
https://github.com/istio/test-infra/blob/master/prow/config.yaml#L86
also you need to add it here for tide to auto-merge the PR:
https://github.com/istio/test-infra/blob/master/prow/config.yaml#L180

PR #1381, opened 2019-06-26
Add an installer prow job to begin the migration

sdake:


prow/cluster/jobs/istio/installer/istio.installer.master.yaml
```diff
@@ -0,0 +1,52 @@
...
+    privileged: true
+  resources:
+    requests:
+      memory: "2Gi"
+      cpu: "3000m"
+    limits:
+      memory: "24Gi"
+      cpu: "3000m"
+
+istio_container_with_kind: &istio_container_with_kind
```
icygalz: You can just delete this istio_container_with_kind section since the istio_container already has KinD package by default.

sdake: thanks this was C&P from istio - but makes complete sene now :)

PR #1383, opened 2019-06-26
Operator cargo cult copy :)

sdake:
Hopefully this is correct and gets things working for the operator. I am unclear
if the tide robot is responsible for running the plugins or if that is some other
robot.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -126,6 +126,7 @@ branch-protection:
                 - "ci/circleci: e2e-pilot-cloudfoundry-v1alpha3-v2"
         operator:
           protect: false
+          branches:
             master:
               protect: true
         proxy:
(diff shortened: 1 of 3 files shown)


icygalz:
/lgtm

tide is only responsible for merging. The plugins are being run by some other prow's component.

prow/config.yaml
```diff
@@ -124,6 +124,10 @@ branch-protection:
                 contexts:
                 - "ci/circleci: racetest"
                 - "ci/circleci: e2e-pilot-cloudfoundry-v1alpha3-v2"
+        operator:
+          protect: false
+            master:
```
icygalz, line 130: i think you need:
 branches:
   master:
     protect: true

PR #1394, opened 2019-06-28
Update build image

howardjohn:
This picks up the kubectl version update change from earlier


howardjohn:
@utka this is causing one of the tests to fail now. Its not a required one so not a big deal, but does cause confusion. The image previously had the changes I needed for this but got built off a stale branch and overwrote the changes

PR #1400, opened 2019-07-01
Check in configs to support RBE

fejta:
ref https://github.com/istio/test-infra/issues/1396

* `.bazelversion` is for use with [bazelisk](https://github.com/bazelbuild/bazelisk)
* Copied most `.bazelrc` values from https://github.com/kubernetes/test-infra/blob/master/.bazelrc
* Fixed implicit bootstrap dependency on yaml library

Note `--remote_instance_name=projects/istio-testing/instances/default_instance` does not yet exist so won't work just yet, but I have tests passing on my RBE cluster.

Issue #1396, opened 2019-07-01
Modernize prow.istio.io


fejta:
- [x] Update the existing cluster to head
  - [x] Apply [breaking changes](https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md) since November to deployments and update, which should get things working - https://github.com/istio/test-infra/pull/1399
      - [x] [Asking](https://istio.slack.com/archives/C6FCV6WN4/p1562019014474000) whether any jobs depend on the default service account (assuming no)
      - [x] Gerrit is irrelevant
      - [x] Create hook rbac
      - [x] build clusters are so far irrelevant
      - [x] run_after_success now irrelevant (yay!)
      - [x] Need to set `--dry-run=false` for horologium and sinker
      - [x] Added explicit `--config-path` flags
  - [x] We will probably miss something, so figure this out and fix it. - https://github.com/istio/test-infra/pull/1406 https://github.com/istio/test-infra/pull/1410

Other activity:
- [x] Create a separate cluster to run jobs and move jobs there
- [x] Clone the prow.k8s.io automation we use  to update prow daily
  - [x] bump script https://github.com/istio/test-infra/pull/1420
  - [x] create periodic job to run the bump script every day
  - [x] Create postsubmit to apply changes to prow/cluster/*.yaml files on merge
- [x] Create an https redirect - https://github.com/istio/test-infra/pull/1427
- [x] Determine if branch protector is used (yes)
  - [x] If so update branch protector and run it before updating other components - https://github.com/istio/test-infra/pull/1397
- [x] Setup ghproxy - https://github.com/istio/test-infra/pull/1428 https://github.com/istio/test-infra/pull/1426
- [x] Migrate branch protector from a cronjob to prowjob (easier to work with, more accessible debugging options)
- [x] Oncall members need to join istio org ([example](https://github.com/istio/community/issues/195))
  - [x] fejta - https://github.com/istio/community/issues/195
  - [x] @cjwagner - https://github.com/istio/community/issues/197 
  - [x] @katharine
  - [x] @michelle192837
  - [x] @chases2
- [x] Switch job viewer from gubernator to spyglass - https://github.com/istio/test-infra/issues/1239 
- [x] Setup RBE for any bazel activity (which at least test-infra uses) - https://github.com/istio/test-infra/pull/1400
- [ ] Delete prow code in istio/test-infra
- [ ] Use bazel to deploy k8s resources
[...]

PR #1401, opened 2019-07-01
Script to install/update rbe into the istio-testing project

fejta:
ref https://github.com/istio/test-infra/pull/1400 #1396 

`bazel --config=remote --remote_instance_name=projects/istio-testing/instances/default_instance` now available to anyone with access to the istio-testing project.

### DIFF
--- a/rbe/install-istio-testing.sh
+++ b/rbe/install-istio-testing.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"
+
+if [[ ! -f install.sh ]]; then
+  echo "ERROR: install.sh not found" >&1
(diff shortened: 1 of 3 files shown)


rbe/install-istio-testing.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [[ ! -f configure.sh ]]; then
+  echo "ERROR: configure.sh not found" >&1
+  echo "  curl https://raw.githubusercontent.com/kubernetes/test-infra/master/rbe/configure.sh -o configure.sh" >&1
+  exit 1
+fi
+
+
+proj=istio-testing
+pool=prow-pool
```
utka, line 19: should this node be created first, or it is part of the install ?

fejta: The script creates an rbe pool with this name

Issue #1408, opened 2019-07-02
config_test failing to parse new durations fields


fejta:
ref #1396 

```console
$ bazel test //prow/... --test_output=streamed

--- FAIL: TestConfig (0.00s)
    config_test.go:27: could not read configs: error unmarshaling ../config.yaml: error unmarshaling JSON: json: cannot unmarshal string into Go struct field DecorationConfig.grace_period of type time.Duration
FAIL
```

https://gubernator.k8s.io/build/istio-prow/pr-logs/pull/istio_test-infra/1401/build-test/782

blocking https://github.com/istio/test-infra/pull/1401
caused by https://github.com/istio/test-infra/pull/1406

PR #1409, opened 2019-07-02
Upgrade vendored prow to fix config_test

fejta:
/assign @utka @howardjohn 

* Favor [vendored deps over external](https://github.com/bazelbuild/bazel-gazelle#bazel-rule)
  - Given that we check in the vendored code, no need to download it each time
* Update prow to master, and match its k8s dependencies (required)
* Update the update_deps.sh script to call update_bazel.sh

https://github.com/istio/test-infra/issues/1408


fejta:
/hold

Let me see if I can switch us to use external

fejta:
Cure is worse than the disease. And for some reason cannot get gazelle external mode to build prow :( So starting with https://github.com/istio/test-infra/pull/1410 for now

/close

PR #1414, opened 2019-07-03
Increase linter memory requests

### LINKED ISSUE #15319: Linter getting OOMKilled in prow
(NOTE: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities/>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>
)

**Bug description**

**Affected product area (please put an X in all that apply)**

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer 

https://gubernator.k8s.io/build/istio-prow/pr-logs/pull/istio_istio/15259/istio-lint-master/779

Need to investigate the menorry requests and node memory amounts. this test needs ~16gb memory so probably should explicitly require that instead of hoping for the best

howardjohn:
Fixes https://github.com/istio/istio/issues/15319

We have a variety of node sizes, some with 52gb, so scheduling these should be ok. The problems were happening when they were scheduled on highcpu nodes which have 6gb

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -53,7 +53,19 @@ presubmits:
     always_run: true
     spec:
       containers:
-      - <<: *istio_container
+      # Lint requires a large amount of memory
+      # See https://github.com/istio/istio/issues/14888 before lowering requests
+      - image: gcr.io/istio-testing/istio-builder:v20190628-31457b43
+        # Docker in Docker
+        securityContext:
+          privileged: true
+        resources:
+          requests:
+            memory: "16Gi"
+            cpu: "3000m"
+          limits:
+            memory: "24Gi"
+            cpu: "3000m"
         command:
         - entrypoint
         - prow/istio-lint.sh


prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -53,7 +53,19 @@ presubmits:
     always_run: true
     spec:
       containers:
-      - <<: *istio_container
+      # Lint requires a large amount of memory
+      # See https://github.com/istio/istio/issues/14888 before lowering requests
+      - image: gcr.io/istio-testing/istio-builder:v20190628-31457b43
+        # Docker in Docker
+        securityContext:
+          privileged: true
```
fejta, line 61: We need this because it calls docker build?

fejta (approved):

fejta (commented):
/lgtm

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
kyessenov: Is this because we call out to docker run somewhere in the script? We should refactor that to be pure binaries (or package it in the builder).

howardjohn: This was a copy paste of the previous `istio_container` we were using. I don't think the linter specifically needs this, we just were using the same config for all our tests and I guess one of them needed it?

I don't think the lint uses docker

Issue #1415, opened 2019-07-03
Add checkconfig to presubmit


howardjohn:
k8s.io/test-infra/prow/cmd/checkconfig

This is really useful for finding issues with the config. It caught at least one bug for me already (submitting a PR now).

One problem is if we do --strict it will fail because we use anchors which are unknown fields, but if we don't do --strict we won't notice failures

Issue #1403, opened 2019-07-02
KinD tests fail to create network overlay


howardjohn:
Related: https://github.com/istio/test-infra/pull/1395, https://github.com/kubernetes-sigs/kind/issues/677

Example failure: https://gubernator.k8s.io/build/istio-prow/logs/istio-kind-simpleTest-master/362

I don't know why it is failing. If I run the job locally with `phaino` it works fine, and I have my own prow cluster set up that has the exact same job and it passes. Maybe something with the GKE cluster is different?

PR #1416, opened 2019-07-03
Fix KinD volume syntax

howardjohn:
https://github.com/istio/test-infra/issues/1415 would have caught this bug

Getting this KinD stuff to work is pretty frustrating because it only runs on postsubmit and there is no easy way to iterate on it locally. I tested this on my own prow cluster I spun up and it works though (https://github.com/howardjohn/istio/pull/4).

Note the test won't actually pass without https://github.com/istio/istio/pull/15308 but this PR isn't blocked by that one; both are needed for the test to pass

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -643,23 +643,25 @@ postsubmits:
         command:
         - entrypoint
         - prow/e2e-kind-simpleTests.sh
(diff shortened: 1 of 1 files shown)


howardjohn:
Related to https://github.com/istio/test-infra/issues/1403

fejta (approved):
/lgtm
/hold

Feel free to `/hold cancel` when you're ready for this to merge

fejta:
no hyphen
/hold cancel

howardjohn:
Thanks!

PR #1426, opened 2019-07-03
Set up ghproxy for Istio's Prow instance.

cjwagner:
I tested this in my testing cluster and it worked as expected.

Once this has merged and been deployed I'll confirm it is working in the production cluster as well and then edit the hook deployment to use it. Assuming no problems I'll make a PR to make all cluster components use it.

/assign @fejta @utka @geeknoid


fejta (approved):
/ok-to-test
/lgtm

fejta:
ref https://github.com/istio/test-infra/issues/1396

PR #1429, opened 2019-07-03
Use RBE for --config=ci

fejta:
/assign @howardjohn @utka 
/cc @cjwagner

ref https://github.com/istio/test-infra/issues/1396

Will it blend?! :jack_o_lantern:


fejta:
Woo! https://gubernator.k8s.io/build/istio-prow/pr-logs/pull/istio_test-infra/1429/build-test/815

PR #1432, opened 2019-07-03
Merge approvers and reviewers list

fejta:
By default the reviewers list == approvers, which is what we generally want


fejta:
Ping!

Issue #1433, opened 2019-07-03
Allow github reviews to mutate the lgtm label


fejta:
New behavior:
* An approving github review is equivalent to leaving a `/lgtm` and `/approve` comment
* A changes requested github review is equivalent to `/lgtm cancel` and `/approve cancel` comment.
* If for some reason you only want to approve or only lgtm, you can still use the `/approve` and `/lgtm` commands
  - In particular, `/lgtm` will no longer also approve.

Reminder about how `/lgtm` and `/approve` work:
* `/lgtm` only adds the `lgtm` label to the PR
  - `/lgtm cancel` removes this label from the PR
* `/approve` only approves changes to files in the PR that you own
  - `/approve cancel` removes your approval to files you own
  - If and only if all each changed files in the PR has an `/approve` from someone who is an approver for that file will the `approved` label appear.
* PRs must have both an `approved` and `lgtm` label in order to merge.

---

Recommended practice is to make an approving github review act as a `/lgtm /approve` comment (and a request changes act as `/lgtm cancel /approve cancel`).

By default a github approve is /approve, but we can/should also do this for /lgtm, with the following stanza:

https://github.com/kubernetes/test-infra/blob/132f32f32a0e0ca2ea4b82915f3c24e7f2906413/prow/plugins.yaml#L123-L129

```
lgtm:
- repos:
  - bazelbuild
  review_acts_as_lgtm: true
- repos:
  - kubernetes/test-infra
  review_acts_as_lgtm: true
```

ref https://github.com/istio/test-infra/issues/1396

fejta:
https://github.com/istio/test-infra/pull/1447 is an example where github approve added both approve and lgtm labels.

/close

fejta:
/reopen

Let's keep `lgtm_acts_as_approve: true` for a bit longer to give people time to adjust to the new system.

PR #1434, opened 2019-07-03
Allow github reviews to add/remove LGTM

fejta:
/hold

What channels should we go through for this sort of proposal?
/cc @howardjohn @cjwagner @utka 

ref https://github.com/istio/test-infra/issues/1433

NOTE: you can still `/lgtm` or `/approve` whenever you want to just lgtm or approve, respectively (aka current behavior remains the same).

These plugins did not behave this way initially because they predate github's review functionality.

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -41,6 +41,9 @@ slack:
 
 lgtm:
 - repos:
+  - istio
+  review_acts_as_lgtm: true
+- repos:
   - istio/api
   - istio/istio
   - istio/operator


howardjohn (approved):
/lgtm

The best channel is probably just asking on #test-and-release or more formally at the test and release WG meeting (every tuesday). 

I don't know that I have ever actually seen someone `/approve` without an lgtm though so I don't think there would be any concerns with it. I have seen people give gituhb approval without realizing it did nothing many times though :slightly_smiling_face:

howardjohn:
In test and release meeting we decided we want this, feel free to cancel

fejta:
Great!

/hold cancel

PR #1435, opened 2019-07-04
Increase memory request 2 -> 3gb

### LINKED ISSUE #1413: Weird behavior with multiple tests runs
https://gubernator.k8s.io/builds/istio-prow/pr-logs/pull/istio_istio/15259/istio-lint-master/

Ive seen this on other ones too - there is a bunch of reported test runs for no apparent reason. I don't think all of these correlates to an actual pod being run though, but haven't investigated much

@fejta

howardjohn:
codecov, racetest, and unit test are using up to 3gb in the tests,
causing them to get frequently OOMKilled.

Fixes https://github.com/istio/test-infra/issues/1413

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -11,7 +11,7 @@ istio_container: &istio_container
     privileged: true
   resources:
     requests:
-      memory: "2Gi"
+      memory: "3Gi"
       cpu: "3000m"
     limits:
       memory: "24Gi"


howardjohn:
/assign @utka @icygalz @fejta 

Alternatives:
* We get metrics on individual test's CPU/mem usage and assign more fine-tuned requests
* We get rid of our highcpu nodes which only have 6GB memory. Right now if you get 2 tests on one of these nodes and one of them is racetest or codecov its basically 100% chance of OOM, whereas with the standard nodes they will go above their requests but not be OOM because the node has extra memory

Or some combination of the above

icygalz:
we set the memory limit to 24G (which is the node's max size). Does the pod not expand to that limit? or is that causing oom in the nodes itself and kill the node?
I saw that we now have several node pools with different configuration. We can add labels to this pools and set certain tests to only run on those pools(for example: high memory requirement tests running on high-mem node pool, etc)

howardjohn:
@icygalz yeah the whole node is getting OOM. This happened because we recently added two nodepools with highcpu-8 config which only have 6GB. I think test pool 4/5. The labels would work too, or just completely get rid of the highcpu nodes. I don't have much preference as long as our tests aren't getting OOMkiled

icygalz:
if the whole nodes getting killed, then just changing the request amount will not help. Because the pods will try to expand their memory until its memory limit is reached(which could be higher than what the node has[24G>6G]). You need to set request==upper limit like what we did for CPUs. Make sure the number fits in the nodes.

howardjohn:
@icygalz I don't think the node itself is getting killed, but what happens is the pod uses 3GB, and then it gets OOMkilled because the node is out of memory -- the node still stays around though. If we set the requests to 3gb then as long as the test never exceeds 3gb it will never get OOMKilled. The problem is we are (for example) packing 3 tests with 2GB requests into the node, each one uses 3GB, so 9Gb total which is over the 6GB node limit. If each were requesting 3GB then only 2 would be scheduled and it would be fine.

The problem I have with setting the memory limit is we are then asking kubernetes to kill our test pods -- which we never want to happen. I would rather the pods go over their requests and be allowed to continue and we can tune the requests.

If we have metrics on the tests CPU and memory usage this would be a lot easier though

icygalz:
/lgtm

ok, we can try this out first and see whether this will help the issue. we can fine tuned the pods&node pools again later.

fejta:
Can we get an OWNERS file in https://github.com/howardjohn/test-infra/tree/dadc1eb18b794620b0e6b8cffef1454ea0e5320c/prow/cluster/jobs/istio/istio (or even .../jobs/istio if we think there's a default set for jobs in all repos). 

That way we'll route these reviews to the right people. Ideally we do not gate job config changes on the test-infra team.

PR #1436, opened 2019-07-08
Simplify test-infra configs

fejta:
/assign @icygalz @utka @howardjohn 

* Remove containter anchor as each is only used once.
* Remove unnecessary branch_spec (test-infra has only one branch)


howardjohn (approved):
/lgtm

PR #1446, opened 2019-07-09
Update rules_go and bazel-gazelle repos

fejta:
/assign @utka @cjwagner @howardjohn 

Also configure gazelle to use the checked in `//vendor/foo/bar` deps (that already exist) rather than `@foo//bar` remote repos.


fejta:
PTAL, sorry was trying to keep the `repos.bzl` update in its own PR but I need to merge it with this one.

fejta:
PR now passes tests in exchange for being more invasive.

PR #1447, opened 2019-07-09
Skip report on installer until stable

howardjohn:
The code needed for this is not in installer yet and the code we have doesn't work yet, will be a bit cleaner to not report this

See https://github.com/istio/installer/pull/258


fejta (approved):

fejta (commented):
Can we get an OWNERS file in https://github.com/istio/test-infra/tree/6d44bebb627b85d7521b1ed0e75056ffdfdc370b/prow/cluster/jobs/istio/installer or https://github.com/istio/test-infra/tree/6d44bebb627b85d7521b1ed0e75056ffdfdc370b/prow/cluster/jobs/istio so blunderbuss routes these PRs to the right people?

PR #1451, opened 2019-07-09
Create owners files for jobs istio and installer

howardjohn:
Owners were chosen by top level owners in the repos + Diana and Maria


fejta (approved):
/hold

PR #1456, opened 2019-07-10
Update branch protection settings.

geeknoid:
- Add support for a number of our repos.

- Standardize on a common release-* for our release branches.

- Disable tide stuff on istio/api to transition to mergify.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -51,31 +51,11 @@ github_reporter:
   - postsubmit
 
 blocked_branches: &blocked_branches
-  release-0.1: &blocking_merge
+  release-*: &blocking_merge
     required_status_checks:
       contexts:
       - "merges-blocked-needs-admin"
     protect: true
-  release-0.2:
-    <<: *blocking_merge
-  release-0.3:
-    <<: *blocking_merge
-  release-0.3.0:
-    <<: *blocking_merge
-  release-0.4:
-    <<: *blocking_merge
-  release-0.4.0:
-    <<: *blocking_merge
-  release-0.5:
-    <<: *blocking_merge
-  release-0.6:
-    <<: *blocking_merge
-  release-0.7:
-    <<: *blocking_merge
-  release-0.8:
-    <<: *blocking_merge
-  release-1.0:
-    <<: *blocking_merge
 
 branch-protection:
   allow_disabled_policies: true
@@ -87,23 +67,35 @@ branch-protection:
       required_pull_request_reviews:
         required_approving_review_count: 1
       restrictions:
-        teams:
-        - repo-admins
       repos:
         api:
           protect: false
           branches:
             <<: *blocked_branches
-            release-1.2:
+            master:
               protect: true
-              required_status_checks:
-                contexts:
-                - "merges-blocked-needs-admin"
-            release-1.1:
+        istio.io:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            master:
               protect: true
-              required_status_checks:
-                contexts:
-                - "merges-blocked-needs-admin"
+        operator:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            master:
+              protect: true
+        cni:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            master:
+              protect: true
+        installer:
+          protect: false
+          branches:
+            <<: *blocked_branches
             master:
               protect: true
         istio:
@@ -207,7 +199,6 @@ tide:
     - needs-ok-to-test
     - needs-rebase
   - repos:
-    - istio/api
     - istio/test-infra
     - istio/istio
     - istio/proxy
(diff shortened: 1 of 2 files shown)


prow/plugins.yaml
```diff
@@ -5,12 +5,17 @@
 triggers:
 - repos:
   - istio/api
-  - sebastienvas/core
+  - istio/common-files
```
fejta: Best practice here is to trigger at the `istio` and `istio-releases` org level, rather than each repo individually.

Historically that would have caused issues if the repos didn't have any prow jobs (I think comment spam). It is no longer the case. 

This is how we configure the kubernetes org:
* https://github.com/kubernetes/test-infra/blob/d4a5312e1597a3be2d8f0ae96ca4c40e94b092a0/prow/plugins.yaml#L5-L12
* https://groups.google.com/d/msg/kubernetes-dev/D8_6DtuhyS4/IQHYNXqRBwAJ

prow/plugins.yaml
```diff
@@ -29,6 +34,9 @@ config_updater:
       namespace: boskos
     prow/cluster/jobs/**/*.yaml:
       name: job-config
+    label_sync/labels.yaml:
```
fejta: Do we want to include the label sync change in this PR?

prow/config.yaml
```diff
@@ -51,31 +51,11 @@ github_reporter:
   - postsubmit
 
 blocked_branches: &blocked_branches
-  release-0.1: &blocking_merge
+  release-*: &blocking_merge
```
fejta, line 54: a) [Tide blockers](https://github.com/kubernetes/test-infra/tree/master/prow/tide/blockers) is a better way to achieve this effect. Ask @cjwagner about it. The [blockade plugin](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/blockade) is something else to consider.

b) I suspect admins simply press the merge button? The [override plugin](https://github.com/kubernetes/test-infra/blob/master/prow/plugins/override/override.go) is probably a better alternative to use here if you're not already aware of it.

c) yaml anchors tend to make it harder to understand these config files, which limits the number of people willing to maintain them. Best to avoid them and inline the values each time. Use a script to generate the file when/if this becomes onerous.

prow/config.yaml
```diff
@@ -87,23 +67,35 @@ branch-protection:
...
             <<: *blocked_branches
-            release-1.2:
+            master:
               protect: true
-              required_status_checks:
-                contexts:
-                - "merges-blocked-needs-admin"
-            release-1.1:
+        istio.io:
+          protect: false
```
fejta, line 78: What branches do we want to leave unprotected and why?

prow/plugins.yaml
```diff
@@ -44,9 +52,7 @@ lgtm:
   - istio
   review_acts_as_lgtm: true
 - repos:
-  - istio/api
```
fejta: What is the intent here?

prow/plugins.yaml
geeknoid: Thanks, updated.

prow/config.yaml
geeknoid: The non-official branches are left unprotected. So branches developers create in an ad-hoc fashion as part of doing a PR. We discourage the practice, but it's common nevertheless.

prow/plugins.yaml
geeknoid: Like most of our other repos, istio/api and istio/istio are being moved to the mergify model for handling merges and stuff, and we want people to just use the GitHub UI in the normal way.

prow/plugins.yaml
geeknoid: Whoops, nope.

prow/config.yaml
geeknoid: As mentioned previously, most of our repos don't use tide and we're moving istio/api and istio/istio off of tide as well. So I don't believe the plugins you suggest  would work for us in that context.

As for the yaml anchors, I think it substantially increases readability and maintainability, so I'd like to keep 'em as-is. For most of our repos so far, very few humans have been dealing with branch protection stuff (i.e. me), so I'm not worried about a constraint on people not being familiar with the syntax.

geeknoid:
PTAL

fejta (commented):
Can you pull the plugins.yaml changes into their own PR? They need further discussion.

Other things LGTM

geeknoid:
Done, changes to plugins.yaml were removed.

prow/config.yaml
```diff
@@ -207,7 +211,6 @@ tide:
     - needs-ok-to-test
     - needs-rebase
   - repos:
-    - istio/api
```
fejta, line 210: Can we also move the tide changes to their own PR?

geeknoid:
Given our discussion, can I check this in as-is?

Thanks.

fejta (approved):
Sure SGTM, merge at your leisure

/hold

fejta:
Prefer a `/hold cancel` over manual merge

fpesce:
Yeah, what just happened here @geeknoid ?

PR #1459, opened 2019-07-11
rewrite config with new branches

ChristinaLyu0710-zz:
The code takes a command line input of new branch name. It looks for branches: sections in the config.yaml file and creates a new branch field with the given branch name. Inside the new branch field are all the information from the master field inside the same branches: field plus the "merges-blocked-needs-admin" part if it is not already present in master.

### DIFF
--- a/prow/rewriteConfig.go
+++ b/prow/rewriteConfig.go
@@ -232,7 +232,7 @@ func main() {
 	var branchName string
 	flag.StringVar(&branchName, "NewBranchName", "", "A new branch to add to the original file.")
 	var repoNames string
-	flag.StringVar(&repoNames, "ReposeToAdd", "proxy, istio, istio-releases", "Repo names to add new branch to, separated by `,`.")
+	flag.StringVar(&repoNames, "ReposeToAdd", "proxy,istio,istio-releases", "Repo names to add new branch to, separated by `,`.")
 
 	flag.Parse()
 
--- a/prow/rewriteConfig.go
+++ b/prow/rewriteConfig.go
@@ -83,12 +83,13 @@ func getContentOfNewBranch(branchContent string, newBranch string) (int, string)
 		contextsString := strings.Repeat(" ", spacesInMaster+tab) + "contexts:\n"
 
 		if strings.Compare(masterContentSplit[len(masterContentSplit)-1], "") != 0 {
-			masterContent = masterContent + "\n"
+			masterContent += "\n"
 		}
 
 		masterContent = masterContent + statusCheckString + contextsString
 	}
 
+	// Find new context index after adding context lines to original master content
 	contextInd = contextRe.FindStringSubmatchIndex(masterContent)
 	contextStart := contextInd[0] + 1
 	contextContent := masterContent[contextStart:]
@@ -97,16 +98,17 @@ func getContentOfNewBranch(branchContent string, newBranch string) (int, string)
 		adminLine := strings.Repeat(" ", contextSpaces) + "- \"merges-blocked-needs-admin\"\n"
 
 		if strings.Compare(string(masterContent[len(masterContent)-1]), "\n") != 0 {
-			masterContent = masterContent + "\n"
+			masterContent += "\n"
 		}
 
-		masterContent = masterContent + adminLine
+		masterContent += adminLine
 	}
 
 	withinMaster := spacesForBranch + newBranch + ":\n" + masterContent
 	return masterStart, withinMaster
 }
 
+// Process source byte slice to extract only the repos specified by users to add new branches to.
 func findRepo(source []byte, repos []string, newBranch string) string {
 	sourceString := string(source)
 	eachRepos := strings.Split(sourceString, "repos:")
--- a/prow/rewriteConfig.go
+++ b/prow/rewriteConfig.go
@@ -1,3 +1,17 @@
+// Copyright 2018 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package main
 
 import (
@@ -167,7 +181,7 @@ func main() {
     newBranch := os.Args[1]
     filename := "config.yaml"
     
-    source, err = ioutil.ReadFile(filename)
+    source, err := ioutil.ReadFile(filename)
     if err != nil {
         log.Println("error when reading file")
         os.Exit(1)
@@ -176,7 +190,7 @@ func main() {
     resultString := findMaster(source, newBranch)
     resultBytes := []byte(resultString)
 
-    err := ioutil.WriteFile("config.yaml", resultBytes, 0600)
+    err = ioutil.WriteFile("config.yaml", resultBytes, 0600)
     if err != nil { 
         log.Println("error when writing file")
     }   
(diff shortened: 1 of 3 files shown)


icygalz:
/ok-to-test

I don't think I am the right person for reviewing this but I think this is pretty safe to run tests so I am giving rubber stamp ok-to-test.
/unassign @icygalz 
/assign @utka

fejta (commented):
Please try and reuse https://github.com/kubernetes/test-infra/tree/master/experiment/config-forker to achieve this effect.

If you need some new behavior from this tool, please try and upstream those changes (so everyone can benefit from them).

howardjohn:
It seems like we would be better off unmarshalling the config into a JobConfig rather than trying to do this raw string parsing

howardjohn:
See also https://github.com/istio/test-infra/issues/1471

utka:
I think the problem with unmarshing is that it changes the order. After this script runs, I would like to have extremely simple PR created, which will be easy to review. Adding re-ordering does not make it easier to review.

ChristinaLyu0710-zz:
I used job config and it's unmarshal the yaml files with anchors to the repeated fields on the top back to each section that contains the anchor. This would cause unnecessary diffs when I marshal job config objects into config.yaml again. Also the original yaml contains uses of double quotation marks that are escaped when unmarshaled to job config objects. Since I want to eliminate unnecessary diffs when updating the yaml file with new branches, I parsed the config file as a string not job config objects.

howardjohn:
What does this actually do? I don't understand from the code. Can you add an example or readme or docs?

ChristinaLyu0710-zz:
I'll add a test code

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// Call function with go run rewriteConfig.go <new_branch_name>.
+func main() {
+	if len(os.Args) < 2 {
+		log.Println("Please provide new branch name")
+		os.Exit(1)
+	}
+	newBranch := os.Args[1]
+	filename := "config.yaml"
```
utka: Change it to a flag with path to the file.

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+
+	source, err := ioutil.ReadFile(filename)
+	if err != nil {
+		log.Println("error when reading file")
```
utka: Print the err as well

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+		os.Exit(1)
+	}
+
+	resultString := findMaster(source, newBranch)
+	resultBytes := []byte(resultString)
+
+	err = ioutil.WriteFile("config.yaml", resultBytes, 0600)
+	if err != nil {
+		log.Println("error when writing file")
```
utka: Print err

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return -1, ""
+}
+
+// Convert read result into string and split it based on "branches:" to make it
+// simpler to find "master:" and its contents. Get output from getContentOfNewBranch().
+// Add new branch content before master section and rejoin the string.
+func findMaster(source []byte, newBranch string) string {
+	sourceString := string(source)
+	eachBranches := strings.Split(sourceString, "branches:")
```
utka, line 213: Instead of looking for all master branches, can you add the check for repos, we want to add new branch info only for 
[proxy, istio, istio-releases]

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+func copySlice(input []string) []string {
+	newSlice := []string{}
+	newSlice = append(newSlice, input...)
+
+	return newSlice
+}
+
+// Find index of master section in each branch split for the new branch and
+// get content from master section with mandatory "merges-blocked-needs-admin" part.
+func getContentOfNewBranch(branchLines []string, newBranch string) (int, string) {
```
utka: This looks extremely complex.
will it work if you find  "  master:" string, count number of spaces for master and find next line with this number of leading spaces?

Is there anything else which need to be done? Am I missing something?

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+		}
+
+		if contextStart != -1 && contextStop == -1 {
+			contextStop = masterStop - masterStart + 1
+		}
+		containsNeedsAdmin := false
+		newBranchSlice := copySlice(branchLines[masterStart : masterStop+1])
+		if contextStart != -1 && contextStop != -1 {
+			for n := contextStart; n < contextStop; n++ {
```
utka: You don't need this loop if you use  branchLines as a string

prow/rewriteConfig.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			if !containsNeedsAdmin {
+				numSpaces := countLeadingSpace(newBranchSlice[contextStart])
+				stringAdmin := strings.Repeat(" ", numSpaces) + "- \"merges-blocked-needs-admin\""
+
+				newBranchSlice = appendAtIndex(newBranchSlice, stringAdmin, contextStop-1)
+			}
+
+		} else {
```
utka: if it is already contains needs admin why we need to do all these additional steps?

utka (changes requested):

prow/rewriteConfig.go
```diff
@@ -0,0 +1,214 @@
...
+	if contextInd == nil {
+		masterContentSplit := strings.Split(masterContent, "\n")
+		masterContentLine := masterContentSplit[0]
+		spacesInMaster := countLeadingSpace(masterContentLine)
+
+		tab := spacesInMaster - spacesForBranchNum
+		statusCheckString := strings.Repeat(" ", spacesInMaster) + "required_status_checks:\n"
+		contextsString := strings.Repeat(" ", spacesInMaster+tab) + "contexts:\n"
+
+		if strings.Compare(masterContentSplit[len(masterContentSplit)-1], "") != 0 {
```
utka, line 95: why do you need this check?

ChristinaLyu0710-zz: I cannot say for sure whether or not the previous line ends with a "\n" so if I just add the lines directly it might end up trailing after the previous line without having it's own line.

prow/rewriteConfig.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		masterContent = masterContent + adminLine
+	}
+
+	withinMaster := spacesForBranch + newBranch + ":\n" + masterContent
+	return masterStart, withinMaster
+}
+
+func findRepo(source []byte, repos []string, newBranch string) string {
```
utka: Can you add comments to what are you looking for in this function?
for example line 118, why there is if statement?

prow/rewriteConfig.go
```diff
@@ -0,0 +1,216 @@
...
+	sourceString := string(source)
+	eachRepos := strings.Split(sourceString, "repos:")
+	if len(eachRepos) < 1 {
+		return sourceString
+	}
+	for i := 1; i < len(eachRepos); i++ {
+		repoSection := eachRepos[i]
+		repoLines := strings.Split(repoSection, "\n")
+		var numSpaceForRepo int
+		if strings.Compare(repoLines[0], "") == 0 && len(repoLines) > 1 {
```
utka: Just trim the end and beginning

prow/rewriteConfig.go
```diff
@@ -0,0 +1,216 @@
...
+		leadingSpace := strings.Repeat(" ", numSpaceForRepo)
+		for _, repo := range repos {
+			repoLine := leadingSpace + repo + ":\n"
+			repoSplit := strings.Split(repoSection, repoLine)
+			if len(repoSplit) == 1 {
+				continue
+			}
+
+			generalReposRe := regexp.MustCompile("\n" + leadingSpace + "\\S+:\n")
+			for j := 1; j < len(repoSplit); j++ {
```
utka: Can you use split instead of loop?

utka:
Add Readme and it's good to go.

PR #1463, opened 2019-07-11
Disable max concurrency on jobs without external dependencies

howardjohn:
These tests do not have external limiting factors (boskos, namely). We can let kubernetes handle scheduling

Right now we have 25 jobs blocked by these https://prow.istio.io/?state=triggered


howardjohn:
@utka can you take a look? This is artificially slowing down our presubmits

PR #1465, opened 2019-07-12
Configure most things at the org level

fejta:
/hold

We should try and maximize what we configure at the org level. Configuring things at the repo level is a recipe for a fragmented contributor experience. Let's try and minimize the volume of repo-level snowflakes.


geeknoid:
Do we still need this PR?

PR #1467, opened 2019-07-12
Make prow jobs non-optional

sdake:
Affects istio-operator


howardjohn:
/lgtm

See https://github.com/istio/test-infra/pull/1466 if you need circleci ones required as well

PR #1468, opened 2019-07-12
Adjust the prow plugins we use on our repos.

geeknoid:


prow/plugins.yaml
```diff
@@ -72,101 +69,117 @@ approve:
...
+    - needs-rebase
+    - size
+    - trigger
+
+  istio/common-files:
+    - needs-rebase
+    - size
+    - trigger
 
   istio/installer:
```
howardjohn: no wip? Seems like something we want

prow/plugins.yaml
```diff
@@ -72,101 +69,117 @@ approve:
...
-  - golint
-  - help
-  - hold
-  - lgtm
-  - lifecycle
-  - slackevents
-  - trigger
-  - verify-owners
-  - wip
+  sebastienvas/core:
```
fejta: Why do we need this?

prow/plugins.yaml
```diff
@@ -72,101 +69,117 @@ approve:
...
-  - help
-  - hold
-  - lgtm
-  - lifecycle
-  - slackevents
-  - trigger
-  - verify-owners
-  - wip
+  istio/pkg:
+    - needs-rebase
```
howardjohn: I think these need to be under external plugins

fejta (commented):
It will be more efficient and consistent to configure things at the org level rather than one snowflake repo at a time.

Special casing repos needs to be rare and discouraged, not the norm.

Let's take https://github.com/istio/test-infra/pull/1465 instead

prow/plugins.yaml
geeknoid: In  general, I don't want to configure any of the /foo commands. You can set the WIP progress label through the GH UI instead. This eliminate a spurious email message to all reviewers and keeps the PR cleaner.

prow/plugins.yaml
geeknoid: I didn't change these, I just moved them around to alphabetize.

prow/plugins.yaml
geeknoid: Not sure. If was there before...

geeknoid:
@fejta I don't want to take #1465 as it enables way more plugins than we want/need in our repos. I'm happy to consolidate the settings I have here at the org level for simplicity.

prow/plugins.yaml
howardjohn: the wip one also applied it based on the pr being a draft or having wip in the title. if we change this we should send out a notice to people so they don't think their PR will be marked as wip and have it get merged

prow/plugins.yaml
geeknoid: I'll remove it.

fejta:
There is generally little harm in enabling plugins more broadly than where they are used -- just don't use them. But I see no need to prevent others from using them -- especially since long term we want to provide people with a consistent experience and expectations across the org using prow.

prow/plugins.yaml
geeknoid: Yep. Fixed

prow/plugins.yaml
geeknoid: Oh, I didn't know about that behavior. I've added wip then.

geeknoid:
The benefit to enabling only meaningful plugins is to drive a consistent experience across repos, to reduce the email spam generated by GitHub, to reduce the spammy content being inserted into PRs which detracts from the real useful payload in a PR.

When I enable mergify on istio/istio, then we can be using a small consistent set of plugins across the whole org.

prow/plugins.yaml
fejta: Please remember to configure mergify to stay aware of these labels.

fejta:
Using mergify instead of tide will be a temporary state of affairs, so we shouldn't plan too much around its behavior and value.

Training people now to avoid prow functionality they already know and like just so that we have to retrain them to start using it again later seems user-hostile and unnecessary churn.

geeknoid:
Mergify is configured to look at the right labels across the repos.

A criteria for us to switch to Tide is that it can operate using the normal
GitHub workflow, like mergify does. Istio engineers have edit permissions
on the labels, for example, and we'd like to keep it that way in order to
stay as close as possible to the mainstream GitHub workflow. And as I
mentioned, use of /foo commands results in email and PR spam which we're
not thrilled about.

geeknoid:
Ping.

fejta:
Tide operates how you configure it, and you can configure it to use normal github workflow.

I do not think we should prioritize work to disable prow functionality.

cjwagner:
> Istio engineers have edit permissions
on the labels, for example, and we'd like to keep it that way in order to
stay as close as possible to the mainstream GitHub workflow.

Prow expects that users do not have write permission to the repo and cannot edit labels. This is how Prow works around GitHub's limited permissions system: https://github.com/kubernetes/test-infra/blob/master/prow/scaling.md#working-around-githubs-limited-acls

While we do prefer mainstream GitHub workflows whenever it is reasonable, we heavily favor safety and usability over following the standard GitHub workflow. Using labels was/is the most appropriate way to attach state to a PR on GitHub to track process completion for things like the CLA check. Status contexts and checks are associated with a commit and checks also require a GitHub App which doesn't get full access to the GH API. That then necessitates that we disallow users from modifying labels directly and instead have Prow act as a permissions broker.

ssuchter:
I'm new-ish here, and don't understand something.

> Using labels was/is the most appropriate way to attach state to a PR on GitHub to track process completion for things like the CLA check.

Ok, so I get that this means a robot will add the label saying "ok the CLA check has been done"

> Status contexts and checks are associated with a commit and checks also require a GitHub App which doesn't get full access to the GH API. That then necessitates that we disallow users from modifying labels directly and instead have Prow act as a permissions broker.

I don't understand why it follows that users must be disallowed from modifying labels directly. The only thing I can imagine is that there is a concern that the users do actions that conflict with what the robots are trying to do. Seems to be that since we're talking about a limited number of users, we could just educate them?

Please help me if I'm missing something.

howardjohn:
Once someone joins the Istio org (https://github.com/orgs/istio/people) they will be given access to modifying labels. Right now the bar for joining the org is pretty low, and contributors join after landing just one PR. One of the reasons for this is that they need to be in the org in order to be assigned issues, and to have tests run automatically (we don't run tests for people outside the org in case their test case attempts to abuse the privileges in the CI system).

What this means is any of the ~160 members can just add the `lgtm` label to their PR and getting a change merged. I agree we can/should educate people that this is not acceptable, but ideally it is impossible not just unacceptable. Historically, even some of our largest contributors have broken this rule.

What isn't mentioned here, is that we don't have to use the labels. Tide works based on issuing a github query. Right now that looks like `label:lgtm label:approved ...` and if it matches all of those it will merge. We can adjust this to `review:approved ...` and just look at native github approvals. Note this will require migrating to CODEOWNERS (github concept) from OWNERS (prow concept) - but shouldn't be too hard.

ssuchter:
Thanks, with that scale of users, I understand now that you wouldn't want to just let anyone add the lgtm label and have their changes get merged.

fejta:
> We can adjust this to review:approved ... and just look at native github approvals.

Yes, changing tide to query for `review:approved` seems like a good approach if we need/want everyone to have write access to all labels in all repos.

> they need to be in the org in order to be assigned issues

This is actually out of date! Today, anyone who has commented on an issue can be assigned to that issue -- even people who are not in the org and are not a repo collaborator (encourage new contributors to try it out with the /assign command).

This changed within the last couple weeks and broke some assumptions prow made about repo collaborators.

> Once someone joins the Istio org (https://github.com/orgs/istio/people) they will be given access to modifying labels.

This is another new-ish github feature! Historically granting label write access also granted the ability to write commits to the repo (bypassing presubmits), which is obviously something inappropriate for a large open community.

Since labels were generally immutable we have tended to consider labels a safe place to "securely" store state. Perhaps this needs to change so we can better work with more flexible label policies? We have not had the chance to adapt to this github change yet though.

geeknoid:
I would like to proceed with this PR, the discussion about labels is not directly relevant to this PR.

We have a number of repos right now (~10) which aren't supported by prow/tide or are in an inconsistent state (like istio/api). I want to get things regularized ASAP and this PR is about getting  us on that road.

geeknoid:
BTW, on the label front:

A big reason why some members of the Istio community have found prow+tide
less charming is the enormous flow of emails that those systems induce. This is due to the fact
developers interact with PRs using /foo commands (which are GitHub comments
and GitHub sends emails for each comment entered), and due to the bot
updating comments as tests execute. The need to use /foo commands is driven
by the fact the users do not have write access to labels and hence cannot
use the normal GitHub UI to manipulate issues and PRs. Users do not have
write access to labels because the tools use labels to store authoritative
state such  as lgtm and CLA checks.

A solution to address these problems is to keep authoritative storage
outside of GitHub. For each issue/PR, have a DB record that holds
authoritative state independent of the GitHub UI. This is how other tools
that integrate with GitHub  do. CircleCI's state is kept in CircleCI,
Mergify's state is kept in Mergify, CodeCov state is kept in CodeCov. The
fact Prow and Tide use GitHub labels for storage is at the core of things I
believe.

The set of plugins I want to have enabled across the repos are those that
do not involve /foo commands and those that don't require state to be kept
in labels. It's OK to conveniently apply utility labels, but nothing that
drives critical part of our processes.

prow/plugins.yaml
```diff
@@ -72,112 +69,64 @@ approve:
...
+    - trigger
+    - wip
 
-  istio/operator:
-  - assign
-  - hold
-  - slackevents
-  - trigger
-  - wip
+  istio/istio:
```
howardjohn, line 79: We should either send a message to community or are the hold plugin back , this is used sometimes. Removing is fine with me though

prow/plugins.yaml
```diff
@@ -72,112 +69,64 @@ approve:
...
-  - approve
-  - assign
-  - blunderbuss
-  - golint
-  - help
-  - hold
-  - lgtm
-  - lifecycle
-  - slackevents
-  - trigger
```
howardjohn: we can't rerun prow tests without this

geeknoid: This is only moved in this file, the functionality is not disabled for the repo.

howardjohn: Oops, I missed it since its defined in `istio` and not `istio/istio`.

howardjohn:
I think this change is fine. For istio/istio this hardly changes the plugins and for other repos most of them don't make sense since they are using mergify.

I do think we should change the tide query to `review:approved` rather than switch to mergify though

geeknoid:
Agree, the use of the the update tide query looks promising. Once the
plugins we run and branch protection are normalized, I'd like to experiment
with tide and this annotation in the istio.io repo. If it works well, then
I'll go through all the repos and update.

Oh, before switching everything over to Tide, I also want to wait until
support is there to do in-repo configuration of the jobs to run. This is
supposedly going to land RSN.

howardjohn:
> Oh, before switching everything over to Tide, I also want to wait until
> support is there to do in-repo configuration of the jobs to run. This is
> supposedly going to land RSN.

sgtm, that will make it much easier to add new jobs, etc. Somewhat related, I have been working on a higher level config for prow jobs since we make a lot of mistakes due to the complexity/repetition of our job specs, if you want to take a look: https://github.com/istio/test-infra/pull/1480. It will be less straightforward to run this if we do in-repo config (but still possible of course. I guess if we really needed to we could just package it in a dockerfile and not even make the other repos depend on this one)

prow/plugins.yaml
```diff
@@ -72,112 +69,64 @@ approve:
...
+    - approve
+    - assign
+    - blunderbuss
+    - golint
+    - lgtm
+    - lifecycle
+    - verify-owners
 
   istio/test-infra:
-  - approve
```
fejta: Will you please leave the indentation at its previous level? This should reduce the deltas and make it easier to review.

prow/plugins.yaml
fejta: The istio plugins do not apply to istio/istio. So this PR right does in fact turn off testing in most places.

Ideally we should configure everything at the org level to keep things simpler and less redundant.

fejta (commented):
> We have a number of repos right now (~10) which aren't supported by prow/tide or are in an inconsistent state (like istio/api). I want to get things regularized ASAP and this PR is about getting us on that road.

Can we limit this PR to solving those problems, rather than removing prow functionality where it already exists?

geeknoid:
@fejta I believe you've mentioned several times now that we should configure things at the org level for consistency. Not all of our repos at the same so I don't understand how to take your feedback.

I tried to move as much shared stuff to the org level in this PR. I didn't realize that repo-level settings completely override org-level settings, I thought it was additive. I'll fix that now.

As for indenting, my editor is doing this.

geeknoid:
PTAL.

This gives a common set of plugins to all Istio repos.

Repos that depend on tide keep that behavior intact.

Tide functionality is disabled on istio/api since that repo has been moved to Mergify last week and is currently in limbo mode since the repo is both  using mergify and tide.

prow/plugins.yaml
```diff
@@ -5,11 +5,16 @@
...
   - istio/istio
   - istio/operator
-  - istio/test-infra
+  - istio/pkg
   - istio/proxy
+  - istio/test-infra
+  - istio/tools
   - istio-releases/pipeline
   trusted_org: istio
 - repos:
```
fejta, line 20: Does the istio-ecosystem org trust members of the istio org? If so then move it to the above section.

prow/plugins.yaml
```diff
@@ -5,11 +5,16 @@
 triggers:
 - repos:
   - istio/api
```
fejta, line 7: Can we set this to:
```yaml
- repos:
  - istio
  - istio-releases
  - istio-ecosystem
```

This will do what we want:
* Nothing in repos without prow jobs
* Immediately start running prow presubmits and postsubmits when we define them

prow/plugins.yaml
fejta: I would like to keep non-noisy, non-conflicting plugins like `assign`, `help`, `hold`, etc at least enabled where they are already.

People who like they can continue using them. People who do not do not need to bother with them.

If this is controversial can we move disabling them out of this PR so we can check in what we have in agreement?

cjwagner:
> The istio plugins do not apply to istio/istio. So this PR right does in fact turn off testing in most places.

This isn't true, the repo and org levels are additive like Martin described. Sounds like the confusion was that the plugin config for individual plugins is not additive (e.g. `lgtm` config defined for `istio/istio` overrides `lgtm` config for `istio`).

prow/plugins.yaml
```diff
@@ -70,82 +71,39 @@ approve:
...
-  - approve
-  - assign
-  - blunderbuss
-  - golint
-  - help
-  - hold
-  - lgtm
-  - lifecycle
-  - slackevents
+  istio:
```
fejta, line 74: Do we want to similarly configure `istio-releases` and `istio-ecosystem`?

fejta (commented):
Yeah, my confusion. Thanks for clarifying Cole.

prow/plugins.yaml
geeknoid: istio-ecosystem is completely unrelated to istio, so I don't think  there's any inherent trust there.

prow/plugins.yaml
geeknoid: I'm fine consolidating the istio repos and only list the orgs. I don't know anything about istio-releases and istio-ecosystem, so I don't want to mess with that configuration state.

PR #1469, opened 2019-07-12
Update branch protection again.

geeknoid:
- Include the tools repo.

- Ask for approval from CODEOWNERS

- Sort repos alphabetically

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -66,12 +66,68 @@ branch-protection:
         - cla/google
       required_pull_request_reviews:
         required_approving_review_count: 1
+        require_code_owner_reviews: true
       restrictions:
       repos:
         api:
           protect: false
           branches:
             <<: *blocked_branches
+            master:
+              protect: true
+        cni:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            master:
+              protect: true
+        common-files:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            master:
+              protect: true
+        installer:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            master:
+              protect: true
+        istio:
+          protect: false
+          branches:
+            <<: *blocked_branches
+            collab-galley:
+              protect: true
+              restrictions:
+                teams:
+                  - istio-hackers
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -66,6 +66,7 @@ branch-protection:
         - cla/google
       required_pull_request_reviews:
         required_approving_review_count: 1
+        require_code_owner_reviews: true
```
fejta, line 121: This seems like the only real change in here? It will be a lot easier to understand this change later if its the only thing in it.

Can you send the sorting PR separately?

prow/config.yaml
```diff
@@ -74,37 +75,19 @@ branch-protection:
             <<: *blocked_branches
             master:
               protect: true
-        istio.io:
-          protect: false
-          branches:
-            <<: *blocked_branches
```
fejta, line 80: I missed/forgot to note this in the other PR: the branch protector requires branches to be listed explicitly, `release-*` will not work, you need an entry for `release-1.0` etc. 

I don't see anything in the https://github.com/kubernetes/test-infra/tree/master/prow/cmd/branchprotector git history to suggest someone added this support

fejta (commented):
We can eliminate most of this structure by just defaulting the org to enable protection (all repos, all branches). It is simpler to define, easier to understand and safer.

Then we can focus on the special cases like the proxy repo.

This is how we set thing up in kubernetes, which has worked well.

https://github.com/kubernetes/test-infra/blob/3ce030d42897b8957c273eb93c17a66d8c72862c/prow/config.yaml#L63-L314

I don't understand the need for istio to reinvent the wheel here?

fejta:
https://github.com/istio/test-infra/pull/1470 is an example of what this looks like

geeknoid:
I'm surprised that the branch protector doesn't support release-*, since the branch protection feature in GitHub is specifically designed for pattern matching like that. I'll expand the branches to deal with this.

prow/config.yaml
geeknoid: I also added the tools repo as mentioned in the PR description.

PR #1470, opened 2019-07-12
Update branch protection

fejta:
* Mergify repos should have at least one approval from code owners
* Tide repos should not use code owners (yet)
* Require two approvals for api
* Outline how to restrict merges to release teams (do these teams exist yet?)
* Unit tests to cover each scenario.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -71,6 +71,10 @@ blocked_branches: &blocked_branches
   <<: *blocked_branches_base
   release-1.1: *blocking_merge
   release-1.2: *blocking_merge
+  #  <<: *blocking_merge
+  #  restrictions:
+  #    teams:
+  #    - release-managers-1-2
 
 branch-protection:
   allow_disabled_policies: true
@@ -80,11 +84,14 @@ branch-protection:
       required_status_checks:
         contexts:
         - cla/google
-      required_pull_request_reviews:
+      required_pull_request_reviews: &mergify_reviews
         required_approving_review_count: 1
+        require_code_owner_reviews: true
       restrictions:
       repos:
         api:
+          required_pull_request_reviews:
+            required_approving_review_count: 2
           branches: *blocked_branches
         istio.io:
           branches: *blocked_branches
@@ -99,6 +106,8 @@ branch-protection:
         common-files:
           branches: *blocked_branches
         istio:
+          required_pull_request_reviews: &tide_reviews
+            require_code_owner_reviews: false
           branches:
             <<: *blocked_branches_base
             collab-galley:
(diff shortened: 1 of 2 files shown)


prow/config.yaml
```diff
@@ -50,81 +50,22 @@ github_reporter:
...
-blocked_branches: &blocked_branches
-  release-*: &blocking_merge
-    required_status_checks:
-      contexts:
-      - "merges-blocked-needs-admin"
-    protect: true
-
 branch-protection:
-  allow_disabled_policies: true
+  protect: true
```
geeknoid: This will break how our community uses branches. There are non-standard branches in  our repos which should not have branch protection enabled. We only want master and release-* to have branch protection.

fejta: 1) What does this help? What does it accomplish that they cannot do with a fork?
2) If these are useful why shouldn't they have branch protection?
3) If they are small in number, we can specifically exclude them (like we do in kubernetes)?

geeknoid:
The branches are created in an ad hoc fashion. For example, in istio.io you
get branches created automatically when you select "Edit this Page" on the
main Istio web site.

Folks create branches against a variety of repos. They could indeed achieve
the same effect with a fork, and we encourage them to do so.  But the fact
is that manh of our developers create branches on our main repos. I
regularly go through the repos and delete old merged branches to keep
things clean.

I think  turning on branch protection for custom branches will break our
existing workflows.

fejta:
1) We have the ability to shape and influence social norms in addition to automated behavior.
2) We can automate the behavior we want to encode
3) This sounds like that might be a good reason to special case the istio.io repo to allow unprotected repos by default

howardjohn:
Regarding 3, I just did the "edit through the ui" option and it automagically created a fork for me (see https://github.com/howardjohn/mergify-test/pull/13)

fejta:
The revert button does the same thing unfortunately

fejta:
A branch-cleanup script would be interesting/useful -- aka delete any branch that does/doesn't match a list and hasn't been updated in some time-period.

geeknoid:
So given that we want consistent branch protection for all our "master" and "release-X.Y" branches across repos, but specifically want no protection for other branches in those same repos, how should we proceed?

geeknoid:
Do we still need this PR?

fejta:
/hold cancel
@geeknoid PTAL this configures branch protection to most of what we talked about (require code owners except where tide is used), 2 approvals for api

geeknoid:
This looks good.

Thanks.

geeknoid (approved):

PR #1472, opened 2019-07-13
Configure tide at the org level, add functionality

fejta:
/assign @geeknoid @howardjohn 
/hold

Here is how I would configure tide, which will work for all repos

This which will allow:
* A consistent and automatically-configured merge experience across repos in all orgs.
  - Quickly and safely retests presubmits against HEAD before merge
  - No write permissions required (but alllowed)
* A label to allow critical issues to delay merging until its finished
* Labels to override the default squashing merge behavior

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -202,23 +202,10 @@ branch-protection:
 
 tide:
   queries:
-  # TODO(fejta): enable at the org level
-  - repos:
-    - istio-releases/pipeline
-    missingLabels:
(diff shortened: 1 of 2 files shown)


geeknoid:
Thanks, but as explained previously, the Istio community has decided to use Mergify. We have many  repos already using Mergify and we're trying to complete a transition to Mergify in the coming days.

We are open to discussing

geeknoid:
This  is quite out of date, should we close?

PR #1474, opened 2019-07-14
Turn off service account for istioctl test

howardjohn:
I am pretty sure this is not needed as the test runs fully locally now
with kind, so it doesnt need boskos or gcr. The istioctl test is already
failing (https://github.com/istio/istio/issues/15500) so this will
reduce risk if this doesn't work as expected. Once this works others can
be removed as well

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -925,8 +925,6 @@ postsubmits:
         testing: test-pool
   - name: integ-istioctl-k8s-postsubmit-tests-master
     <<: *job_template
-    labels:
-      preset-service-account: "true"
     spec:
       containers:
       - <<: *istio_container


howardjohn:
/assign @geeknoid 

This label is giving access to secrets to access GCR, boskos, etc. This isn't needed now because we are using kind so everything is local

fejta:
Does this use pod-utils? If not then it likely won't upload logs. If so then hopefully you are good to go!

howardjohn:
podutils means `decorate: true` right? If so then yes, but we have tons of other jobs without this and they seem fine? I'll keep an eye on this tests, I accidentally changed the post submit one not presubmit so have to wait for a commit to merge to see if this still works

howardjohn:
Nevermind I read that wrong. Yes we use pod utils so yes this shold work and seems to: https://prow.istio.io/view/gcs/istio-prow/logs/integ-istioctl-k8s-postsubmit-tests-master/313 (not the failure is expected)

fejta:
Right, with pod-utils enabled then uploading to GCS happens in other containers that we inject into the pod, and we give them the necessary credentials to upload. The test container does not need credentials for this purpose.

Jobs that do not specify `decorate: true` means the test container is responsible for uploading job results to GCS (probably using the deprecated bootstrap library) and will need the credentials required to do this. If you can, upgrading to pod-utils is a better option.

Issue #1471, opened 2019-07-13
Prow job config should be generated from a simpler config


howardjohn:
from slack: 

would it be a bad idea to have a job config generation for the prow jobs? Is anyone else doing this? It seems like most of our jobs could be greatly simplified with a config like
  ```- name: integ-istioctl-k8s-presubmit-tests
    command: [prow/integ-suite-kind.sh, test.integration.istioctl.kube.presubmit]
    requirements: [kind]```
Rather than ~20 lines of config. It could be something very opinionated, such as adding things like `if requirements == boskos, set max_concurrency=5`. Some benefits to this:
• All options are standardized. We won't have people accidentally configuring tests slightly different in ways they don't understand. I think there are a number of tests that are misconfigured (extra permissions they don't need, always_run = false when they didn't mean it, inconsistent naming, etc)
• We also have a ton of duplication across branches that could be simplified
• We can get rid of anchors which is not recommended
• Single place to config something. Instead of having the builder image across 50 different jobs we can configure it in one place. Instead of specifying details like resource requests for each job, we could have a couple profiles maybe (default, highmem, highcpu, ...)

howardjohn:
I made a PoC for this, the config is 137 lines and generates the same exact config as the 1224 line yaml we have for istio/istio master. I also found at least 4 bugs in the original config too which is more justification for this.

TODO is determine how to hande branches and post submits with minimal repetition

howardjohn:
Done, see prow/config/jobs

PR #1475, opened 2019-07-14
Add racetest to testgrid

howardjohn:
This is our first test using testgrid annotation. If this works we can
add other tests using this method.

@fejta does this work? Or is there some other setup we need to do to process these?

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.yaml
@@ -84,6 +84,8 @@ presubmits:
         - prow/racetest.sh
       nodeSelector:
         testing: test-pool
+    annotations:
+      testgrid-dashboards: istio-presubmits-master
 
   - name: istio-shellcheck-master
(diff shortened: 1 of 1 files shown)


fejta:
Sadly testgrid annotations are not going to work outside of jobs defined in kubernetes/test-infra right now

howardjohn:
Bummer. What is our best option for this? Its not feasible to manually updating `testgrid/config.yaml`, we make mistakes and forget to add tests there all the time. I was hoping to just set this annotation via https://github.com/istio/test-infra/issues/1471.

fejta:
I would make some autobump logic, so:
* Somehow figure out the canonical set of istio testgroups and dashboards we want, and then output this into one or more `istio-*.yaml` files
* Move all the istio dashboards/test groups into their own files in kubernetes/test-infra
* Have a postsubmit prow job compute these files and then diff them against what is in kubernetes/test-infra
* Send a PR (or update the existing one) whenever they differ

Issue #1460, opened 2019-07-11
Remove custom context from prow jobs


howardjohn:
Right now we have configs like

```
  - name: integ-framework-local-presubmit-tests-master
    context: prow/integ-framework-local-presubmit-tests.sh
```

Not sure what purpose this has but it has caused confusion. Especially for the ones like above -- ` prow/integ-framework-local-presubmit-tests.sh` isn't even a real file!

If we just remove the context it will default to the job name which I think makes sense.

WDYT @utka @icygalz

howardjohn:
@cjwagner is this safe to do once we have https://github.com/kubernetes/test-infra/tree/master/prow/cmd/status-reconciler#status-reconciler deployed?

cjwagner:
Yes, and +1 for just using the job name as the context. That is what we do for prow.k8s.io.

howardjohn:
Thanks! Can you let me know what work would need to be done to have a status-reconciler deployment? Or is that part of the prow upgrades you guys have been working on?

cjwagner:
> Thanks! Can you let me know what work would need to be done to have a status-reconciler deployment? Or is that part of the prow upgrades you guys have been working on?

It is on the list of things to do. https://github.com/istio/test-infra/issues/1396
I don't think there are any blockers to deploying it now. Someone just has to add a deployment file like [this one](https://github.com/kubernetes/test-infra/blob/master/prow/cluster/statusreconciler_deployment.yaml) with the arguments tweaked for Istio and update the [Makefile](https://github.com/kubernetes/test-infra/blob/master/prow/Makefile).

clarketm:
status-reconciler deployed in: #1481

howardjohn:
This is 90% done, just a few repos left

howardjohn:
Done for everything

PR #1480, opened 2019-07-15
Generate prow config from simpler config file

### LINKED ISSUE #1471: Prow job config should be generated from a simpler config

howardjohn:
Fixes https://github.com/istio/test-infra/issues/1471

I may have messed up the vendor code, it got a little funky pulling in the new test-infra repo. I will double check this part.

This PR adds the code to do the transformation and creates a file using the new job syntax for istio/istio master. The generated config is virtually identical, with the following differences:

(with context field ignored - will be removed soon in https://github.com/istio/test-infra/issues/1460):
```
Presubmit diff:

Diff for release-test-master
JobBase.MaxConcurrency: 0 != 5


Postsubmit diff:

Diff for e2e-mixer-no_auth-master
JobBase.MaxConcurrency: 0 != 5

Diff for istio-pilot-e2e-envoyv2-v1alpha3-master
JobBase.MaxConcurrency: 0 != 5

Diff for e2e-dashboard-master
JobBase.MaxConcurrency: 0 != 5

Diff for e2e-bookInfoTests-envoyv2-v1alpha3-master
JobBase.MaxConcurrency: 0 != 5

Diff for e2e-simpleTests-master
JobBase.MaxConcurrency: 0 != 5

Diff for istio-kind-simpleTest-master
JobBase.Spec.Containers[0].Resources.Requests["memory"].i.value: 2147483648 != 3221225472
JobBase.Spec.Containers[0].Resources.Requests["memory"].s: "2Gi" != "3Gi"

Diff for istio-pilot-e2e-envoyv2-v1alpha3-k8s-latest-master
JobBase.MaxConcurrency: 0 != 5
```

So basically just some post submit tests are missing a maxConcurrency field the current config

In the future I will add a presubmit gate to ensure the config is generated (and check in the new generated istio/istio config - it is semantically the same but syntactical differences)


howardjohn:
The vendor update is to pull in an updated kubernetes/test-infra. I'll break that into a different PR once this is ready to merge

howardjohn:
@chases2

howardjohn:
@utka can you take a look?

utka:
The tests are still failing

howardjohn:
@utka I'll make a separate PR to update the vendor before checking this is, I was hoping you could look at the new config before then

utka:
Sorry, which config? there is more then 2k files in this PR :)

howardjohn:
@utka I created https://github.com/istio/test-infra/pull/1497 to split the change to what actually needs review, please take a look at that. Should be 3 files only

PR #1482, opened 2019-07-16
Make cjwagner a root level OWNERS approver.

cjwagner:
/assign @fejta 
/cc @howardjohn


fejta (commented):
/approve
but would like someone else to weigh in here

PR #1486, opened 2019-07-16
Restore plugins file to undo the effect of my previous accidental push.

geeknoid:


fejta (approved):
If that was because of an accidental push you may want to do something like `git remote set-url --push upstream no-push` or whatever.

Also, assuming nothing is broken we can just continue discussing here?
/hold
`/hold cancel` if you prefer to go back to the previous state.

Or if you prefer (as in less work for you) we start just close this and I can open a PR with additional changes I think are a good idea?

geeknoid:
Yeah, I'll set the no-push option.

If you're happy to iterate relative to the current state, then let's close this PR and go from there.

geeknoid:
OK, so we'll keep the accidental commit and work on additional PRs to further refine the setup.

Thus, closing.

PR #1487, opened 2019-07-16
Enable size, trigger and wip across all istio orgs

fejta:
/assign @geeknoid @cjwagner @howardjohn @utka 

Configure a base set plugins the same way throughout istio, istio-releases, istio-ecosystem

* size auto-manages the `size/L` or whatever label to PR, which helps gubernator.k8s.io/pr sorting
* wip auto-manages the work-in-progress label based on draft status and/or title
* trigger causes any presubmit/postsubmit prow jobs defined for the repo to run as configured
  - does nothing until these jobs are added

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -5,10 +5,9 @@
 triggers:
 - repos:
   - istio
-  - istio-releases/pipeline
+  - istio-ecosystem
+  - istio-releases
   trusted_org: istio
-- repos:
-  - istio-ecosystem/authservice
 
 config_updater:
   maps:
@@ -66,6 +65,16 @@ plugins:
   - trigger
   - wip
 
+  istio-ecosystem:
+  - size
+  - trigger
+  - wip
+
+  istio-releases:
+  - size
+  - trigger
+  - wip
+
   istio/istio:
   - approve
   - assign
@@ -97,7 +106,6 @@ plugins:
 
   istio-releases/pipeline:
   - lifecycle
-  - trigger
   - hold
 
   istio-ecosystem/authservice:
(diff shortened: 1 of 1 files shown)


prow/plugins.yaml
```diff
@@ -5,10 +5,9 @@
 triggers:
 - repos:
   - istio
-  - istio-releases/pipeline
+  - istio-ecosystem
+  - istio-releases
   trusted_org: istio
-- repos:
-  - istio-ecosystem/authservice
```
cjwagner, line 11: Note that this is a breaking change: Previously user trust in `istio-ecosystem/authservice` was based on membership in `istio-ecosystem`. Now it will be base on membership in `istio`.

fejta: This doesn't remove the istio-ecosystem trust. It means we trust anyone in the istio-ecosystem OR istio org.

https://github.com/kubernetes/test-infra/blob/2c53c2d7e3e07a95005d04ca7e36dd53c438582e/prow/plugins.yaml#L5-L13

fejta: And more specifically: https://github.com/kubernetes/test-infra/blob/master/prow/plugins/trigger/trigger.go#L168-L178

geeknoid:
I don't know anything about istio-releases and istio-ecosystem, how they're used, etc. So I don't know if this appropriate/desirable or not. Perhaps @rshriram can comment on istio-ecosystem.

fejta:
Plan to merge this tomorrow unless I hear a need to do otherwise (any concerns Shriram/others?)

It is quick to revert when/if it causes problems too.

PR #1489, opened 2019-07-16
Add plugins back to istio, proxy and test-infra

fejta:
/assign @utka @cjwagner @geeknoid @howardjohn 

Add back plugins removed in https://github.com/istio/test-infra/pull/1468 

Not necessary if we can reach agreement on https://github.com/istio/test-infra/pull/1490 (which I would personally prefer)

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -70,16 +70,24 @@ plugins:
   - approve
   - assign
   - blunderbuss
+  - golint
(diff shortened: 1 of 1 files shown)


fejta:
Can we get this one in? Sounds like #1490 is unlikely to get in anytime soon.

cjwagner (approved):
/hold

PR #1490, opened 2019-07-16
Enable opt-in plugins everywhere

fejta:
These plugins work when people choose to use them, do not otherwise comment and do not conflict with any alternative workflows.

We should allow people who like these workflows to continue using them.

https://github.com/istio/test-infra/pull/1489 is similar but more focused

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -61,39 +61,35 @@ approve:
   implicit_self_approve: true
 
 plugins:
-  istio:
-  - size
-  - trigger
-  - wip
+  istio: # TODO(fejta): make this the same for other orgs
+  - assign # manage assignees/reviewers via email
+  # blunderbuss TODO(fejta): teach blunderbuss to read CODEOWNERS, or is this just unnecssary?
+  - config-updater # auto-update requested configmaps
+  - golint # inline golint auto-review
+  - hold # manual hold command
+  - lifecycle # close or mark issues as stale via email
+  - override # admins can force a context to pass
+  - size # search for PRs with a small label to review quickly
+  - skip # cleans up non-blocking statuses from old commits
+  - slackevents # post requested messages to slack, if any
+  - trigger # run tests
+  - verify-owners # ensures any OWNERS changes are valid
+  - wip # wip label based on title/draft status
 
   istio/istio:
-  - approve
-  - assign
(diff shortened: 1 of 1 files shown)


geeknoid:
Within the Istio community, we've always has challenged around the volume of email generated by GitHub and the amount of stuff injected into PRs by automation making it hard to separate wheat from chaff.

It has therefore been a specific goal of ours to completely move away from /foo commands from our workflow. We don't want to make these commands optional, we simply do not want our community to interact with PRs and issues in this way.

As such, I do not want this PR to be merged.

Thanks.

fejta:
Minimizing the email volume seems like a good and understand goal.

Trying to ban chatops seems like a blunt and ineffective way to make progress against it. Will anything help you soften the anti-chatops stance (while holding onto your desire to reduce email)?

geeknoid:
If there is value in the chatbot, then we can use it. IMO, anything that is
providing an alternate to the canonical GitHub workflow, or is completely
superfluous (like yuks and shrugs) are simply not compelling reasons to
allow this functionality and the ensuing emails & PR obfuscation.

The chatbot makes sense when using a constrained permission model for
GitHub as in Kubernetes, but in the Istio world where users can manipulate
labels and other metadata using established mechanisms, enabling the
chatbot doesn't seem like a desirable goal. In the Istio context, it
doesn't add new functionality and introduces spam.

fejta:
Any subset of these we can convince you of? 

slackevents and config-updater, for example, are not related to `/commands` in github.

geeknoid:
What's the effect/benefit of enabling config-updater on repos where it's not in use?

Similarly, we don't generally post commits to Slack, so what's the benefit of enabling slackevents on all repos?

fejta:
My thinking is it makes it easier for people to decide to try it out and doesn't otherwise do anything.

AKA I suspect someone who thinks wants a merge warning for their repo to be willing to add an item to https://github.com/istio/test-infra/blob/52952762499b8302adcd7ca97b35054b524fad13/prow/plugins.yaml#L27-L34 but potentially balk if this also involves convincing someone to also turn it on for that repo (do I have permission to do that? Maybe its disabled on my repo for a reason and I'm not senior enough to override that decision, etc).

It's also less lines of config.

None of those are critical though right now, so fine just leaving things turned on where we want to use them if that's your preferred strategy.

Figured I would at least ask about those two as they aren't slash commands :grinning:

Issue #1452, opened 2019-07-09
Prow maybe dropping webhooks


fejta:
https://github.com/istio/test-infra/issues/1450 looks like prow didn't process an `/assign` command, which usually means it didn't get a webhook.

Could be a github problem, or could be a misconfigured deployment. 

Will see if we can get the ability to view webhook logs for prow.

fejta:
I'm struggling to find logs for prow...

https://pantheon.corp.google.com/logs/viewer?project=istio-testing&minLogLevel=0&expandAll=false&timestamp=2019-07-09T23:50:06.412000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2019-07-09T17:50:06.662Z&dateRangeEnd=2019-07-09T23:50:06.662Z&interval=PT6H&resource=container%2Fcluster_name%2Fprow%2Fnamespace_id%2Fdefault

fejta:
Turning on legacy stackdriver monitoring:

Update logging and monitoring services for Kubernetes Engine cluster "prow"
Just now
istio-testing

fejta:
/reopen
cc @Katharine

@geeknoid should we try and investigate further here? https://github.com/istio/test-infra/pull/1491#pullrequestreview-263377110

We'll either need someone who can view the webhook history or else your help in looking through for concrete things that should have been delivered but weren't.

PR #1491, opened 2019-07-16
Protect master and release branches

fejta:
/assign @geeknoid @cjwagner 

Follow-up to https://github.com/istio/test-infra/pull/1469

Martin is there a reason we closed it? This should do the same thing as in that PR with less redundant code (if we want everything defaulting unprotected, then define it that way at the top level)

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -50,15 +50,31 @@ github_reporter:
   - presubmit
   - postsubmit
 
-blocked_branches: &blocked_branches
-  release-*: &blocking_merge
+blocked_branches: &blocked_branches_base
+  master:
+    protect: true
+  release-0.1: &blocking_merge
     required_status_checks:
       contexts:
       - "merges-blocked-needs-admin"
     protect: true
+  release-0.2: *blocking_merge
+  release-0.3: *blocking_merge
+  release-0.4: *blocking_merge
+  release-0.5: *blocking_merge
+  release-0.6: *blocking_merge
+  release-0.7: *blocking_merge
+  release-0.8: *blocking_merge
+  release-1.0: *blocking_merge
+
+blocked_branches: &blocked_branches
+  <<: *blocked_branches_base
+  release-1.1: *blocking_merge
+  release-1.2: *blocking_merge
 
 branch-protection:
   allow_disabled_policies: true
+  protect: false
   orgs:
     istio:
       required_status_checks:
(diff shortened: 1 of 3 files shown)


fejta:
Added unit tests

prow/config.yaml
```diff
@@ -50,15 +50,31 @@ github_reporter:
...
+  release-0.2: *blocking_merge
+  release-0.3: *blocking_merge
+  release-0.4: *blocking_merge
+  release-0.5: *blocking_merge
+  release-0.6: *blocking_merge
+  release-0.7: *blocking_merge
+  release-0.8: *blocking_merge
+  release-1.0: *blocking_merge
+
+blocked_branches: &blocked_branches
```
geeknoid, line 70: Why are these branches separate from the ones above? Does this have a semantic meaning?

prow/config.yaml
```diff
@@ -69,61 +85,21 @@ branch-protection:
...
-              protect: true
+          branches: *blocked_branches
         istio.io:
-          protect: false
-          branches:
-            <<: *blocked_branches
-            master:
-              protect: true
+          branches: *blocked_branches
         operator:
```
geeknoid, line 91: I  think the operator also needs the same blocked branches.

prow/config.yaml
```diff
@@ -50,15 +50,31 @@ github_reporter:
...
+  release-0.7: *blocking_merge
+  release-0.8: *blocking_merge
+  release-1.0: *blocking_merge
+
+blocked_branches: &blocked_branches
+  <<: *blocked_branches_base
+  release-1.1: *blocking_merge
+  release-1.2: *blocking_merge
 
 branch-protection:
```
geeknoid, line 75: We need to turn on "requires code owners review" for any istio repo other than istio/proxy and istio/istio

geeknoid: I presume the behavior when something like required_status_check is not stated for a branch is that whatever is in GitHub is left alone, correct?

prow/config.yaml
fejta: This is because istio/istio has its own complications for `release-1.1` and `release-1.2` checks

prow/config.yaml
fejta: See the `required_pull_request_reviews` clause defined at the `istio` org level.

All repos will inherit this unless they override it. So right now all istio repos will require the `cla/google` context 
and require 1 an approving review, unless they define otherwise.


Note that [right now](https://github.com/istio/istio/settings/branch_protection_rules/1843352) istio/istio and istio/proxy both requiring these checks on any branches with protection enabled, and this PR is not changing that (are you saying it should?)

prow/config.yaml
fejta: Done

prow/config.yaml
geeknoid: What I'm asking for is different  than required_pull_request_reviews, we need require_code_owner_reviews: true since we use GitHub's CODEOWNERS files to control who can approve.

fejta: Ack, wow those are confusingly similar settings.

Can we merge this as is -- it fixes the `release-*` branches -- and I put the `required_code_owner_reviews` changes in the next PR?

These controls exist but we basically don't use them in kubernetes so I don't have experience with them and am therefore less confident in their behavior:
* https://github.com/kubernetes/test-infra/blob/7aef384140833379d3e72f33bc0bb9d10770aefe/prow/config/branch_protection.go#L62-L71
* https://github.com/kubernetes/test-infra/blob/7aef384140833379d3e72f33bc0bb9d10770aefe/prow/config/branch_protection.go#L38-L39

Let me figure out how to express that to the tool and I'll send a PR

So long as we leave `required_code_owner_reviews` as `nil` the tool should not alter these settings where they already exist.

fejta:
@geeknoid can you try again with an approving review instead of the slash commands? and/or would you want to dig back through the logs to see whether we dropped that webhook?

howardjohn (approved):
possibly related to missing that: https://github.com/istio/test-infra/issues/1501 and https://github.com/istio/test-infra/issues/1500

fejta:
Reopening #1452 to see if we can get traction on this. We do not ever expect prow to drop webhooks

fejta:
hello

/test lint-check

world

fejta:
That comment worked :-/ 

I wonder what the deal is...

geeknoid:
I  don't know how this behaves at runtime, but the YAML appears invalid. Line 70 reuses the same key (blocked_branches)  which  my  editor is highlighting as invalid.

geeknoid:
Also, the operator repo doesn't have  branch protection for the release-1.1 branch. Perhaps it's related to the YAML issue from above?

fejta:
This is part of why I wrote the unit test to prove that it was working correctly. I find yaml anchors kind of not worth the effort (although probably necessary in this instance so long as we require random branches to remain unprotected).

So I'm pretty sure it is valid yaml -- prow parses it correctly and the concrete values match what we want thus far. Let me see if I can repo the release-1.1 branch issue.

fejta:
https://github.com/istio/test-infra/pull/1509 proves things are configured correctly (at least at the yaml level) as best as I can tell.

Any chance your editor has a yaml parsing bug?

cjwagner:
FYI you can see if the config is being parsed as you expect by checking the parsed config at https://prow.istio.io/config

geeknoid:
Quite possible my editor doesn't understand the anchor syntax.

Any idea why the operator repo doesn't have branch protection for its release-1.1 branch?

fejta:
Yes, https://github.com/istio/test-infra/issues/1511 I think

PR #1492, opened 2019-07-16
Read error

ChristinaLyu0710-zz:
Read paths to pr directory from Google Spreadsheet and examine the build-log.txt in the directories to find errors and warnings. Output a csv file that contains lines of error, number of prs that contains the error and all the pr directories.


howardjohn (commented):
this should have a readme describing when why and how to run this. Otherwise this will just rot like many other things in this repo

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,394 @@
...
+	defer rc.Close()
+
+	data, err := ioutil.ReadAll(rc)
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
+
+func (f *ErrorFinder) getContent(ctx context.Context, filePaths []string) (map[string]string, map[string][]string, map[string][]string, error) {
```
howardjohn: Why do we need all of this manual parsing when the tests already output a well formatted junit file?

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+
+	}
+}
+
+func main() {
+	var fileName string
+	flag.StringVar(&fileName, "OutputFileName", "", "A file name for output csv file with .csv at the end.")
+	var spreadsheetID string
+	flag.StringVar(&spreadsheetID, "SpreadsheetId", "", "Spreadsheet ID for the spreadsheet containing file path")
```
howardjohn: this seems odd to me... Where does the spreadsheet come from? If we have a spreadsheet that contains all of the failures then clearly we have some code that makes this. Why can't we just do this logic there?

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,394 @@
...
+	bucket := client.Bucket(f.bucketName)
+	buildFile := bucket.Object(prefix + "build-log.txt")
+
+	rc, err := buildFile.NewReader(ctx)
+	if err != nil {
+		return "", err
+	}
+	defer rc.Close()
+
+	data, err := ioutil.ReadAll(rc)
```
howardjohn: This might crash/OOM. We have some build-log.txts that are >4GB large

toolbox/errorfinder/main.go
ChristinaLyu0710-zz: I'm reading the build-log.txt files because I'm interested in errors that occur when running the commands that can be both fatal and not fatal that comes from the build and run commands but not from tests.

toolbox/errorfinder/main.go
ChristinaLyu0710-zz: The spreadsheet contains a list of prs from istio-prow not all of the failures.

howardjohn: Where did the spreadsheet come from though? I assume someone didn't make it by hand?

toolbox/errorfinder/main.go
ChristinaLyu0710-zz: Just fixed it.

toolbox/errorfinder/main.go
ChristinaLyu0710-zz: Maria shared with me. I don't think she made it by hand. https://docs.google.com/spreadsheets/d/1XK4Nbq7nwqwXMC7waMsEqhe_wOf1J1rqwbECrop9fQI/edit?ts=5d277e24#gid=0

howardjohn: Yes, so what I am saying is: Someone wrote code to go through all of our tests and find failures and create this spreadsheet. Now you are creating code to read the spreadsheet and then go through all of our tests and read all of the failures.. Why don't we just have one piece of code that does all of this?

Also @therealmitchconnors already has a bunch of code that parses and categorizes the errors, I would sync up with him to avoid all of this code duplication

utka: I asked @therealmitchconnors if he needs help with categorizing the errors and that's what she is doing here. Finding common errors, that later we can identified what is common failures.

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,405 @@
...
+			ind := strings.Index(processedString, "exit status")
+
+			if ind == -1 {
+				break
+			}
+			splitSlice := strings.SplitAfterN(processedString, "exit status", 2)
+			beforeExit := strings.Split(splitSlice[0], "\n")
+			afterExit := strings.Split(splitSlice[1], "\n")
+
+			var tenLinesBefore []string
```
utka: the name of the variable tenLInes, but it gets 20?

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,405 @@
...
+			}
+
+			var tenLinesAfter []string
+			if len(afterExit) > 20 {
+				tenLinesAfter = afterExit[:20]
+			} else {
+				tenLinesAfter = afterExit[:]
+			}
+
+			total := append(tenLinesBefore, tenLinesAfter...)
```
utka: So you are only looking at the lines around exit status?

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,405 @@
...
+	errorFinder, err := NewErrorFinder(client, bucketName)
+	if err != nil {
+		log.Fatalf("Unable to create new error finder: %v", err)
+	}
+
+	_, errorMap, warningMap, err := errorFinder.getContent(context, listOfPrs)
+	if err != nil {
+		fmt.Println(err)
+	}
+	writeCSV(errorMap, warningMap, fileName)
```
utka: How long usually getContent takes?

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,518 @@
...
+				newLineSplit := []string{}
+
+				for indexD, section := range lineSplit {
+
+					// Ignore empty strings
+					if strings.Compare(section, "") == 0 {
+						continue
+					}
+
+					// If word does not contain key words, generalize word to characters necessary.
```
utka, line 237: I don't understand what you are trying to do here.
why not to just 
newLineSplit = append(newLineSplit, section)
instead of lines 221 to 235.

utka, line 237: WHy do you want to look at every word instead of checking the line as whole?

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,518 @@
...
+					}
+					errorMap[line] = prSlice
+				}
+			}
+		}
+	}
+	return errorMap, warningMap, nil
+}
+
+// Read destination csv file to get previously generated errors and warnings.
```
utka, line 301: What is your csv format? 
Is your format: error type (error or warning),  error message, list of paths

utka (changes requested):

toolbox/errorfinder/main.go
ChristinaLyu0710-zz: I look at each word to see if I need to replace it with /s or /d. I want to keep the original error format but not leave it too detailed. For example for error:
Error: Fail to do A. Error: Fail to do B. Error: Fail to do C.
I will generalize the A, B, C to avoid making it three different errors.

ChristinaLyu0710-zz: I want to generalize the error lines. A lot of errors are the same errors but just with different file names that are specific to the content of the run. I am looking at lines in general to see if it is an error line or warning line. If it is I will generalize it to add to the slice.

toolbox/errorfinder/main.go
ChristinaLyu0710-zz: I have two lines in my csv file. One containing only the word "error" and one containing only "warning". The lines under "error" line are errors. Each line except the "error" and "warning" lines are composed of error/warning message, number of paths that contain the error/warning and the list of paths.

toolbox/errorfinder/main.go
```diff
@@ -0,0 +1,509 @@
...
+
+					// Ignore empty strings
+					if strings.Compare(section, "") == 0 {
+						continue
+					}
+
+					// If word does not contain key words, generalize word to characters necessary.
+					if !strings.Contains(strings.ToLower(section), "error") &&
+						!strings.Contains(strings.ToLower(section), "fail") && !strings.Contains(strings.ToLower(section), "warn") {
+						if indexD > 0 && strings.Compare(lineSplit[indexD-1], "container") == 0 {
```
utka: Can it be done with less ifs? Is it necessary to add \\s

ChristinaLyu0710-zz: I added the \s-es because I think it would make the error message easier to understand. I'll try with fewer ifs

PR #1493, opened 2019-07-17
Update Bazel to 0.28.0.

PiotrSikora:

### DIFF
--- a/docker/istio/shared/tools/install-bazel.sh
+++ b/docker/istio/shared/tools/install-bazel.sh
@@ -2,7 +2,7 @@
 
 set -eux
 
-BAZEL_VERSION='0.27.0'
+BAZEL_VERSION='0.28.0'
 BAZEL_BASE_URL='https://github.com/bazelbuild/bazel/releases/download'
 BAZEL_SH="bazel-${BAZEL_VERSION}-installer-linux-x86_64.sh"
 BAZEL_URL="${BAZEL_BASE_URL}/${BAZEL_VERSION}/${BAZEL_SH}"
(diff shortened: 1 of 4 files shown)


fejta (approved):
/hold 
What produces the images?
Do a /hold cancel when you're ready for it to merge

PiotrSikora:
No idea who/what produces the images. I believe it's a manual step, but I don't have permissions to push the images, so I don't have the details. cc @utka

utka:
/lgtm
I lost permissions to push images as well. I think it can be done by prow on call

PiotrSikora:
So we should cancel the hold and once it's merged we should get prow on-call to build & push the image? I that correct?

utka:
This is my understanding how prow on call works

Katharine:
I've built and pushed [the image](https://gcr.io/istio-testing/prowbazel:0.5.8).

fejta:
This isn't really a prow service. I suspect But we should have access to the `istio-testing`... and while I'm writing this looks like Katharine already did it (thanks!). I suspect she did this by running a command on her workstation (hence why it isn't prow).

Ideally we create postsubmits to build and push these images, like we do for kubernets/test-infra images. Then we can make sure prow is running those correctly. Examples of these jobs:

* Using GCB -- https://github.com/kubernetes/test-infra/blob/7aef384140833379d3e72f33bc0bb9d10770aefe/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L155-L182
* Using bazel -- https://github.com/kubernetes/test-infra/blob/7aef384140833379d3e72f33bc0bb9d10770aefe/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L55-L74

PR #1494, opened 2019-07-17
Update prowbazel in proxy/master and proxy/wasm to 0.5.8.

PiotrSikora:

### DIFF
--- a/docker/prowbazel/Makefile
+++ b/docker/prowbazel/Makefile
@@ -1,5 +1,5 @@
 PROJECT = istio-testing
-VERSION ?= 0.5.7
+VERSION ?= 0.5.8
 
 # Note: The build directory is the root of the istio/test-infra repository, not ./
 image:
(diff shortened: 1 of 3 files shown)


PiotrSikora:
Blocked on https://github.com/istio/test-infra/pull/1493.

docker/prowbazel/Makefile
```diff
@@ -1,5 +1,5 @@
 PROJECT = istio-testing
-VERSION ?= 0.5.7
+VERSION ?= 0.5.8
```
fejta: Why is this a patch update?

fejta (approved):
/hold

docker/prowbazel/Makefile
PiotrSikora: Why wouldn't it be? (genuine question) There doesn't seem to be any pattern for minor and patch updates, as far as I can tell. See: [docker/prowbazel/README.md](https://github.com/istio/test-infra/blob/master/docker/prowbazel/README.md).

Having said that, I've now noticed that this file should be updated in https://github.com/istio/test-infra/pull/1493, since image recipes are updated there and not this PR. Let me fix that.

Katharine:
This image now exists so this is mergeable, but I don't think I'll be able to unhold. Let's try it anyway…

/hold cancel

PR #1497, opened 2019-07-17
Create code to generate prow configs

howardjohn:
This will not work until vendor is updated. Detaching this PR since that change will be massive.

This adds:
* generate.go to generate prow config
* jobs/istio.yaml, the equivalent to istio.istio.master.yaml in the new config format
* README.md describing the process


utka:
I don't mind the change, but is it suppose to help with complexity of prow configs? It's not looking simpler then original config %)

howardjohn:
The current job config is 1224 lines and has a lot of bugs in it due to misconfigurations. The new config is only 235 lines.

Current config (using anchors too, meaning if you want to change resource request you need to add ~6 more lines):
```yaml
  - name: integ-security-k8s-postsubmit-tests-master
    <<: *job_template
    labels:
    spec:
      containers:
      - <<: *istio_container
        command:
        - entrypoint
        - prow/integ-suite-kind.sh
        - test.integration.security.kube
        # Volumes needed to support KinD running in Kubernetes
        # See https://github.com/kubernetes-sigs/kind/issues/303
        volumeMounts:
        - mountPath: /lib/modules
          name: modules
          readOnly: true
        - mountPath: /sys/fs/cgroup
          name: cgroup
      volumes:
      - hostPath:
          path: /lib/modules
          type: Directory
        name: modules
      - hostPath:
          path: /sys/fs/cgroup
          type: Directory
        name: cgroup
      nodeSelector:
        testing: test-pool
```

New config:
```
  - name: integ-security-k8s-postsubmit-tests
    type: postsubmit
    command: [prow/integ-suite-kind.sh, test.integration.security.kube]
    requirements: [kind]
```

Personally I find it a lot simpler... that is the whole point of this

howardjohn:
Future plans are to generate test grid config from this as well

howardjohn:
This change was motived by making it harder to do the wrong thing: adding a presubmit without a postsubmit, marking the test as always_run: false or optional accidentally, forgetting to add to testgrid, forgetting to update the builder image, using root when not needed etc. And making it easier to do things like add support for kind, adding the same test to presubmit and postsubmit, easily see which tests are optional and not, etc

howardjohn:
@utka this is ready for review. 

Right now this contains:
* Readme describing it
* Code to generate
* Some test cases
* A config that will generate our current config for istio/istio master branch

I haven't yet actually checked in the generated config or added a presubmit check to enforce the generation. I will do that after this is merged

prow/config/jobs/istio.yaml
```diff
@@ -0,0 +1,257 @@
+repo: istio/istio
+branches:
+  - master
```
utka, line 3: Are you going to add new branches here? 
What will happen when we add more tests to master, but not to another branch?

howardjohn: I think we will probably need to fork the config, which isn't great. I couldn't think of any better ideas, do you have any?

howardjohn:
@utka can you take a look? I would like to either get this checked in or start working on making changes need to it

PR #1502, opened 2019-07-18
Modernize operator-test-master

fejta:
/assign @sdake @howardjohn @geeknoid @sbezverk 

Here's how I would recommend [converting istio/operator's circleci job](https://github.com/istio/operator/blob/1670827469a376fd755f8d7425dc333bc00d6a9c/.circleci/config.yml) to run on prow.

This job is clear, does everything it needs and should be approachable by general developers, aka it has an easy to understand image: (golang:1.12.7), a comprehensible command `go test -race ./...` and an environment variable (to force the use of go modules).

I would like to try and encourage this pattern rather than making CI testing appear to use and do things that are foreign to developers.

ref https://github.com/istio/test-infra/issues/1499

There's basically zero delta between what a developer will do normally, when running the presubmit locally and what the presubmit will do in prod.

```console
$ mkpj --job=operator-test-master --config-path=../istio-test-infra/prow/config.yaml --job-config-path ../istio-test-infra/prow/cluster/jobs --pull-number=91 | tee ~/job.yaml

$ phaino ~/job.yaml
?   	istio.io/operator/cmd	[no test files]
?   	istio.io/operator/cmd/iop	[no test files]
?   	istio.io/operator/cmd/manager	[no test files]
?   	istio.io/operator/pkg/apis	[no test files]
?   	istio.io/operator/pkg/apis/istio/v1alpha1	[no test files]
?   	istio.io/operator/pkg/apis/istio/v1alpha2	[no test files]
?   	istio.io/operator/pkg/component/component	[no test files]
ok  	istio.io/operator/pkg/component/controlplane	4.229s
?   	istio.io/operator/pkg/component/feature	[no test files]
?   	istio.io/operator/pkg/controller	[no test files]
?   	istio.io/operator/pkg/controller/istiocontrolplane	[no test files]
?   	istio.io/operator/pkg/helm	[no test files]
?   	istio.io/operator/pkg/helmreconciler	[no test files]
ok  	istio.io/operator/pkg/kubectlcmd	1.060s
ok  	istio.io/operator/pkg/manifest	1.280s
?   	istio.io/operator/pkg/name	[no test files]
ok  	istio.io/operator/pkg/patch	1.400s
ok  	istio.io/operator/pkg/translate	1.360s
ok  	istio.io/operator/pkg/util	1.205s
?   	istio.io/operator/pkg/util/fswatch	[no test files]
ok  	istio.io/operator/pkg/validate	1.242s
ok  	istio.io/operator/pkg/values	1.324s
[...]

PR #1506, opened 2019-07-18
Finish modernization of operator

sdake:
As per @fejta's PR: https://github.com/istio/test-infra/pull/1502

This moves the rest of the jobs to the model outlined by @fejta.  Please
use this structure when adding new jobs for the operator repo.  When Kind
becomes a necessity, please consult either @fejta or @sdake and we will
get you moving (or you can sort it out on your own if you prefer :)

### DIFF
--- a/prow/cluster/jobs/istio/operator/istio.operator.master.yaml
+++ b/prow/cluster/jobs/istio/operator/istio.operator.master.yaml
@@ -22,28 +22,32 @@ presubmits:
   istio/operator:
 
   - name: operator-lint-master
-    <<: *job_template
-    context: prow/lint.sh
+    path_alias: istio.io/operator
+    decorate: true
     always_run: true
     spec:
       containers:
-      - <<: *istio_container
+      - image: golang:1.12.7
         command:
-        - entrypoint
-        - prow/lint.sh
+        - make
+        - lint
       nodeSelector:
         testing: test-pool
 
   - name: operator-build-master
-    <<: *job_template
-    context: prow/build.sh
+    path_alias: istio.io/operator
+    decorate: true
     always_run: true
     spec:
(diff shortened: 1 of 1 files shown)


sdake:
/assign @fejta 

Erick,

If you would be kind enough to merge this after your PR I would appreciate it.  I'd like to remove the prow/ dir from istio/operator, and this makes that possible :)

Cheers
-steve

prow/cluster/jobs/istio/operator/istio.operator.master.yaml
```diff
@@ -22,28 +22,32 @@ presubmits:
   istio/operator:
 
   - name: operator-lint-master
-    <<: *job_template
```
howardjohn, line 25: you can remove this part now I think? the job_template definition above

prow/cluster/jobs/istio/operator/istio.operator.master.yaml
```diff
@@ -22,28 +22,32 @@ presubmits:
...
     always_run: true
     spec:
       containers:
-      - <<: *istio_container
+      - image: golang:1.12.7
         command:
-        - entrypoint
-        - prow/lint.sh
+        - make
+        - lint
```
howardjohn, line 33: Does this test need the lint tool preinstallled in the image?

ayj: `make lint` runs `scripts/run_golangci.sh` which downloads the linter binary each time. The previous prow based test worked the same way.

ayj (approved):
lgtm for operator

fejta (commented):
This already merged but thanks so much for doing this (and to the others for reviewing!)

PR #1507, opened 2019-07-18
Add manifest diff test for operator.

### LINKED ISSUE #15534: Create tests to compare helm output with iop
Tests are needed to compare the output from helm template with iop for a variety of parameter settings. There should be no diff between the outputs, which can be verified using a YAML diff tool.

morvencao:
Resolve: https://github.com/istio/istio/issues/15534

### DIFF
--- a/prow/cluster/jobs/istio/operator/istio.operator.master.yaml
+++ b/prow/cluster/jobs/istio/operator/istio.operator.master.yaml
@@ -73,15 +73,18 @@ presubmits:
         testing: test-pool
 
   - name: operator-mandiff-master
-    <<: *job_template
-    context: prow/mandiff.sh
+    path_alias: istio.io/operator
+    decorate: true
     always_run: true
     spec:
       containers:
-      - <<: *istio_container
+      - image: gcr.io/istio-testing/istio-builder:v20190624-478286dc
         command:
-        - entrypoint
-        - prow/mandiff.sh
+        - make
+        - mandiff
+        env:
+        - name: GO111MODULE
+          value: "on"
       nodeSelector:
         testing: test-pool
 postsubmits:
--- a/prow/cluster/jobs/istio/operator/istio.operator.master.yaml
+++ b/prow/cluster/jobs/istio/operator/istio.operator.master.yaml
@@ -71,6 +71,19 @@ presubmits:
           value: "on"
       nodeSelector:
         testing: test-pool
+
+  - name: operator-mandiff-master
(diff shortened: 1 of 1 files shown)


sdake:
@morvencao not trying to make your life difficult, but going to make it a little more difficult :)

The model laid out by @fejta is what I want to follow in the operator and install repos.

See the following PRs for exxamples:  https://github.com/istio/test-infra/pull/1506 https://github.com/istio/test-infra/pull/1502.

The key advantage is the gating code is located in one place rather than spread across different locations (plus the old yaml was very difficult to understand).

Cheers
-steve

morvencao:
@sdake Thanks for reminder. 

Have rebased the code changes. 
What I'm not sure is if adding prow script directly to the entrypoint of the prow job template is the recommended way to create new prow job.
Maybe there are other better way to do that after https://github.com/istio/test-infra/pull/1502
Any insight? @sdake @fejta

sdake:
@morvencao ideally all changes for CI should go to test-infra repo (for now).  I have deleted the prow dir in https://github.com/istio/operator/pull/102

I am struggling to find the implementation of mandiff.  My recommendation would be to create `make` targets for any test cases that should be run via prow and execute make as per: https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio/operator/istio.operator.master.yaml#L29-L33

morvencao:
@sdake Thanks for the advice, have added `mandiff` target in operator and  updated prow job to run `make mandiff`.

prow/cluster/jobs/istio/operator/istio.operator.master.yaml
```diff
@@ -71,6 +71,22 @@ presubmits:
...
+    path_alias: istio.io/operator
+    decorate: true
+    always_run: true
+    spec:
+      containers:
+      - image: gcr.io/istio-testing/istio-builder:v20190624-478286dc
+        command:
+        - make
+        - mandiff
+        env:
```
sdake, line 85: do you need GO111MODULE on?

sdake (approved):
looking more at the make path - it looks like we lack a make build target which is used under the hood in mandiff.  Will address that separatelhy, but for now merging.

/lgtm
/approve

prow/cluster/jobs/istio/operator/istio.operator.master.yaml
morvencao: Oh, yes. it's already in env section below.

PR #1509, opened 2019-07-18
More validation for operator branch protection

fejta:
/assign @geeknoid 

As best as I can tell release-1.1 is protected. It passes now, and if I add the `unprotected: true` to the test case, it fails as we'd expect:

```console

--- FAIL: TestConfig (0.38s)
    --- FAIL: TestConfig/operator_protects_release-0.1 (0.00s)
        config_test.go:163: protect: true (actual) but we expect false
FAIL
```

### DIFF
--- a/prow/config/config_test.go
+++ b/prow/config/config_test.go
@@ -94,6 +94,18 @@ func TestConfig(t *testing.T) {
 			branch: "master",
 		},
 		{
+			name:   "operator protects release-0.1",
+			org:    "istio",
+			repo:   "operator",
+			branch: "release-0.1",
+		},
+		{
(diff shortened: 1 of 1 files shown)


geeknoid:
Something funny is happening then, cause it's not protected according to GitHub.



Screen Shot 2019-07-18 at 11 39 21

fejta:
Per the attached log, it isn't getting to operator, I suspect due to https://github.com/istio/test-infra/issues/1511

[bp.txt](https://github.com/istio/test-infra/files/3408306/bp.txt)

fejta:
Do we want this?

howardjohn (approved):

PR #1510, opened 2019-07-18
Fix typo

sdake:


prow/cluster/jobs/istio/operator/istio.operator.master.yaml
```diff
@@ -64,7 +64,7 @@ presubmits:
         - -race
         - ./...
         env:
-        - name: GO111MODULE
```
fejta: This is actually correct, both this name and this job. You want to change the `operator-build-master` job above.

howardjohn (approved):

PR #1513, opened 2019-07-18
Use original iamge which came from istio C&P

sdake:


fejta (approved):

prow/cluster/jobs/istio/operator/istio.operator.master.yaml
```diff
@@ -60,7 +60,7 @@ presubmits:
     always_run: true
     spec:
       containers:
-      - image: golang:1.12.7
+      - image: gcr.io/istio-testing/istio-builder:v20190624-478286dc
```
fejta, line 63: Do we need to also change this one? It is already working, right?

sdake: curious to see what image we are using - it is clear that few if anyone are aware of what image we are using across the project.

PR #1514, opened 2019-07-19
Turn on elide_skipped_contexts

howardjohn:
Current any context across all branches will show up as "Skipped". This
change makes it so only tests that are run will be posted.

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -8,6 +8,7 @@ triggers:
   - istio-ecosystem
   - istio-releases
   trusted_org: istio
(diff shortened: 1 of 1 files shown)


prow/plugins.yaml
```diff
@@ -7,6 +7,7 @@ triggers:
   - istio
   - istio-releases/pipeline
   trusted_org: istio
+  elide_skipped_contexts: true
 - repos:
   - istio-ecosystem/authservice
```
cjwagner: Enable here as well?

cjwagner (approved):

PR #1527, opened 2019-07-23
Move istio/pkg from CircleCi+Mergify to Prow+Tide.

geeknoid:


prow/cluster/jobs/istio/pkg/OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
+  - geeknoid
+  - ozben
```
howardjohn: I don't think this is the right name?

geeknoid:
PS: how can I validate this stuff before submitting a PR and waiting for things to propagate?

prow/cluster/jobs/istio/pkg/OWNERS
geeknoid: Whoops, I used his Google alias rather than his GitHub alias.

prow/cluster/jobs/istio/pkg/istio.pkg.master.yaml
```diff
@@ -0,0 +1,79 @@
+job_template: &job_template
+  branches:
+    - "^master$"
+  decorate: true
+  path_alias: istio.io/pkg
+
+istio_container: &istio_container
```
fejta: I don't see any usage of `*istio_container` nor `*job_template`, I'd suggest removing this dead config.

prow/cluster/jobs/istio/pkg/istio.pkg.master.yaml
```diff
@@ -0,0 +1,79 @@
...
+              - -coverprofile=coverage.txt
+              - -covermode=atomic
+              - ./...
+            env:
+              - name: GO111MODULE
+                value: "on"
+        nodeSelector:
+          testing: test-pool
+
+postsubmits:
```
fejta: Remove?

fejta (approved):
/hold

cjwagner:
> PS: how can I validate this stuff before submitting a PR and waiting for things to propagate?

- The OWNERS file changes are validated by the `verify-owners` plugin.
- The config changes are validated by the `pull-test-infra-prow-checkconfig` presubmit job.
- To test the jobs themselves you can use [`phaino`](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/phaino#phaino) locally.
   - I'm working on making it possible to run ProwJob pods locally in a kind cluster with local output to the host instead of upload to GCS. I still need to do some polishing, but it is technically working now if you want to give it a try: https://github.com/cjwagner/test-infra/blob/local-pod/prow/cmd/mkpod/local-kind.sh You'd just need to change the `mkpj` command to point to istio's job config instead of k8s'.

prow/cluster/jobs/istio/pkg/istio.pkg.master.yaml
geeknoid: I just did  cut & paste from the operator's job definitions. I've removed this, and removed the similar stuff from the operator's config too.

prow/cluster/jobs/istio/pkg/istio.pkg.master.yaml
geeknoid: Removed, and removed in the operator's config too.

fejta:
Will `bazel test //...` also run these tests? Ideally they will

prow/cluster/jobs/istio/pkg/istio.pkg.master.yaml
```diff
@@ -0,0 +1,62 @@
+job_template: &job_template
```
fejta: I still don't see this being used? Consider deleting.

fejta (approved):
/hold

prow/cluster/jobs/istio/pkg/istio.pkg.master.yaml
geeknoid: OK, deleted...

fejta (approved):
`/hold cancel` at your leisure! (or just remove the label without the command)

PR #1529, opened 2019-07-23
Adding privileged back to postsubmit

utka:
Privileged gives several benefits for this particular test, like ability to create symlinks outside of build context. I was able to walk around this particular symlink issue, but the test still fails to copy data to the bucket, it fails with insufficient permissions. We did not change bucket permissions, so I want to test if privileged going to help with this or not.

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
@@ -115,6 +115,8 @@ postsubmits:
       - image: gcr.io/istio-testing/prowbazel:0.5.8
         command:
         - ./prow/proxy-postsubmit.sh
+        securityContext:
+          privileged: true
         resources:
           requests:
             memory: "8Gi"


fejta (commented):
Can you add some details about why?

utka:
Sorry, the issue with symlink warnings during build https://github.com/istio/istio/issues/15714

fejta:
I would like to either see something in the PR description or else a comment in the changed line.

utka:
Done. Updated description

fejta (commented):
> without privilege container the symlunk is not created

a) The description has sloppy grammar and spelling. Will you put in a bit more effort into this please?

b) This is not correct: docker and linux do not require privileged containers to create symlinks. Any chance you can dig into this issue a little more deeply to determine the root cause and a more robust fix?

utka:
Sure I can look more into it, but it was the most recent change before test started to fail.

utka:
I was able to fix symlink issues, but now I have problem with copying data to bucket, which i don't really understand, but I would like to test if privileged helps to solve that problem 
https://prow.istio.io/view/gcs/istio-prow/logs/proxy-postsubmit/630
WDYT? 

<blockquote><div><strong><a href="https://prow.istio.io/view/gcs/istio-prow/logs/proxy-postsubmit/630">proxy-postsubmit #630</a></strong></div></blockquote>

michelle192837 (approved):

michelle192837:
Have not found anything particular for a root cause, so willing to try this out for a bit to check. ^^

PR #1530, opened 2019-07-23
Add job to test that gcsweb is alive

Katharine:
This job just tries fetching gcsweb with TLS verification enabled, and fails if it can't. We can hook it up to testgrid to get emails when it fails, creating a basic but functional monitoring system.

gcsweb going down can cause end-user issues (e.g. istio/istio#15699) as well as upstream developer issues, so it is particularly worthwhile having monitoring on this.


fejta (approved):
/ok-to-test

PR #1532, opened 2019-07-24
Add required contexts for operator and pkg repos.

geeknoid:


prow/config.yaml
```diff
@@ -99,12 +99,25 @@ branch-protection:
...
         cni:
           branches: *blocked_branches
         installer:
           branches: *blocked_branches
         pkg:
           branches: *blocked_branches
+            master:
+              required_status_checks:
+                contexts:
+                  - pkg-lint-master
```
howardjohn, line 120: you shouldn't need these. By adding the config its already required. If you want to make a test not required you set `optional: true` on the job

fejta: Yeah right now this PR seems unnecessary. We only need to define these for foreign contexts that prow does not manage -- like CircleCI's or cla: yes.

Prow job contexts should not go in here as they are added automatically

PR #1535, opened 2019-07-24
Try to fix the test job.

geeknoid:


fejta:
PS: rlenglet, consider using github approve rather than `/lgtm` as the former will work in all repos (aka it will add lgtm and approve in this repo, will work in the new tide config in istio/operator and will work with mergify)

rlenglet:
@fejta Sure. I did that because @geeknoid asked explicitly on Slack: `Can someone give me an /lgtm...`. I did exactly what he asked.

fejta:
haha :) well that makes total sense. Just trying to spread awareness of the snazzy new prow functionality. Github approve now == `/lgtm` as far as prow is concerned

PR #1537, opened 2019-07-25
Enable tide for the istio/common-files repo.

geeknoid:


prow/plugins.yaml
```diff
@@ -126,6 +126,9 @@ external_plugins:
   - name: needs-rebase
     events:
       - pull_request
+  - name: cherrypicker
```
fejta: This will result in the `hook` deployment attempting to forward the PR webhooks it receives to a `cherrypicker` service in kubernetes, like it does for needs-rebase (L126).

We will need to deploy this service before enabling this external plugin:

https://github.com/istio/test-infra/blob/b619624f8725f9596431ad5cc8436eb5e11f5732/prow/cluster/needs-rebase_service.yaml#L16-L18

 Can we maybe put this in its own PR?

fejta (approved):
/hold

Everything aside from the external plugin LGTM

fejta:
@cjwagner confirms that -- aside from spamming logs with error messages -- nothing else bad should happen due to a misconfigured external plugin.

prow/plugins.yaml
geeknoid: OK, I removed the stuff to enable the cherrypicker.

Can you help getting  that setup please? Once this is enabled, I'll be able to move more repos to Tide.

fejta (approved):
At some point we should consider configuring tide with a blacklist of repos in the org to skip, rather than a whitelist

PR #1540, opened 2019-07-25
Switch review settings for operator, pkg, and common-files.

geeknoid:
Need to put the right constraints for tide-based repos.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -98,14 +98,23 @@ branch-protection:
         istio.io:
           branches: *blocked_branches
         operator:
+          required_pull_request_reviews: &tide_withcodeowners_reviews
+            require_code_owner_reviews: true
+          restrictions: &tide_withcodeowners_restrictions
+            teams:
+              - repo-admins
           branches: *blocked_branches
         cni:
           branches: *blocked_branches
         installer:
           branches: *blocked_branches
         pkg:
+          required_pull_request_reviews: *tide_withcodeowners_reviews
+          restrictions: *tide_withcodeowners_restrictions
           branches: *blocked_branches
         common-files:
+          required_pull_request_reviews: *tide_withcodeowners_reviews
+          restrictions: *tide_withcodeowners_restrictions
           branches: *blocked_branches
         istio:
           required_pull_request_reviews: &tide_reviews


prow/config.yaml
```diff
@@ -98,14 +98,29 @@ branch-protection:
...
         pkg:
+          required_pull_request_reviews: &tide_reviews
+            require_code_owner_reviews: false
+          restrictions: &tide_restrictions
+            teams:
+              - repo-admins
           branches: *blocked_branches
         common-files:
+          required_pull_request_reviews: &tide_reviews
+            require_code_owner_reviews: false
```
howardjohn: don't we want codeowners?

geeknoid: Yes of course. Cut & paste syndrome.

What I really wanted with this PR is just restricting merges to admins.

prow/config.yaml
```diff
@@ -98,14 +98,29 @@ branch-protection:
...
+              - repo-admins
           branches: *blocked_branches
         cni:
           branches: *blocked_branches
         installer:
           branches: *blocked_branches
         pkg:
+          required_pull_request_reviews: &tide_reviews
+            require_code_owner_reviews: true
+          restrictions: &tide_restrictions
```
fejta: The tide_restrictions anchor should only be defined once. Thereafter we only need to point to it: `restrictions: *tide_restrictions` (the teams part is already included in the `*`).

If you need to extend the anchor then you need the
```yaml
restrictions:
  <<: *tide_restrictions
  my-extensions: whatever
```
pattern, but you don't here, so just `restrictions: *tide_restrictions` is right

prow/config.yaml
```diff
@@ -98,14 +98,29 @@ branch-protection:
...
+            teams:
+              - repo-admins
           branches: *blocked_branches
         common-files:
+          required_pull_request_reviews: &tide_reviews
```
fejta: This anchor has already been defined on L106/L112, so this should just be `required_pull_request_reviews: *tide_reviews`

prow/config.yaml
```diff
@@ -98,14 +98,29 @@ branch-protection:
...
+            require_code_owner_reviews: true
+          restrictions: &tide_restrictions
+            teams:
+              - repo-admins
           branches: *blocked_branches
         istio:
           required_pull_request_reviews: &tide_reviews
```
fejta, line 120: This should now become `required_pull_request_reviews: *tide_reviews`, dropping L112/L127

Same thing for L113/L128: should become `restrictions: *tide_restrictions` and not redefine the anchor

fejta (commented):
LGTM intent, yaml anchor nits

geeknoid:
PTAL

fejta (approved):

PR #1553, opened 2019-07-29
Add annotations for testgrid

howardjohn:
Next step is checking the generated config into kubernetes/test-infra

### DIFF
--- a/testgrid/default.yaml
+++ b/testgrid/default.yaml
@@ -42,3 +42,5 @@ default_dashboard_tab:
 dashboards:
   - name: istio-presubmits-master
   - name: istio-postsubmits-master
+  - name: istio-presubmits-release-1.2
+  - name: istio-postsubmits-release-1.2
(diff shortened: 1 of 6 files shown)


testgrid/default.yaml
```diff
@@ -42,3 +42,7 @@ default_dashboard_tab:
 dashboards:
   - name: istio-presubmits-master
   - name: istio-postsubmits-master
+  - name: istio-presubmits-release-1.2
+  - name: istio-postsubmits-release-1.2
+  - name: istio-release
+  - name: istio-infrastructure
```
chases2, line 48: nit: trailing newline

testgrid/istio-generated.yaml
```diff
@@ -29,23 +29,715 @@ default_test_group:
   use_kubernetes_client: true
```
chases2, line 29: A new version of Configurator should be able to regenerate without the default settings; see https://github.com/kubernetes/test-infra/pull/13636

You can include the default settings, but they won't be respected; the defaults are set by instance of TestGrid

chases2 (approved):
These look good to me. A few comments for awareness

michelle192837 (approved):

PR #1561, opened 2019-07-30
Enable tide for istio/api.

geeknoid:


howardjohn (approved):

prow/config.yaml
```diff
@@ -243,6 +245,7 @@ tide:
     istio/istio: squash
     istio/operator: squash
     istio/pkg: squash
+    istio/api: squash
```
cjwagner, line 248: FYI, if all `istio` repos should use squash it is sufficient to specify `istio: squash` instead of enumerating each repo.

cjwagner (approved):
Feel free to `/hold cancel` 
/hold

prow/config.yaml
geeknoid: Yeah, we'll get there... :-)

PR #1563, opened 2019-07-30
Enable prow and tide for istio.io

geeknoid:


prow/cluster/jobs/istio/istio.io/istio.istio.io.master.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - name: presubmit-master
+      path_alias: istio.io/istio.io
+      decorate: true
+      always_run: true
+      spec:
+        containers:
+          - image: gcr.io/istio-testing/website-tools:2019-07-25
+            command:
+              - make
+              - prow
```
howardjohn: nit: is this something a dev would want to run locally? just wonder if it would make sense to call it `make test` or something since it isn't doing anything prow specific

prow/cluster/jobs/istio/istio.io/istio.istio.io.master.yaml
```diff
@@ -0,0 +1,16 @@
+presubmits:
+
+  istio/istio.io:
+
+    - name: presubmit-master
+      path_alias: istio.io/istio.io
+      decorate: true
+      always_run: true
```
howardjohn: I think you can add `branches: (release-1.2|master|release-1.3)` and consolidate, or if you want them isolate you need to add `branches: master` or something similar. Not 100% sure on this though

prow/cluster/jobs/istio/istio.io/istio.istio.io.master.yaml
geeknoid: It has to run from within the container. There are already targets for devs to use in that repo.

I plan on cleaning all of this kind of stuff up in the container-based  build world. Then the names of targets will be normalized.

prow/cluster/jobs/istio/istio.io/istio.istio.io.master.yaml
geeknoid: @cjwagner Can you clarify  what I should be using here?

And what does it mean when  there is no branches spec at all, such as is the case in the  yaml for the pkg and community repos? Does that imply "for all branches"? if so, then should  I rename the yaml files to remove "master" from them, or is the name of the files somehow significant?

prow/cluster/jobs/istio/istio.io/istio.istio.io.master.yaml
```diff
@@ -0,0 +1,16 @@
+presubmits:
+
+  istio/istio.io:
+
+    - name: presubmit-master
```
cjwagner: nit: I recommend including the repo in the job name. It is good convention to use globally unique job names so that jobs can be identified by name and so that naming schemes for presubmit/postsubmit/periodic variants of a job don't result in collisions.

prow/cluster/jobs/istio/istio.io/istio.istio.io.master.yaml
cjwagner: If there are no `branches` or `skip_branches` then the job runs on all branches. So as written, this PR adds 3 identical jobs with different names that all run on all branches.
Unless you need the job to run with a different name on different branches you can just keep one version of the job and let it run against all branches (or specify `branches: [master, release-1.2, release-1.3]` to just run on these branches).

The name of the file is not significant except that it needs to be unique from the other file names since the file name is used as a configmap key.

geeknoid:
I've cleaned up the jobs for the istio.io repo as suggested, and cleaned up some definitions/naming for operator, pkg, and community repos.

PTAL.

cjwagner (approved):

PR #1565, opened 2019-07-31
Add testgrid config for all tests, remove default config

howardjohn:
This is needed to get https://github.com/kubernetes/test-infra/pull/13699 passing tests, and also adds more stuff to the test grid in the process


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -41,6 +45,11 @@ periodics:
   name: test-infra-update-deps
   labels:
     preset-service-account: "true"
+  annotations:
+    testgrid-dashboards: istio-infrastructure
+    # Do not alert since these always fail
+    # testgrid-alert-email: istio-oncall@googlegroups.com
```
chases2, line 51: Testgrid alerting will send an email once, and it won't send another unless the test starts succeeding and then failing again.

Or, if the job is completely failing, comment out the entire prow job and don't run it.

howardjohn: Ah that's good to know. I just found out today it is always failing, I don't want to disable it until we decide it either needs to fixed or we can kill it - tracking in https://github.com/istio/istio/issues/15965

howardjohn:
/assign @utka 

This config is already checked into k8s/test-infra and live at https://testgrid.k8s.io/

PR #1568, opened 2019-08-01
Reducing cost by reducing number of resources used.

sebastienvas:
Also reducing number of threads for cleaner as cleaner is only used during resource deletion, and was to sending multiple requests per second.

Used https://velodrome.istio.io/dashboard/db/boskos-dashboard?orgId=1&from=now%2Fw&to=now%2Fw to come up with numbers

Note the on demand resource creation does not work just yet ( use min to guarantee you have enough resource), because we need to update the mason client to use request ids. Which I will do in after I am back from vacation.

/cc @utka


howardjohn (approved):
lgtm, as we move to kind this will only go down

sebastienvas:
I lost my power.

PR #1571, opened 2019-08-01
Update container API to allow creating GCP projects with IDNS

pitlv2109:
Manually tested the new API with my own project, should work™ with Mason. Should not have any risk since current Istio-owned Boskos projects do not concern with identity namespace.

Also update the vendor/deps by running:
 `GO111MODULE=on go get -u google.golang.org/api/container/v1beta1@latest` then
`./script/update_deps.sh`.

Too many vendor additions and deletions :cry:


pitlv2109:
Hey guys, this PR is a feature request from the Istio security team. PTAL.

/cc @quanjielin @myidpt

quanjielin (commented):
/lgtm

myidpt:
Thank you Phillip. Do we have to update the vendors?

pitlv2109:
> Thank you Phillip. Do we have to update the vendors?

I believe so. The repo didn't have the new container API so I `go get` it, then I just updated all the deps using an existing script.

PR #1576, opened 2019-08-02
Switch the bots repo to Tide+Prow.

geeknoid:
And enable Tide on the gogo-genproto repo.

### DIFF
--- a/prow/cluster/jobs/istio/bots/istio.bots.yaml
+++ b/prow/cluster/jobs/istio/bots/istio.bots.yaml
@@ -1,0 +1,54 @@
+presubmits:
+
+  istio/bots:
+
+    - name: bots-lint
+      path_alias: istio.io/bots
+      decorate: true
+      always_run: true
+      spec:
+        containers:
+          - image: gcr.io/istio-testing/circleci:2019-04-22
+            command:
+              - make
+              - lint
+            env:
+              - name: GO111MODULE
+                value: "on"
+        nodeSelector:
+          testing: test-pool
+
+    - name: bots-build
+      path_alias: istio.io/bots
+      decorate: true
+      always_run: true
+      spec:
+        containers:
+          - image: gcr.io/istio-testing/circleci:2019-04-22
+            command:
+              - go
+              - build
+              - ./...
+            env:
+              - name: GO111MODULE
+                value: "on"
+        nodeSelector:
+          testing: build-pool
+
+    - name: bots-test
+      path_alias: istio.io/bots
+      decorate: true
+      always_run: true
+      spec:
+        containers:
+          - image: gcr.io/istio-testing/circleci:2019-04-22
+            command:
+              - go
+              - -race
+              - test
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

prow/cluster/jobs/istio/bots/istio.bots.yaml
```diff
@@ -0,0 +1,54 @@
+presubmits:
+
+  istio/bots:
+
+    - name: bots-lint
+      path_alias: istio.io/bots
+      decorate: true
+      always_run: true
```
cjwagner, line 8: FYI, when configuring new jobs we often initially specify `skip_report: true` to prevent the jobs from actually reporting to GH. They still run automatically on PRs, but do so silently. This is a good way to get signal that the jobs are working as intended and once that is validated we remove the `skip_report: true`.

This isn't necessary, especially if there is high confidence that the jobs will WAI or a small number of developers impacted, but it is a handy pattern for trying out new jobs without risking negative impact on developers so I thought I'd mention it.

prow/cluster/jobs/istio/bots/istio.bots.yaml
```diff
@@ -0,0 +1,54 @@
...
+        nodeSelector:
+          testing: test-pool
+
+    - name: bots-build
+      path_alias: istio.io/bots
+      decorate: true
+      always_run: true
+      spec:
+        containers:
+          - image: gcr.io/istio-testing/circleci:2019-04-22
```
cjwagner, line 27: nit: Are these custom images necessary? Can you instead use the `golang` image?

prow/config.yaml
```diff
@@ -259,15 +267,7 @@ tide:
...
-    istio/istio: squash
-    istio/operator: squash
-    istio/pkg: squash
-    istio/api: squash
-    istio/common-files: squash
-    istio/community: squash
-    istio/istio.io: squash
-    istio/proxy: squash
-    istio/test-infra: squash
+    istio: squash
```
cjwagner, line 270: :+1:

cjwagner (approved):
Feel free to `/hold cancel`.

prow/cluster/jobs/istio/bots/istio.bots.yaml
geeknoid: Oh,  that's indeed handy. I think in the case we're fine since there are few developers affected and I'm pretty sure this will just work. But definitely for future changes.

prow/cluster/jobs/istio/bots/istio.bots.yaml
geeknoid: Right now, this is a lateral move from the existing CircleCi-based jobs. This image is used by other prow jobs too at this point.

We've embarked on a grand container image cleanup. We will likely be consolidating a lot of our images into a small number, used across all repos. This is starting to happen now, should be done in a few weeks I expect.

PR #1579, opened 2019-08-02
Adding service account to proxy postsubmit test

utka:


cjwagner:
`service-account` is the secret we use for uploading job results to `gs://istio-prow`. I know that this secret also has permission to upload to `gs://istio-build`, but that should be revoked. Ideally we should be using a separate service account for upload to `gs://istio-build` instead of sticking all permissions that might be needed into one service account.
This service account will work though...
/shrug

I think you'll also need to specify the `GOOGLE_APPLICATION_CREDENTIALS` env var. e.g. https://github.com/kubernetes/test-infra/blob/5de6cdf6fe7bb5277afdb0884afd8356154820a3/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L50-L59

utka:
GOOGLE_APPLICATION_CREDENTIALS already specified.

prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
```diff
@@ -124,5 +124,9 @@ postsubmits:
           limits:
             memory: "60Gi"
             cpu: "16000m"
+      volumes:
```
howardjohn, line 127: do we need a volumeMount?

howardjohn: Also doesn't `preset-service-account: "true"` do this already?

utka: I am trying to figure out how to make the test be happy :|

utka: https://github.com/istio/istio/issues/15714#issuecomment-517483833

howardjohn: We have plenty of other jobs that push to GCR though.. every e2e test in istio/istio does and just ` preset-service-account: "true"` is sufficient. The ` preset-service-account: "true"` preset looks like it already adds this volume actually?

utka:
Not necessary

PR #1582, opened 2019-08-02
add time comparer

ChristinaLyu0710-zz:
Read build-log.txt files according to paths given in spreadsheet. Find time commands and filter them out to csv output in sorted order based on the average run time.


toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,551 @@
...
+					}
+				}
+			}
+		}
+	}
+	return commandToTime
+}
+
+// Sort the time value array from low to high.
+func quicksort(timesArray []float64) []float64 {
```
utka: Use existing go library instead

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,551 @@
...
+	sysOutput := regexp.MustCompile(`^sys(\s)*([0-9])*m([0-9])*\.([0-9])*s`)
+	for _, filePath := range filePaths {
+		fileSlice, err := tc.query(ctx, filePath)
+		if err != nil {
+			continue
+		}
+		for start := 0; start < len(fileSlice); start++ {
+			line := fileSlice[start]
+			if timeCommand.MatchString(line) {
+				for end := start; end < len(fileSlice)-2; end++ {
```
utka: What about the nested time commands
time do_a
time_do_b
....
real for_b
real for_a

What about using stack?

utka (changes requested):

PR #1585, opened 2019-08-03
Add a test job for make lint using dind

sdake:
Going to self-approve on the weekend as part of our progression
to moving all builds to docker(in docker)(in K8s)(in docker). :)

### DIFF
--- a/prow/cluster/jobs/istio/operator/istio.operator.yaml
+++ b/prow/cluster/jobs/istio/operator/istio.operator.yaml
@@ -2,6 +2,45 @@ presubmits:
 
   istio/operator:
 
+  - name: operator-lint-dind
+    path_alias: istio.io/operator
+    decorate: true
+    always_run: true
+    optional: true
+    spec:
+      containers:
+      - image: docker.io/sdake/prow-launcher:2019-08-03
+        command:
+        - entrypoint
+        - make
+        - lint
+        resources:
+          limits:
+            cpu: "5"
+            memory: 24Gi
+          requests:
+            cpu: "3"
+            memory: 3Gi
+        securityContext:
+          privileged: true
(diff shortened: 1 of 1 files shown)


prow/cluster/jobs/istio/operator/istio.operator.yaml
```diff
@@ -2,6 +2,45 @@ presubmits:
...
+          limits:
+            cpu: "5"
+            memory: 24Gi
+          requests:
+            cpu: "3"
+            memory: 3Gi
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /lib/modules
```
howardjohn, line 27: do we need this for dind? I thought it was just kind is special somehow. afaik we have other tests using docker (hadolint runs in docker?) without this.

Also why do we need dind for CI? can't we just set the image to our new build container

sdake: this is prototype work.  We can't just set the image to our new build container because we want to run make commands naturally without non-CI workflows in the makefiles.   See this PR: https://github.com/istio/operator/pull/151

Notice make lint fails, but the others succeed because they are callling `go build` and `go test` directly.

FWIW @geeknoid has greenlit this work :)

sdake: Not sure on the volume mounts, I just want something that works for the moment and copied the install - as it uses a complex Kind in K8s model.  Can refine after prototype work is done.

howardjohn:
I would suggest using mkpj and mkpod in the future to test these type of
changes - you can create a k8s Pod definition that is exactly what is run
and run it in your own cluster. I'll write up some docs for this soon

PR #1597, opened 2019-08-05
add nogo vet with default analyzers

clarketm:
This is in the effort towards modernization, #1396. 

Enable [nogo](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst) vet static analysis using the [*default*](https://github.com/bazelbuild/rules_go/blob/master/go/nogo.rst#running-vet) analyzers. 

> **Note:** these checks are run alongside the compiler when building any Go target (e.g. go_library)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -3,6 +3,7 @@
 
 load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
 load("@bazel_gazelle//:def.bzl", "gazelle")
+load("@io_bazel_rules_go//go:def.bzl", "nogo")
 
 buildifier(
     name = "buildifier",
@@ -24,4 +25,10 @@ platform(
       }
       {PARENT_REMOTE_EXECUTION_PROPERTIES}
     """,
+)
+
+nogo(
+    name = "nogo_vet",
+    vet = True,
+    visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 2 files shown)


howardjohn:
Istio has a common linter configuration that this repo has already been migrated too. See https://github.com/istio/common-files.

Does this add any value on top of this?

I would rather see us moving away from bazel and towards standard istio tooling than the other direction

clarketm:
@howardjohn - `nogo` is a Bazel go rule that uses heuristic to find common errors missed by the compiler. It is the equivalent of the [`go vet`](https://golang.org/cmd/vet/) command with a subset of the default analyzers enabled.

If the desire is to get away from bazel then what are your thought of adding this as a **common** file using `go vet` to get the same benefit but without the bazel-specific implementation?

howardjohn:
@clarketm I believe we are already running go vet with https://github.com/istio/common-files/blob/master/files/.golangci.yml#L56

fejta (approved):
> Does this add any value on top of this?

This provides more consistency with upstream in how we build and validate prow and the rest of our test-infra.

/hold

clarketm:
@howardjohn - thanks for pointing me to the reference 👍 . I missed that on initial inspection.

geeknoid:
I don't think we need this since we run the full suite of Go linters on this repo already, consistent with all the Istio repos.

So can we close?

PR #1599, opened 2019-08-06
Update prow to v20190805-b43f41119

clarketm:
In the future, this should be run on an interval as a periodic job (#1396). 

@cjwagner I know we spoke about this briefly - should creating a periodic for autobump be done *before* or *after* migrating to use a separate build cluster?

### DIFF
--- a/prow/bump.sh
+++ b/prow/bump.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -o errexit
 set -o nounset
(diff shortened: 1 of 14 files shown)


cjwagner (approved):
> @cjwagner I know we spoke about this briefly - should creating a periodic for autobump be done _before_ or _after_ migrating to use a separate build cluster?

We need to start using a build cluster config file before we can run trusted jobs, but we don't necessarily need to move all the existing (untrusted) jobs to a separate build cluster first. I think we might want to though based on the discussion here: https://github.com/kubernetes/test-infra/pull/12382

PR #1600, opened 2019-08-06
add time comparer

ChristinaLyu0710-zz:
Read build-log.txt files according to paths given in spreadsheet. Find time commands and filter them out to csv output in sorted order based on the average run time.


fejta (commented):
Can you put the errorfinder change into its own PR? I don't understand how these two commits are related.

ChristinaLyu0710-zz:
@fejta If I don't put them in the same pr the time comparer pr would fail the the lint check for not changing the errorfinder code.

fejta:
Change the errorfinder code first?

ChristinaLyu0710-zz:
@fejta Updated pr at https://github.com/istio/test-infra/pull/1608

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+		}
+
+		fileStack := stack.New()
+		for start := 0; start < len(fileSlice); start++ {
+			line := fileSlice[start]
+			if timeCommand.MatchString(line) {
+				fileStack.Push(line)
+			}
+
+			if realOutput.MatchString(fileSlice[start]) && userOutput.MatchString(fileSlice[start+1]) && sysOutput.MatchString(fileSlice[start+2]) {
```
utka, line 388: What if is the end of the file will fileSlice[start+1] crash?

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+						commandToTime[corespondingCommand] = rCombination
+					}
+				}
+			}
+		}
+	}
+	return commandToTime
+}
+
+// For each spliter section of file paths, read previously generated errors and warnings.
```
utka: Are you combining two scripts, finding errors and finding time?

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+
+	var apiKey string
+	flag.StringVar(&apiKey, "APIKey", "", "API key to create new google sheets service")
+
+	var spliter int
+	flag.IntVar(&spliter, "SplitBy", 2, "The number of run paths to process for each read/write section.")
+
+	flag.Parse()
+
+	if strings.Compare(fileName, "") == 0 {
```
fejta, line 543: `fileName != ""`

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+	}
+
+	if public {
+		if err := obj.ACL().Set(ctx, storage.AllUsers, storage.RoleReader); err != nil {
+			log.Fatalf("Unable to set object to public: %v", err)
+		}
+	}
+
+	_, err = obj.Attrs(ctx)
+	if err != nil {
```
fejta: combine with previous line please `if _, err := obj.Attrs(ctx); err != nil {`

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+func (tc *TimeComparer) CopyToGCS(ctx context.Context, source, bucketName, fileName string, public bool) {
+	r, err := os.Open(source)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer r.Close()
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		log.Fatalf("Unable to create new client: %v", err)
```
fejta: why don't you return an error rather than blowing up?

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+		n++
+	}
+	startInd := (n - 1) * spliter
+	endInd := len(gcsFilePaths)
+	tc.findErrorForEachSection(ctx, gcsFilePaths, outputFileName, startInd, endInd)
+
+}
+
+// Optional on whether or not to copy and share output file to gcs.
+func (tc *TimeComparer) CopyToGCS(ctx context.Context, source, bucketName, fileName string, public bool) {
```
fejta: What other package needs to use `CopyToGCS` right now? If none please do not export it`

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+		stringTime: timePart,
+	}
+
+	return rPair
+
+}
+
+// Compute average user/real/sys time to compare with others.
+func getAverageTime(realTime, userTime, sysTime runPair) float64 {
+	averageSlice := []float64{}
```
fejta: `var averageSlice []float64` (everywhere)

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+	realT := realTime.time
+	userT := userTime.time
+	sysT := sysTime.time
+
+	for i := 0; i < 2; i++ {
```
fejta: Does this really justify a for loop?

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+		}
+		if len(timeSplit) > 1 {
+			secString := timeSplit[len(timeSplit)-1]
+			secValueString := strings.Split(secString, "s")[0]
+			fmt.Println("get seconds")
+			fmt.Println(secValueString)
+			secFloat, err := strconv.ParseFloat(secValueString, 64)
+			if err != nil {
+				fmt.Println("error parsing float second")
+				timeSlice = append(timeSlice, -1.0)
```
fejta: Why do we want to append -1 rather than returning an error?

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+		return []string{}, err
+	}
+	defer rc.Close()
+	lines := []string{}
+
+	scanner := bufio.NewScanner(rc)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if len(line) != 0 {
```
fejta: In general try to minimize indentation and return early.

So 
```golang
if len(line) == 0 {
  continue
}
lines = append(lines, line)
```

has lower indentation than the current

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,532 @@
...
+
+	if len(resp.Values) == 0 {
+		fmt.Println("No data found.")
+		return nil
+	}
+	filePaths := []string{}
+
+	for _, row := range resp.Values {
+		filePath := row[0]
+		filePathString, err := filePath.(string)
```
fejta: `err` is not the correct name. It is a `bool` that determines if the cast was successful, not an `error`.

Should you log something if you cannot cast the value?

toolbox/timecomparer/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Compare Time Commands From Build-log.txt
+
+The file reads dirctory paths to prs that contain build-log.txt in GCS from Google Spreadsheet. It processes each build-log.txt file from the directories and finds places where `time (*)` commands are called. It then checks the output for time command to get the amount of time the commands run for. It finally generates an output that sort the time commands based on the average run time they take.
+
+To run the file, you will need a `credentials.json` file from https://developers.google.com/sheets/api/quickstart/go when you click `ENABLE THE GOOGLE SHEETS API` and `DOWNLOAD CLIENT CONFIGURATION` to download the `credentials.json` file and use it as path to flag -CredentialsPath.
+
+You will also need a google spreadsheet id that contains the path to pr folder in a bucket in Google Cloud Storage. For a weblink `https://docs.google.com/spreadsheets/d/1fUmFA2CcnSQuCwEilAewftwitvSOLZGcj2BQiur37rI/edit#gid=935149480`, the spreadsheet id to pass to flag SpreadsheetID is `1fUmFA2CcnSQuCwEilAewftwitvSOLZGcj2BQiur37rI`.
+
+```
```
fejta: Use something like ` ```console` or ` ```shell`

toolbox/timecomparer/README.md
```diff
@@ -0,0 +1,34 @@
...
+$ git clone https://github.com/istio/test-infra.git
```
fejta, line 12: Consider dropping the `$` so this is easier to copy and paste. 

Alternatively if you keep the `$` then it should show some example output

ChristinaLyu0710-zz: Most of the commands do not have output when they are run successfully.

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,543 @@
...
+
+			if realOutput.MatchString(fileSlice[start]) && userOutput.MatchString(fileSlice[start+1]) && sysOutput.MatchString(fileSlice[start+2]) {
+				if fileStack.Len() != 0 {
+					previousCommand := fileStack.Pop()
+					corespondingCommand := previousCommand.(string)
+					// Find each section of time in the output
+					realTime := extractTime(fileSlice[start])
+					userTime := extractTime(fileSlice[start+1])
+					sysTime := extractTime(fileSlice[start+2])
+					average := getAverageTime(realTime, userTime, sysTime)
```
utka: Why do you want to use average time? I would use real.

vendor/github.com/golang-collections/collections/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2012 Caleb Doxsey
```
fejta, line 1: I find it confusing that there are vendor changes without any `go.mod` changes.

This seems wrong and makes me suspect it will cause issues the next time someone runs the update-deps script

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bufio"
+	con "context"
```
fejta, line 19: Will you leave this as context? Everyone always imports this package as `context`

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+	}
+
+	defer file.Close()
+
+	writer := csv.NewWriter(file)
+	defer writer.Flush()
+
+	for _, sortedTime := range sortedArray {
+		comAndTime := timeMap[sortedTime]
+		emptyCom := commandAndTime{}
```
fejta: `var emptyCom commandAndTime`

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+		if reflect.DeepEqual(comAndTime, emptyCom) {
```
fejta: I would be more inclined to add an Equals method to the commandAndTime struct rather than using reflection

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+		completeResult := append(originalResult, newTimeResult...)
+
+		averageTime := strconv.FormatFloat(sortedTime, 'E', -1, 64)
+
+		var toWrite []string
+		toWrite = append(toWrite, com, averageTime)
+		toWrite = append(toWrite, completeResult...)
+		err = writer.Write(toWrite)
+		if err != nil {
+			return err
```
fejta: a) why not wait until you exit the for loop to make the expensive write call? AKA just write once, outside of the loop and do a bunch of appends inside it
b) `return fmt.Errorf("write %s error: %v", sortedTime, err)` will be a more helpful message to return

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+	defer writer.Flush()
+
+	for _, sortedTime := range sortedArray {
+		comAndTime := timeMap[sortedTime]
+		emptyCom := commandAndTime{}
+		if reflect.DeepEqual(comAndTime, emptyCom) {
+			continue
+		}
+
+		com := comAndTime.command
```
fejta, line 208: Seems redundant to have both comAndTime and com. Consider getting rid of comAndTime

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+	timeSplit := strings.Fields(runTimeString)
+	length := len(timeSplit)
+	var ind int
+	for ind = length - 1; ind > 0; ind-- {
+		if strings.Compare(timeSplit[ind], "") != 0 {
+			break
+		}
+	}
+	timePart := timeSplit[ind]
+	timeSlice, _ := convertStringToTime(timePart)
```
fejta: don't ignore the error, please

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+	realT := realTime.time
+	seconds := realT[0]*60 + realT[1]
+
+	return seconds
+}
+
+// Get error contents of files to update command-time map with commands and path to build logs.
+func (tc *TimeComparer) findTimeCommands(
+	ctx context.Context, filePaths []string, commandToTime map[string]runCombination) map[string]runCombination {
+	timeCommand := regexp.MustCompile(`^time (.*);`)
```
fejta, line 371: Consider defining these vars outside the function, so we can reuse them between calls

toolbox/timecomparer/main.go
fejta: This seems unresolved still?

toolbox/timecomparer/main.go
```diff
@@ -0,0 +1,537 @@
...
+		log.Fatal(err)
+	}
+	client, err := storage.NewClient(context)
+	if err != nil {
+		log.Fatalf("Unable to create new client: %v", err)
+	}
+
+	timeComparer := NewTimeComparer(client, bucketName)
+	timeComparer.divideToSections(context, spliter, listOfPrs, fileName)
+	err = timeComparer.copyToGCS(context, fileName, bucketName, fileName, public)
```
fejta: Send the storage client into here?

toolbox/timecomparer/main.go
ChristinaLyu0710-zz: Because I have another import that has the name context so i decided to use "con" for context instead because a lot of people use the package and therefore will understand what I'm doing when i call con.Context for example.

toolbox/timecomparer/main.go
ChristinaLyu0710-zz: a) (csv) writer.Write() only writes one record of csv (one line)

PR #1595, opened 2019-08-05
Use KinD for optional E2E tests

howardjohn:
Migrate these tests over to using KinD instead of boskos. This will make
them much cheaper and more reproducible locally.

Currently only doing the optional tests as a canary, once these are
stable more will be migrated over.

Blocked by https://github.com/istio/istio/pull/16062

PR #1604, opened 2019-08-06
Move more tests to using kind

howardjohn:
Do not merge until https://github.com/istio/test-infra/pull/1595 has soaked for a day or so

I manually tested all of these


howardjohn:
OK the initial wave went in and seems to be passing, we can let this soak more but I think its safe to merge this now.

costinm (approved):

PR #1607, opened 2019-08-06
Remove legacy testenv folder

howardjohn:
This has been replaced by
https://github.com/istio/tools/tree/master/perf. The configs here have
not been updated since Istio 0.2 and are no longer relevant or used by
anything.


fejta (approved):
/hold

PR #1610, opened 2019-08-06
Remove unsupported sisyphus tool

howardjohn:
This tool is unused, unmaintained, and does not work as is. If we do
decide we want this again in the future we can always revert git
history.


geeknoid (approved):

fejta:
Run update_bazel.sh?

fejta (approved):

PR #1614, opened 2019-08-06
Enforce branch protection settings even for admins.

geeknoid:


howardjohn:
Is there any way to override this if something needs manual merge?

geeknoid:
Yes, admins can still force merge things as before.

Turns out right now, anybody that's an admin is considered effectively as a code owner by GitHub. So if you're an admin and you approve a PR, whether or not you're in CODEOWNERS, GitHub will consider it approved. Flipping this bit will hopefully fix this behavior.

howardjohn:
Interesting
/lgtm

fejta (approved):
/hold

fejta:
I suspect this will prevent tide from merging correctly in repos that don't have CODEOWNERS, like this one.

geeknoid:
@fejta Why is that?

My reading of the setting says that admins are subject to the normal branch protection rules like normal user, and tide runs as a repo admin so this does affect it. But tide should only ever be merging things that passes all of GitHub's mergability rules, so this should be OK.

PR #1617, opened 2019-08-07
Upgrade helm to the recommended minimum version: 2.10

jwendell:
It has goodies like `--set-string`.


howardjohn:
Ah you beat me :slightly_smiling_face: https://github.com/istio/test-infra/pull/1618 does the same but also updates the image. We don't have this automated yet unfortunately

jwendell:
Ah ok, sounds good, thanks.

PR #1623, opened 2019-08-07
Revert "Revert "Move more tests to using kind (#1604)""

howardjohn:
Reverts istio/test-infra#1621

I attempted to fix this in #1618, then decided it didn't work and instead reverted. Turns out, it did work and the test I was looking at didn't pick up the change. 

See this test result, which shows the simpleTest passing with KinD: https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio/16066/e2e-simpleTests-master/5148.


nmittler (approved):
Great PR title! :)

PR #1628, opened 2019-08-08
Remove vendor dir.

geeknoid:


geeknoid:
This isn't as easy as it seemed. Presumably, bazel is the root cause.

Looking....

geeknoid:
I'm giving up on this for now

PR #1635, opened 2019-08-08
Remove Bazel usage from the repo

howardjohn:
All presubmits are now off bazel, and none of this will work now thay we
removed vendor. I think I got everything, and hopefully didn't remove
anything we actually do want.


geeknoid (approved):
Nice.

fejta:
My desire is for this repo to continue using bazel.

Others can merge this, but it does not interest me. Merging it will minimize my enthusiasm for contributing to this repo.

kyessenov:
Is there anything to be gained by removing bazel?
All proxy/google3 developers use bazel daily and are familiar with it.

howardjohn:
> Others can merge this, but it does not interest me. Merging it will minimize my enthusiasm for contributing to this repo.

You are one of two owners, so I have much choice otherwise

> Is there anything to be gained by removing bazel?
> All proxy/google3 developers use bazel daily and are familiar with it.

This is the istio repo not proxy/google3. Most istio developers are not familiar with bazel. A goal we have is to get more istio developers contributing here, rather than having a single person take on all the burden (especially painful when they leave and no one knows anything anymore). All other istio repos are not using bazel. In fact, this repo is not even using bazel at all in the presubmits, this is just a cleanup. It is misleading to have all this bazel stuff that is unused

kyessenov:
All other istio repos -- what about envoy/envoy-wasm/proxy? That is a core part of istio, and is 100% dependent on bazel at this point.

kyessenov:
If there is a problem with automating bazel declarations, that needs to be addressed. But removing things for the sake of change does not seem right to me. In fact, it will likely make things more complicated once we need to make builds in bazel-land (e.g. third party).

howardjohn:
>  But removing things for the sake of change does not seem right to me. 

Its not removing things for the sake of change. Bazel is not used in this repo... everything is just using `go test`/`go build`, etc.

Just like it is over complicated to have 2 CIs, it is just needlessly complicated to have 2 build systems.

> In fact, it will likely make things more complicated once we need to make builds in bazel-land (e.g. third party).

1. I have seen no indication that this will ever happen. Are there any plans?
2. I don't think we need to keep things around "just in case". Seems equivalent to leaving around commented out code.



I do not see any benefits to leaving around a bunch of BUILD.bazel files that are not tested or used

fejta:
> I do not see any benefits to leaving around a bunch of BUILD.bazel files that are not tested or used

I'll certainly keep using them -- I don't like Makefiles and I don't like having to spend a bunch of effort resolving problems that are already solved by bazel (ensuring I'm building each repo with the correct golang version, for example). Nor do I like turning my machine into a wind tunnel every time I build.

kyessenov:
@howardjohn Bits of istio already exist in third_party. But changes like this, I'm sure, makes it harder to bring the rest of it.

fejta:
> this repo is not even using bazel at all in the presubmits

Primary reason for this is that we don't yet have a way to build kubernetes/test-infra via bazel w/o a vendor dir. Once we resolve this I would hope we can start using bazel in this repo again.

howardjohn:
I would think we should either use bazel in istio/istio and istio/test-infra or use it in neither. I don't understand why we would want to have different build systems in different repos

kyessenov:
Because they don't conflict with one another? The argument doesn't work well once you consider that istio/api and istio/proxy both depend on bazel. We'll have other languages for WASM, and mandating opinionated go tooling from istio/istio on everything else is not rational. I don't consider Makefiles as  a real build system TBH.

fejta:
I don't see why they need to be related. Different people work on those two repos, right?

Does istio/istio even have a build system? The presence of bazel rules does not prevent `go test ./whatever` from working the same as normal,

howardjohn:
> I'll certainly keep using them 

Does it even work without vendor now? `bazel build //...` seems to fail for me...
(note - I didn't realize this would break when I removed vendor, as it wasn't used in presubmit)

> I don't see why they need to be related. Different people work on those two repos, right?

This is exactly the problem. We want this to be accessible/understandable to most Istio contributors instead of just a few. 

> The argument doesn't work well once you consider that istio/api and istio/proxy both depend on bazel.

My understanding was just proxy used it because of its tight coupling to Envoy, I didn't realize API did or wasm would as well.


If we really want to keep bazel around I don't have that strong feelings, and we could close this.

geeknoid:
@fejta @kyessenov 

We want Istio to offer a common infrastructure experience to Istio developers. This means a single CI system and a single way to declare build steps. For the time being, this means using Prow for CI and using Makefiles to declare build steps.

As John explains, Bazel is really superfluous in this repo, it is not used. Keeping the cruft there is misleading and unnecessary. It merely serves to disenchant Istio developers which  rejected Bazel quite thoroughly in the past. So let's clean this up.

If we decide *holistically* that Bazel is needed throughout Istio due to the WASM work we're doing, then let's *create a plan* around  that and then we can reintroduce Bazel in this repo at that time.

kyessenov:
Martin, I don't think it's fair to put a burden of planning as a post-factum after the common infrastructure change. We are already using Bazel for WASM work. I don't see any other reasonable build system given the complexity of envoy build. I think we should have considered this before designing the common infrastructure and making vast changes across all repos. I haven't heard any discussion how common infrastructure or new prow tooling is going to support bazel developers. We're mostly victims of constant breakages due to onslaught of changes to the build infra.

fejta:
Upstream prow already requires bazel.

This repo used bazel until last week when John removed bazel from the presubmits and then broke bazel builds.

People who like not using bazel can continue using not using bazel. They don't need it to edit prow job yaml files.

Preventing people who like bazel from benefiting from it seems needlessly antagonistic.

howardjohn:
Bazel currently does not work as a result of removing vendor (not my intent when making the change). I don't think it makes sense to keep bazel around if it doesn't work -- that is just actively confusing to developers.

If someone wants to spend time maintaining bazel here, then *maybe* we could keep it around, but from what I have heard that cannot be done without adding back vendor, which goes against the projects goals.

> Preventing people who like bazel from benefiting from it seems needlessly antagonistic.

We are also preventing people from using maven, gradle, ... Having multiple build systems seems similar to me to having multiple CI systems.

fejta:
> cannot be done without adding back vendor

Vendor is not needed. Just requires someone doing the work to get it to build without it.

sebastienvas:
@howardjohn how did you get around to merge your PR ? bazel tests were part of the presubmits. I am trying to make a change to the mason component and nothing works.

howardjohn:
@sebastienvas what do you mean by "nothing works? `go build ./...` and `go test ./...` pass

sebastienvas:
FWIW not even the makefile work.

fejta:
Once https://github.com/kubernetes/test-infra/pull/14036 goes in I plan on updating this repo to also build w/o a vendor dir, which should get bazel back into a working state of affairs.

fejta:
k/t-i head has removed vendor, plan to work on pulling that into into this repo and getting the bazel rules working again.

PR #1637, opened 2019-08-08
Simplify how we update prow

fejta:
/assign @michelle192837 @cjwagner 

* Ensure we target the correct context
* Drop all the redundant rules -- we don't want people updating prow this way, and using `apply -f cluster/foo_deployment.yaml` instead of just `apply -f cluster/` we're likely to cause issues by forgetting to add a new component

### DIFF
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -10,6 +10,7 @@ CLUSTER ?= prow
 
 # ensures that kubectl has access to the prow cluster's config context
 get-cluster-credentials:
+	# TODO(fejta): https://github.com/istio/test-infra/issues/1636
 	gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
 
 update-config-dry-run: get-cluster-credentials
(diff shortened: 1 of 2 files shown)


michelle192837:
README needs an update too.

/hold

michelle192837 (approved):

howardjohn (approved):

fejta:
/hold cancel

Updated documentation

prow/README.md
```diff
@@ -188,36 +126,25 @@ This repository (istio/test-infra) also provides Prow jobs.
...
 ```bash
-$ git clone https://github.com/kubernetes/test-infra
-$ bazel build //prow/cmd/mkpj
-```
-And then specify the script and the job you want to trigger:
-``` bash
-$ bazel-bin/prow/cmd/mkpj/mkpj --config-path ~/istio/test-infra/prow/config.yaml --job test-infra-presubmit | kubectl create -f -
-```
+# Assuming you cannot click the rerun button on prow.istio.io,
+# if you are oncall, do the following:
```
cjwagner: nit: They could still use the rerun command. The mkpj example is still useful for when there aren't any existing runs of the job though.

prow/README.md
```diff
@@ -45,60 +30,13 @@ horologium-617344823-js4mk   1/1       Running   0          50m
...
-* In case of deployment failure, check the logs by running
-
 ```
-$ kubectl logs deck-3621325446-9pdqw
-```
-
-* You might notice some issues related to config error. In that case fix and redeploy the config
-
-```
-$ make update-config
```
cjwagner, line 60: We should probably keep this part?

prow/Makefile
```diff
@@ -25,82 +26,5 @@ update-config: get-cluster-credentials
 		--plugins-config-path=plugins.yaml \
 		--wet
 
-# use to guarantee that on command completion all deployments see new configs
-update-configs: update-config update-plugins
-
-.PHONY: update-config update-plugins get-cluster-credentials update-configs
```
cjwagner, line 31: Lets keep the applicable `.PHONY` targets and the `update-configs` target?
Also it looks like `update-plugins` doesn't exist and is covered by `update-config`.

cjwagner (approved):
/hold

prow/Makefile
fejta: Done

prow/README.md
fejta: A plugin does this automatically on merge now. It is no longer a common operation.

prow/README.md
fejta: Hence the previous line

prow/README.md
cjwagner: Agreed, but if there is a dropped webhook or some other error (e.g. the cluster is deleted) we need a mechanism to bootstrap the config.

prow/README.md
cjwagner: Even if you cannot click the rerun button, if you are oncall you can still run the rerun command.

fejta: Exactly, if you can click the rerun button, the rest of the sentence does not apply

prow/README.md
fejta: Right, and we don't need that to be in top level documentation, just like we don't document how to create a new node pool or a billion other generally irrelevant things.

PR #1644, opened 2019-08-09
Remove specialized prow runner scripts from istio

howardjohn:
Right now we have some scripts that are mostly just very light wrappers
around make. It makes (no pun intended) more sense to just directly call
make and clean up these scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+export GO111MODULE ?= on
+
 lint:
 	@scripts/run_golangci.sh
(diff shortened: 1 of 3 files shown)


geeknoid (approved):

howardjohn:
/assign @fejta 

for Makefile approval

fejta (approved):
UGH, Makefiles are a terrible abomination. Long-term I will not participate any repo that requires my using them

/hold

PR #1658, opened 2019-08-12
Fix a couple bugs.

geeknoid:
- Enable tide on test-infra. Somehow, that got dropped.

- Remove test-infra from some of the old-school plugins so it works
cleanly with CODEOWNERS.

- Only run prow jobs for official branches in the test-infra repo. This will
fix the branch protector.

Combined with enabling write permission for istio developers on the test-infra repo, like all our other repos, should get  everything working normally on this repo.

### DIFF
--- a/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.yaml
+++ b/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.yaml
@@ -2,6 +2,9 @@ presubmits:
   istio/test-infra:
 
   - name: build-test
+    branches:
+      - ^master$
+      - ^release-1.*$
     always_run: true
     decorate: true
     path_alias: istio.io/test-infra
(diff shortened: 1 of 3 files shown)


howardjohn:
I still don't have write access and therefor cannot approve

howardjohn (approved):

geeknoid:
@howardjohn For some reason, you weren't in the Istio Developers team. Now you are, and now you  have write access to this repo.

sdake (approved):
looks good.  if we can't get htis moving - we can always revert this and codeowners.

However, I prefer to move forward rather than backwards...

geeknoid:
Forcing this in, since this in blocking our 1.3 release work.

fejta:
We should resolve outages by rolling back to the known good version. This is faster and safer.

PR #1662, opened 2019-08-13
Remove merges-blocked-needs-admin from release-1.3

howardjohn:
Instead of having a context that never passes, we will instead change
the owners to be just the release managers, so we can properly utilize
tide instead of force merging changes.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -82,7 +82,7 @@ blocked_branches: &blocked_branches
       teams:
       - release-managers-1-2
   release-1.3: &release13
-    <<: *blocking_merge
+    protect: true
     restrictions:
       teams:
       - release-managers-1-3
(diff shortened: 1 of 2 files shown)


prow/config.yaml
```diff
@@ -82,7 +82,7 @@ blocked_branches: &blocked_branches
       teams:
       - release-managers-1-2
   release-1.3: &release13
-    <<: *blocking_merge
+    protect: true
```
rlenglet, line 85: Isn't that already done below?

rlenglet (approved):
LGTM

prow/config.yaml
howardjohn: it was part of the `blocking_merge` block.. yaml magic

geeknoid (approved):

fejta (approved):

PR #1666, opened 2019-08-14
Simplify by collapsing now common things up to the org level.

geeknoid:


prow/config.yaml
```diff
@@ -99,61 +99,19 @@ branch-protection:
         required_approving_review_count: 1
         require_code_owner_reviews: true
       restrictions:
+        teams:
+          - repo-admins
+      branches: *blocked_branches
```
fejta: This doesn't do anything -- https://github.com/kubernetes/test-infra/blob/0a2e51dc8afa02adebff7c2a5e899a316e042893/prow/config/branch_protection.go#L190-L194

(hence the unit test failures)

Things specified at the org level becomes the default for all branches in all repos (unless overridden at the repo or org level)

geeknoid: Too bad this didn't work, it was a lot of cruft gone away. I've backed down to something not as delete-happy, but at least it works now :-)

clarketm (approved):

PR #1669, opened 2019-08-14
Enable tide for CNI

howardjohn:


geeknoid (approved):

prow/config.yaml
```diff
@@ -120,7 +120,16 @@ branch-protection:
             contexts:
             - deploy/netlify
         cni:
+          required_pull_request_reviews: *tide_withcodeowners_reviews
```
clarketm:
```suggestion

```

Now that #1666 has merged , the `tide_withcodeowners_reviews` and `tide_withcodeowners_restrictions` anchors are no longer defined. You should be able to remove these lines and use org *defaults* (or override as needed).

prow/config.yaml
```diff
@@ -120,7 +120,16 @@ branch-protection:
...
+          restrictions: *tide_withcodeowners_restrictions
```
clarketm:
```suggestion

```

clarketm (changes requested):

prow/config.yaml
clarketm: Otherwise, there may be unintentional conflicts or reversions.

howardjohn:
Thanks @clarketm , should be fixed now

geeknoid (approved):

clarketm (approved):

PR #1672, opened 2019-08-14
Add E2E test for CNI repo

howardjohn:


prow/cluster/jobs/istio/cni/istio.cni.master.yaml
```diff
@@ -113,6 +113,54 @@ postsubmits:
           privileged: true
       nodeSelector:
         testing: test-pool
+  - annotations:
+      testgrid-alert-email: istio-oncall@googlegroups.com
+      testgrid-dashboards: istio-cni-postsubmits-master
+      testgrid-num-failures-to-alert: "1"
```
fejta, line 119: Is this supposed to be a string instead of int? Not sure if the quotes change things.

prow/cluster/jobs/istio/cni/istio.cni.master.yaml
```diff
@@ -113,6 +113,54 @@ postsubmits:
...
+    name: cni-e2e-master
+    path_alias: istio.io/cni
+    spec:
+      containers:
+      - command:
+        - entrypoint
+        - make
+        - prow-e2e
+        image: gcr.io/istio-testing/istio-builder:v20190807-7d818206
+        name: ""
```
fejta, line 137: Feel free to drop as its ignored anyway

prow/cluster/jobs/istio/cni/istio.cni.master.yaml
```diff
@@ -113,6 +113,54 @@ postsubmits:
...
+        resources:
+          limits:
+            cpu: "3"
+            memory: 24Gi
+          requests:
+            cpu: 500m
+            memory: 3Gi
+        securityContext:
+          privileged: true
```
fejta, line 146: Would be nice if we can figure out how to do these builds without privilege at some point, like by using kaniko

prow/config/jobs/cni.yaml
```diff
@@ -13,7 +13,10 @@ jobs:
     command: [make, lint]
   - name: cni-install
     command: [make, docker, test]
-
+  - name: cni-e2e
```
fejta, line 16: Can we have something that ensures the generated version is up to date?

fejta (approved):
/hold

prow/config/jobs/cni.yaml
howardjohn: yes we do have a presubmit check (generate-check)

prow/cluster/jobs/istio/cni/istio.cni.master.yaml
howardjohn: yes, I think because k8s annotations must be strings

prow/cluster/jobs/istio/cni/istio.cni.master.yaml
howardjohn: I think we would need to change the tag in the struct in kubernetes repo to be omit_if_empty for this

prow/cluster/jobs/istio/cni/istio.cni.master.yaml
howardjohn: can we run kind without privilege? if possible I would love to but I didn't think that was feasible


for other tests though like unit tests we should definitely

geeknoid (approved):

PR #1676, opened 2019-08-16
Set BUILD_WITH_CONTAINER to 0 for all jobs.

geeknoid:
By default, we want humans building Istio stuff to be building through a
container. But the CI system already cranks up jobs within said container,
so we never want to have the CI's call to 'make' to induce a dind call
to the container. So we set BUILD_WITH_CONTAINER to 0 for all CI jobs.

### DIFF
--- a/prow/cluster/jobs/istio/operator/istio.operator.yaml
+++ b/prow/cluster/jobs/istio/operator/istio.operator.yaml
@@ -18,10 +18,6 @@ presubmits:
         env:
         - name: GO111MODULE
           value: "on"
-        - name: BUILD_WITH_CONTAINER
-          value: "0"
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

howardjohn (commented):
you need to quote the 0 though

sdake (approved):

sdake:
Must remove the redundant operator settings for this (this was part of a transition):

```
    config_test.go:29: could not read configs: job operator-lint failed to merge presets for podspec: env var duplicated in pod spec: BUILD_WITH_CONTAINER
```

fejta (approved):
/hold

PR #1678, opened 2019-08-16
Update test image for 1.1/1.2

howardjohn:
New envoy changes depend on a newer libc. Updating the image to the one
used in master should fix this. As far as I know there are no
incompatible changes in the newer image. I tested the unit tests locally
with this change.

fyi @ozevren


ozevren:
Thanks for fixing this!

geeknoid (approved):

PR #1681, opened 2019-08-17
Install libc++ in istio-builder images.

PiotrSikora:

### DIFF
--- a/docker/istio/istio/install.sh
+++ b/docker/istio/istio/install.sh
@@ -32,6 +32,7 @@ gem install --no-ri --no-rdoc fpm
 ./install-helm.sh
 ./install-protoc.sh
 ./install-yamllint.sh
+./install-libcxx.sh
 
 apt-get clean
 rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


duderino (approved):

PiotrSikora:
Note: it's using new `install-libcxx.sh` instead of the existing `install-clang.sh`, to avoid installing LLVM and Clang in the image, but I'm happy to switch to `install-clang.sh` if that's preferred.

PiotrSikora:
Note: we only need this change for the images used to build `release-1.1` and `release-1.2` branches, where we need libc++-7 instead of libc++-8.

howardjohn (approved):
thanks, I'll build and update tomorrow if someone doesn't before then

howardjohn:
@PiotrSikora when you say this is only needed on 1.1 and 1.2 does that mean we cannot use this for 1.3 or we just don't need it for 1.3?

PiotrSikora:
@howardjohn we don't need it, and it's non-issue if it's in the image.

clarketm (approved):

PR #1686, opened 2019-08-17
Switch the pkg and bots repos to the build-tools image.

geeknoid:


prow/cluster/jobs/istio/pkg/istio.pkg.yaml
```diff
@@ -30,14 +27,11 @@ presubmits:
       always_run: true
       spec:
         containers:
-          - image: gcr.io/istio-testing/istio-builder:v20190624-478286dc
+          - image: gcr.io/istio-testing/build-tools:2019-08-16T20-50-36
             command:
               - go
```
sdake, line 32: might consider introducing a make target for this build operation.

sdake (approved):

PR #1687, opened 2019-08-17
Update common files.

geeknoid:


docker/OWNERS
```diff
@@ -1,4 +0,0 @@
-approvers:
```
howardjohn, line 1: why remove this? I think it's appropriate the users if the dockerfiles are the owners rather than prow team. Although long term I guess these will be in tools

howardjohn (approved):

geeknoid:
This repo is no longer using OWNERS files, so it's just unused.

.commonfiles.sha
```diff
@@ -1 +1 @@
-950d08195523c9f44f15e7283274259ee0ace593
+dd59c4efb573c9046fb80cbbef1c28382f58e8b6
```
fejta: What is this?

Makefile.common.mk
```diff
@@ -5,7 +5,7 @@
 # common-files repo, make the change there and check it in. Then come back to this repo and run
 # "make updatecommon".
 
-# Copyright 2018 Istio Authors
```
fejta, line 8: Shouldn't this stay as the original date?

.commonfiles.sha
geeknoid: The sha of the istio/common-files repo when the copy of the common-files was inserted into this repo.

Makefile.common.mk
geeknoid: Per CNCF recommendation, we'll be removing the date entirely in new copyright notices. In fact, this has been done in the common-files repo and will be replicated here the next time the common files are updated.

fejta:
lgtm

geeknoid:
@fejta Approve please.

Thanks.

clarketm (approved):

fejta:
Thanks Travis!

PR #1689, opened 2019-08-18
Use Bazelisk launcher for Bazel.

PiotrSikora:

### DIFF
--- a/docker/prowbazel/Dockerfile
+++ b/docker/prowbazel/Dockerfile
@@ -34,9 +34,9 @@ WORKDIR ${HOME}
 ENV TEST_TMPDIR /home/bootstrap/.cache/bazel
 
 # Move preloaded Bazel binaries to bootstrap's cache.
-RUN mkdir -p /home/bootstrap/.cache/
-RUN sudo mv /root/.cache/bazelisk ${HOME}/.cache/
-RUN sudo chown -R bootstrap:bootstrap ${HOME}/.cache/
+RUN mkdir -p /home/bootstrap/.cache/ \
+    && sudo mv /root/.cache/bazelisk ${HOME}/.cache/ \
+    && sudo chown -R bootstrap:bootstrap ${HOME}/.cache/
 
 # Add bootstrap, the test harness (checks out code, captures log and status)
 ADD https://raw.githubusercontent.com/nlandolfi/test-infra-1/d7900388fa1fb69738e231a21230373fd4b5ede9/jenkins/bootstrap.py ${HOME}
--- a/docker/prowbazel/Dockerfile
+++ b/docker/prowbazel/Dockerfile
@@ -33,6 +33,11 @@ WORKDIR ${HOME}
 # code and running bazel in different dir than the calling script.
 ENV TEST_TMPDIR /home/bootstrap/.cache/bazel
 
+# Move preloaded Bazel binaries to bootstrap's cache.
+RUN mkdir -p /home/bootstrap/.cache/
+RUN sudo mv /root/.cache/bazelisk ${HOME}/.cache/
+RUN sudo chown -R bootstrap:bootstrap ${HOME}/.cache/
+
 # Add bootstrap, the test harness (checks out code, captures log and status)
 ADD https://raw.githubusercontent.com/nlandolfi/test-infra-1/d7900388fa1fb69738e231a21230373fd4b5ede9/jenkins/bootstrap.py ${HOME}
 RUN sudo chmod +rx ${HOME}/bootstrap.py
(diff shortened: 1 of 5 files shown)


PiotrSikora:
Note: someone needs to push new images, since I don't have the permission to do it.

howardjohn (approved):

scripts/tools/linux-install-bazel
```diff
@@ -8,31 +8,32 @@ fi
...
+# Use Bazelisk launcher for Bazel (automatically downloads required version of Bazel).
+BAZELISK_VERSION="v1.0"
+BAZELISK_BASE_URL="https://github.com/bazelbuild/bazelisk/releases/download"
+BAZELISK_BIN="bazelisk-linux-amd64"
+BAZELISK_URL="${BAZELISK_BASE_URL}/${BAZELISK_VERSION}/${BAZELISK_BIN}"
+BAZELISK_DIRECTORY="${TOOLS_DIR}/bazelisk"
 
 function update_bazel() {
-  local bazel_current="$(bazel version | grep -oh '[0-9]*\.[0-9]*\.[0-9]*')" \
+  local bazel_current="$(bazel version | grep -oh 'Bazelisk version: v[0-9]*\.[0-9]*')" \
```
clarketm, line 19: This should also be `bazelisk` correct?
```suggestion
  local bazel_current="$(bazelisk version | grep -oh 'Bazelisk version: v[0-9]*\.[0-9]*')" \
```

PiotrSikora: No, it's installed as `/usr/local/bin/bazel`.

fejta (commented):
Is this for developers or test containers?

PiotrSikora:
> Is this for developers or test containers?

No, this installs Bazelisk on build images (in place of Bazel).

fejta:
Will this be used in CI? If so I'm concerned that this will result in every job downloading the one of a couple versions of bazel over and over again. Is there a way we can prepopulate the image with the versions of bazel we expect?

scripts/tools/linux-install-bazel
clarketm: Okay, I see.

PiotrSikora:
@fejta yes, it's going to be used and downloaded on every build in CI... unless we can store/restore cache in Prow? See https://github.com/istio/proxy/pull/2371 for how it's stored/restored in CircleCI.

This is unfortunate side-effect of this change, but the productivity drain that results from merging new images each time new Bazel is needed are worth it (...at least, for me).

fejta:
> Prow doesn't provide a runtime. Your image is your runtime. So you need to make sure your image has already downloaded the right bazel versions.
>
> From https://github.com/bazelbuild/bazelisk#how-does-bazelisk-know-which-bazel-version-to-run-and-where-to-get-it-from it looks one way to do this would be to `export USE_BAZEL_VERSION=0.28.1 bazelisk version` (and do that for a couple versions) in the Dockerfile.
> 
> That way we will create images that have 0.28.1 (etc) of bazel preinstalled

Yes, we could preinstall the few versions we need for the Istio Proxy releases we still maintain, but that will get outdated pretty soon (i.e. on the next release).

PiotrSikora:
@fejta added few know versions that we need.

docker/prowbazel/Dockerfile
```diff
@@ -33,6 +33,11 @@ WORKDIR ${HOME}
 # code and running bazel in different dir than the calling script.
 ENV TEST_TMPDIR /home/bootstrap/.cache/bazel
 
+# Move preloaded Bazel binaries to bootstrap's cache.
+RUN mkdir -p /home/bootstrap/.cache/
```
fejta: Each `RUN` statement creates a new docker layer. I'd recommend trying to combine these as much as possible to reduce the number of layers we need to download and extract:

Running multiple commands in one layer (at least combine these three commands into one):
* https://github.com/kubernetes/test-infra/blob/1972ae8f994d61b88f7a84c9947aa88bcb3aae18/images/bootstrap/Dockerfile#L30-L52

Adding multiple env vars into one layer (if you want to do a more aggressive refactor):
* https://github.com/kubernetes/test-infra/blob/1972ae8f994d61b88f7a84c9947aa88bcb3aae18/images/bootstrap/Dockerfile#L21-L22

PiotrSikora: Done & I don't.

fejta (approved):

PiotrSikora:
@fejta could you `prowbazel:0.5.9` image to `gcr.io/istio-testing/`?

howardjohn:
I built and pushed

PR #1703, opened 2019-08-21
Filter test used in k8s version tests

howardjohn:
Currently, these run all tests. This has basically a 100% failure rate,
giving us almost no signal. It will be more worthwhile to only run
nonflaky tests so that we can see when real breakages occur


fejta (approved):

duderino (approved):
Currently these tests fail so often that we have become conditioned to the overall test suite failing and ignore future failures.   By disabling these we can start paying attention to other tests in the suite that generally pass and which we should pay attention to when they fail.

fejta:
100% if we're releasing through broken tests, please delete the tests as they have negative value.

Issue #1670, opened 2019-08-14
Create a dedicate build cluster


howardjohn:
I thought there was another issue for this, but I can't find it. 

Currently we run our tests in the same cluster as prow runs. This is a security issue, and makes it so that changes we make in the build cluster can impact the prow infra, etc etc.

We should create a second cluster that will run our jobs and has minimal permissions possible

howardjohn:
I see the docs here: https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#run-test-pods-in-different-clusters

But they aren't clear enough to me to actually do this myself.. not sure where to put that config? 

It looks like k8s has a cluster 'test-infra-trusted' but doing a grep for that doesn't actually show where the cluster is configured.

clarketm:
@howardjohn - there is a reference to adding a separate build cluster https://github.com/istio/test-infra/issues/1396#issue-462905287 but no dedicated issue that I am aware of. 

I am actively working on getting a **build / test** cluster setup for Istio. I began provisioning the resources for the node pools yesterday. `test-infra-trusted` is actually the primary k8s-prow service cluster.

howardjohn:
Nice! thanks @clarketm . Is the cluster in istio-testing project still?

clarketm:
For isolation and organization, I am creating the cluster in a separate project `istio-prow-build`. Once the cluster is running, I will ensure IAM roles are delegated appropriately to the team and then we can plan the (incremental) migration of the jobs.

howardjohn:
@clarketm I was hoping to get some bigger nodes so our tests can run faster/less flaky. Should I just wait for this to do that? We currently run 8cpu nodes with 5core limit, I want at least 8 core limits, as our tests are pretty throttled right now

clarketm:
@howardjohn - I am using the current resource allocations from the default (`istio-testing`) prow cluster build/test pools as the starting point.

![image](https://user-images.githubusercontent.com/9957358/63404519-6b0d7480-c398-11e9-9c1e-5113b351819f.png)

If you have any specific tweak to this please provide them *now*. If possible, provide **all** node spec details you wish to customize for *both* the `build-pool` (`ssd-build-pool-2`) and `test-pool` (`test-pool-1`).

**e.g.**
![image](https://user-images.githubusercontent.com/9957358/63404876-aeb4ae00-c399-11e9-91f4-469a55fd5804.png)

clarketm:
Build cluster now active.
/close

[...]

PR #1713, opened 2019-08-23
update sinker, plank, deck deployments with "kubeconfig" secret

clarketm:
This is the required configuration for multi-cluster builds/tests, #1670. Currently, this is a **noop**, as the *default* build/test cluster will remain the same (i.e. current service cluster). An announcement will be made when (and how) the auxiliary build can be used - when it is properly vetted.

### DIFF
--- a/prow/cluster/deck_deployment.yaml
+++ b/prow/cluster/deck_deployment.yaml
@@ -30,12 +30,16 @@ spec:
         - --redirect-http-to=prow.istio.io
         - --spyglass
         - --tide-url=http://tide/
+        - --kubeconfig=/etc/kubeconfig/istio-config
         volumeMounts:
         - name: config
           mountPath: /etc/config
           readOnly: true
         - name: job-config
           mountPath: /etc/job-config
+          readOnly: true
+        - mountPath: /etc/kubeconfig
+          name: kubeconfig
           readOnly: true
       volumes:
       - name: config
@@ -44,5 +48,9 @@ spec:
       - name: job-config
         configMap:
           name: job-config
+      - name: kubeconfig
+        secret:
+          defaultMode: 420
+          secretName: kubeconfig
       nodeSelector:
         prod: prow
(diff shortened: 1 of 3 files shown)


prow/cluster/sinker_deployment.yaml
```diff
@@ -18,19 +18,27 @@ spec:
...
       - name: config
         configMap:
           name: config
       - name: job-config
         configMap:
           name: job-config
+      - name: kubeconfig
+        secret:
+          defaultMode: 420
+          secretName: kubeconfig
```
howardjohn, line 42: where does this secret come from? Do we have some way to set this up (script/docs)? Or is this a magic kubernetes thing that will always be present?

prow/cluster/deck_deployment.yaml
```diff
@@ -30,19 +30,27 @@ spec:
         - --redirect-http-to=prow.istio.io
         - --spyglass
         - --tide-url=http://tide/
+        - --kubeconfig=/etc/kubeconfig/config
```
cjwagner: The data is at the `istio-config` key not `config`.

cjwagner (changes requested):

prow/cluster/sinker_deployment.yaml
clarketm: We have a subset of the tooling and documentation today for *how the contents of* this was created and setup.
* [Run test pods in different clusters](https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#run-test-pods-in-different-clusters)
* [Tool to create cert/key (mkbuild-cluster)](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/mkbuild-cluster)

There is an issue that I am working on to make this a **one** command operation kubernetes/test-infra#13820. I will update the documentation as part of resolving that issue to make this more clear.

fejta: Long term we will build a tool for managing secrets, but don't have one yet. The current process is securely transfer the secret to oncall, who then securely adds it to the cluster.

cjwagner (approved):

PR #1720, opened 2019-08-26
update boskos deployments

clarketm:
Updates to `boskos` deployment and documentation (#1670):
1. Include `ServiceAccount`, `ClusterRole`, `ClusterRoleBinding` definitions in `Deployment`s.
2. Add helpful deployment-related command to `Makefile`
3. Update `README.md`.


boskos/Makefile
```diff
@@ -58,12 +57,22 @@ endif
 		-n $(NAMESPACE) --dry-run -o yaml \
 		| kubectl apply -f -
 
-get-cluster-credentials:
-	gcloud container clusters get-credentials "$(CLUSTER)" --project="$(PROJECT)" --zone="$(ZONE)"
+copy-serviceaccount: get-service-cluster-credentials
```
sebastienvas: THis feels like a one time thing right ?

clarketm: Yes, similar to creating the service account rule, this is a one-time operation. I added a few lines in the [README](https://github.com/istio/test-infra/pull/1720/files#diff-e5d6dd775238fcda214229ed5f1eebfdR31-R40) that mentions that it is for first-time setup of *build* cluster.

prow/plugins.yaml
```diff
@@ -19,9 +19,6 @@ config_updater:
     boskos/resources.yaml:
       name: boskos-config
       namespace: boskos
-    boskos/configs.yaml:
```
clarketm, line 22: `mason-config` has been coalesced into `boskos-config`. Also, `boskos-config` will need to be moved to a postsubmit once the *build* cluster contains boskos instance.

boskos/Makefile
```diff
@@ -14,7 +15,7 @@ mason-image:
 mason-client:
 	go build -o cmd/mason_client/mason_client istio.io/test-infra/boskos/cmd/mason_client
 
-deploy: boskos-deployment cleaner-deployment reaper-deployment janitor-deployment mason-deployment metrics-deployment boskos-service metrics-service
+deploy: get-build-cluster-credentials boskos-deployment cleaner-deployment reaper-deployment janitor-deployment mason-deployment metrics-deployment boskos-service metrics-service
```
cjwagner: The `get-build-cluster-credentials` target should be a dependency of all the individual deployment and service targets instead of the top level `deploy` target so that if I do `make boskos-deployment` the correct cluster is still targeted.

boskos/Makefile
```diff
@@ -14,7 +15,7 @@ mason-image:
...
 
 boskos-deployment:
```
cjwagner: nit: We don't need these per deployment targets. Lets get rid of them to encourage deploying everything.

cjwagner (approved):

fejta (approved):

PR #1723, opened 2019-08-26
Migrate istio/test-infra to "build" cluster

clarketm:
Attempt to validate *build* cluster is functional, #1670 and can run standard presubmits.


howardjohn:
Nice, I don't think the changes are applied until after the PR merges though

clarketm:
@howardjohn - Yes, good point. We can *validate* with a subsequent PR.

geeknoid (approved):

PR #1724, opened 2019-08-26
Update boskos README to reflect latest mason config

clarketm:
Minor semantic updates since `mason-config` has been coalesced into `boskos-config` and is now contained in `resources.yaml` file.

### DIFF
--- a/boskos/README.md
+++ b/boskos/README.md
@@ -39,12 +39,13 @@ boskos, janitor, reaper images are coming from
 so we should only use those images. However for mason, we need to add our own
 resources implementations.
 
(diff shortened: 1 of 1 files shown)


fejta (approved):

howardjohn:
@clarketm fyi seems jobs are not starting

clarketm:
@howardjohn
There is an upgrade (kubernetes/test-infra#13980) to the plank component needed that merged just before yesterday’s Prow bump. Working on getting this in asap. Should unblock presubmits, otherwise will revert.

PR #1728, opened 2019-08-28
Install CMake in /usr/local/bin in prowbazel images.

PiotrSikora:
While there, fix a bunch of non-working mkdir calls.

### DIFF
--- a/docker/prowbazel/Dockerfile
+++ b/docker/prowbazel/Dockerfile
@@ -21,7 +21,7 @@ RUN /tmp/istio_tmp/scripts/linux-install-software \
 VOLUME /var/lib/docker
 EXPOSE 2375
 
-ENV PATH /usr/local/go/bin:/usr/local/cmake/bin:/usr/lib/google-cloud-sdk/bin:/opt/go/bin:${PATH}
+ENV PATH /usr/local/go/bin:/usr/lib/google-cloud-sdk/bin:/opt/go/bin:${PATH}
 
 # Create bootstrap user
 ENV HOME /home/bootstrap
--- a/scripts/tools/linux-install-bazel
+++ b/scripts/tools/linux-install-bazel
@@ -19,7 +19,7 @@ function update_bazel() {
   local bazel_current="$(bazel version | grep -oh 'Bazelisk version: v[0-9]*\.[0-9]*')" \
     || bazel_current='none'
 
-  if [[ -d "${BAZELISK_DIRECTORY}" ]]; then
+  if [[ ! -d "${BAZELISK_DIRECTORY}" ]]; then
     mkdir -p "${BAZELISK_DIRECTORY}"
   fi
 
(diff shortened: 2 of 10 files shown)


docker/prowbazel/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN /tmp/istio_tmp/scripts/linux-install-software \
 VOLUME /var/lib/docker
 EXPOSE 2375
 
-ENV PATH /usr/local/go/bin:/usr/local/cmake/bin:/usr/lib/google-cloud-sdk/bin:/opt/go/bin:${PATH}
+ENV PATH /usr/local/go/bin:/usr/lib/google-cloud-sdk/bin:/opt/go/bin:${PATH}
 
 # Create bootstrap user
 ENV HOME /home/bootstrap
```
fejta, line 27: I'd recommend combining these into a single ENV layer

scripts/tools/linux-install-bazel
```diff
@@ -19,7 +19,7 @@ function update_bazel() {
   local bazel_current="$(bazel version | grep -oh 'Bazelisk version: v[0-9]*\.[0-9]*')" \
     || bazel_current='none'
 
-  if [[ -d "${BAZELISK_DIRECTORY}" ]]; then
+  if [[ ! -d "${BAZELISK_DIRECTORY}" ]]; then
     mkdir -p "${BAZELISK_DIRECTORY}"
```
fejta: why not just run `mkdir -p`? It has the exact same behavior.

fejta (approved):
/hold

docker/prowbazel/Dockerfile
PiotrSikora: Those are separate logical steps (and unrelated change, so I'd rather not do it in this PR).

scripts/tools/linux-install-bazel
PiotrSikora: Good catch, thanks! Fixed.

docker/prowbazel/Dockerfile
fejta: Hmm, well I was all excited to link to the best practices, but it appears that this has changed in the past few years:

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

```
In older versions of Docker, it was important that you minimized the number of layers in your images to ensure they were performant. The following features were added to reduce this limitation:

Only the instructions RUN, COPY, ADD create layers. 
```

fejta (approved):
/hold

fejta:
/hold cancel
I mean

PiotrSikora:
@howardjohn @sdake @geeknoid @duderino @linsun @ericvn this is blocked on `CODEOWNERS` for `/docker/*`, could one of you approve?

howardjohn (approved):

PiotrSikora:
@fejta is this image going to be built and pushed automatically or is it a manual process? I don't see any postsubmits running.

howardjohn:
it is manual, but we should change that

fejta:
We have recipes for doing this, which we can start using now that there is a trusted cluster.

PR #1733, opened 2019-08-28
Update prowbazel in all active proxy branches.

PiotrSikora:


howardjohn (approved):
I pushed the images

PR #1739, opened 2019-08-30
The "build" cluster is now the implicit default.

clarketm:
The "build" cluster is now the implicit default. Removing redundant `cluster` key.

/assign @cjwagner


cjwagner (approved):
Cool. Lets follow this up by removing the `build` context from the kubeconfig.

PR #1749, opened 2019-08-31
Add e2e test of operator for manifest apply

richardwxn:
hold until https://github.com/istio/operator/pull/242 is done


howardjohn (changes requested):
This won't work with the docker image used (it doesn't support kind yet), and you will need the magic volumeMounts for Kind to work (you can look at istio/istio jobs for an example).

See also comments on the operator PR

richardwxn:
not sure why it is being closed by bot and cannot be reopened, I would need to create a new one and continue the effort

PR #1750, opened 2019-09-01
Add jobs for the tests, client-go, and common-files repos.

geeknoid:


howardjohn (approved):
why are we not using the config generation for these? is the only thing the builder image?

prow/cluster/jobs/istio/client-go/istio-client-go.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: client-go-lint
+      branches:
+        - ^master$
+        - ^release-1.*$
+      path_alias: istio.io/client-go
+      decorate: true
+      always_run: true
+      spec:
+        containers:
```
howardjohn, line 13: do we need resource request and limits specified?

geeknoid:
@howardjohn Yes, we should be using the generator tool for these and many others. We just need to add support to specify the build image.

And regarding the resource values, this is all purely cut & paste from other existing jobs. I don't know how to access how much these jobs consume in order to fine tune. Ideally, I suppose we could have a job to check logs and automatically create a PR to fine tune these values :-)

howardjohn:
that would be cool. I forget what happens if you don't specify requests or
limits, we will need to make sure we don't starve the nodes. All these jobs
are pretty light anyways and take a minute or 2 so I'm not too concerned

PR #1752, opened 2019-09-02
Use the config generator for the client-go repo.

geeknoid:
- Add support to the config generator to specify a per-repo build image, as well
as a per-job override for same.

- Add support to suppress the output of the 'entrypoint' command in generated config.

If this looks good, I'll convert several other jobs to this model since it is substantially simpler. Note that I'm still unsure about the resource usage for these jobs. I just copied whatever was being used for the CNI repo.


prow/config/jobs/client-go.yaml
```diff
@@ -0,0 +1,24 @@
+org: istio
+repo: client-go
+image: gcr.io/istio-testing/build-tools:2019-08-31T04-48-30
+suppress_entrypoint: true
```
howardjohn, line 4: todo: remove this as soon as we use build-tools everywhere

howardjohn: Or build tools may need an `entrypoint` anyways

prow/config/jobs/client-go.yaml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  - name: client-go-build
+    command: [make, build]
+  - name: client-go-lint
+    command: [make, lint]
+  - name: client-go-test
+    command: [make, test]
+
+resources:
+  default:
```
howardjohn, line 18: I wonder if we should make this just the implicit default?

howardjohn (approved):

PR #1754, opened 2019-09-03
Use the config generator for most of our job config state.

geeknoid:
- Extend the config generator to have baked-in default resource requirements
to avoid having to repeat that info in every yaml file.


howardjohn (commented):
Nice! before we merge this, I was thinking about making a change to make the generated test names follow the format `name-type-branch` like `istio-test-presubmit-master`. This is because prow requires all job names to be unique. Technically it works with duplicates (as we have now) but many tools break like testgrid (see https://github.com/istio/istio/issues/16721) and the local job running tools.

The only reason I haven't done this yet is since history will be lost for the jobs - but j think we can manually migrate the job history (just copy and paste to the new gcs folder).

Anyhow my point is if we do this before we migrate all of these over it may be easier. Or we can say we don't care too much about the history of these and just do it after. What do you think?

I can probably take on migrating the names this week

geeknoid:
Done. PTAL.

geeknoid:
Er, I didn't mean to close this.

Issue #1402, opened 2019-07-02
Prow job names should be unique


howardjohn:
I cannot run `mkpj --config-path config.yaml --job istio-kind-simpleTest-master --job-config-path cluster/jobs/istio/istio | phaino` because the job is defined in two places, master and prow-staging. It doesn't crash, but it does uses the prow-staging config

See also https://github.com/kubernetes/test-infra/issues/13280

fejta:
Yes please, and I would also recommend eliminating the context: field and just using the job name (default)

Aside: piping to phaino doesn't work super great as it will ask you to input the location of various volumes and secrets.

Instead I would recommend something like 
```console
$  mkpj --job=build-test --config-path prow/config.yaml --job-config-path prow/cluster/jobs > ~/test.yaml && phaino ~/test.yaml
```

howardjohn:
This is done

PR #1755, opened 2019-09-03
Use the config generator to produce most of our job configs.

geeknoid:
- Extend the config generator to understand default resource values,
which avoids repeating them over and over.


prow/config/jobs/istio.io.yaml
```diff
@@ -0,0 +1,11 @@
+org: istio
+repo: istio.io
+image: gcr.io/istio-testing/build-tools:2019-08-31T04-48-30
+suppress_entrypoint: true
+branches:
+  - master
+  - release-1.2
```
howardjohn, line 7: not 1.3 as well? because of the different branching approach I assume?

geeknoid: Correct, there is no 1.3 branch at the moment.

howardjohn:
Can a prow expert confirm that in order to rename a test job (as we are doing here, to remove duplicate names: https://github.com/istio/test-infra/issues/1402), we just need to switch the names and things should work? History for the jobs will be lost, but we can copy the test logs/results over in GCS manually and get them back?

LGTM pending verification on the above ^

prow/cluster/jobs/istio/istio/istio.istio.master.yaml
```diff
@@ -713,7 +713,7 @@ postsubmits:
     branches:
     - ^master$
     decorate: true
-    name: integ-galley-local-postsubmit-tests-master
+    name: integ-galley-local-postsubmit-tests-postsubmit-master
```
howardjohn: can you modify these to not have postsubmit in the original name? I think there is the same for presubmit

geeknoid: Done.

geeknoid:
@howardjohn For the sake of shorter names, how about we tack on -postsubmit for post submit tests, but nothing for presubmit tests (which are the ones that show up in the GitHub UI)

WDYT?

howardjohn:
Yes good idea

howardjohn:
we could also prefix the repo name? So I specify `test` and it becomes
`istio-test-postsubmit-master`. I am neutral about that though, not sure if
its helpful or too much magic

geeknoid:
OK, I've removed "presubmit", but left "postsubmit".

As for tacking on the repo name, that's a good idea IMO. But it'll require going through all the job names and removing their existing repo prefixes, and in the case of the istio/istio jobs, it'll require picking new naming schemes since the generated names end up being too long (limit is 63 chars). So I've left this for later.

howardjohn:
If we are going to change job names, which will require some manual
efforts, it may be better to do it all at once.

geeknoid:
@howardjohn PTAL.

howardjohn:
lgtm!

@clarketm can you advise us on the best way to do a big change of job names like this?

> Can a prow expert confirm that in order to rename a test job (as we are doing here, to remove duplicate names: #1402), we just need to switch the names and things should work? History for the jobs will be lost, but we can copy the test logs/results over in GCS manually and get them back?

clarketm:
> @clarketm can you advise us on the best way to do a big change of job names like this?
> 
> > Can a prow expert confirm that in order to rename a test job (as we are doing here, to remove duplicate names: #1402), we just need to switch the names and things should work? History for the jobs will be lost, but we can copy the test logs/results over in GCS manually and get them back?

@fejta @cjwagner - can one of you confirm that the assumption made by @howardjohn here is correct regarding job renaming and history preservation?

fejta (commented):
Can we put the code changes in their own commit and the generated code change in a separate commit? I'm struggling to find what I'm supposed to review here.

Also, can we get a https://github.com/kubernetes/kubernetes/blob/master/.generated_files so that these PRs are easier to review? This will cause github to default to collapsing diffs in these files (you can still expand if you want to see them).

fejta:
> Can a prow expert confirm that in order to rename a test job (as we are doing here, to remove duplicate names: #1402), we just need to switch the names and things should work?

Yes, with some caveats.

In addition to copying GCS files if you want to preserve history, you'll also need to use the status reconciler to ensure any status contexts get migrated (unless you're explicitly setting the context, which you should not do)

https://github.com/kubernetes/test-infra/tree/master/prow/cmd/status-reconciler

howardjohn:
> .generated_files 

That seems to be for a kubernetes tool not github? But it doesn't seem like we can do it with https://help.github.com/en/articles/customizing-how-changed-files-appear-on-github? Which would be a great improvement

> In addition to copying GCS files if you want to preserve history, you'll also need to use the status reconciler to ensure any status contexts get migrated (unless you're explicitly setting the context, which you should not do)

I believe we have status reconciler set up, as we previous change the status contexts (to no longer set them explicitly), so seems like this should be good to go

fejta:
Yes sorry I meant https://github.com/kubernetes/kubernetes/blob/master/.gitattributes

geeknoid:
Generated yaml files now have the .gen.yaml suffix, and a .gitattributes file now excludes them.

PTAL.

fejta (approved):
Pretty nifty!
/hold
merge at your leisure

howardjohn (approved):

howardjohn:
I would recommend we don't merge this until after 1.3 is out? Last time we did a status reconcile it was pretty slow, and I don't want to block any P0 changes going out

@fejta I recall last time we found the reconciliation was artificially rate limited, can we speed it up?

fejta:
Not safely, no. Otherwise it will essentially cause a prow outage when it runs out of tokens. Was ghproxy being used last time?

Perhaps try changing job contexts more gradually?

Github provides a limited number of tokens per time period, and migrating each token consumes a few tokens, at least three: get old context, push new context, override old context. And github requires this to be done one context on one commit at a time.

howardjohn:
> Not safely, no. Otherwise it will essentially cause a prow outage when it runs out of tokens. Was ghproxy being used last time?

I am not sure if we used ghproxy or not. It seems we don't have token usage stats on https://velodrome.istio.io so not sure if we are anywhere near limits or not anyways. Ideally we could have this run at night (or over the 3 day weekend.. oops) or something and just use elevated amount of tokens.

If not, for now we could revert the changes to istio/istio for now, and roll out gradually? That works for me, we will just need to disable the generate config check?

howardjohn:
This can be done later, but opened https://github.com/istio/test-infra/issues/1762 to improve job UX further.

current config looks great, we agreed offline to merge friday after 1.3 is built

fejta (approved):

PR #1759, opened 2019-09-03
Make branchprotector run as a trusted ProwJob instead of a k8s cronjob.

cjwagner:
I've already added the `oauth-token` secret to the `test-pods` namespace. I'll delete the branchprotector cronjob from the cluster before `/hold cancel`ing this to prevent any races.
ref: #1396 
Based on: https://github.com/kubernetes/test-infra/blob/86fbc418c31dcb5084a89a41d67a59fbbd423369/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L936-L975

/hold


clarketm (approved):

cjwagner:
Whoot!
CronJob deleted and new ProwJob is passing: https://prow.istio.io/view/gcs/istio-prow/logs/ci-test-infra-branchprotector/1168990827033137156

PR #1777, opened 2019-09-07
Use the build-tools container where it makes sense.

geeknoid:
- Use the build tools container for test, lint, and check-config
targets. These CI jobs are now prduced through our config generator
instead of being hand-written.

- Remove deprecated scripts whose functionality is subsumed by
the common files and the build-tools container.


Makefile.core.mk
```diff
@@ -1,21 +1,33 @@
-export GO111MODULE ?= on
```
howardjohn, line 1: no go modules? or is it just in the top level?

Makefile.overrides.mk
```diff
@@ -13,4 +13,4 @@
 # limitations under the License.
 
 # this repo is not yet on the container plan by default
-BUILD_WITH_CONTAINER ?= 0
+BUILD_WITH_CONTAINER ?= 1
```
howardjohn, line 16: this will break a lot of the stuff relating to the prow cluster I think, as it won't mount the kubernetes context. Or those makefiles get the context on demand with gcloud tool which is not in the docker image.

It's possible the makefiles for prow stuff is completely separate and you need to be in the prow/ folder to run it and none of this applies

Makefile.core.mk
```diff
@@ -1,21 +1,33 @@
...
-	@bazel run //:buildifier -- -showlog -mode=check $(git ls-files| grep -e BUILD -e WORKSPACE | grep -v vendor)
+lint: lint-go
+
+test:
+	@go test -race ./...
 
 .PHONY: testgrid
 testgrid:
-	configurator --prow-config prow/config.yaml --prow-job-config prow/cluster/jobs --output-yaml --yaml testgrid/default.yaml --oneshot --output testgrid/istio-generated.yaml
+	@configurator --prow-config prow/config.yaml --prow-job-config prow/cluster/jobs --output-yaml --yaml testgrid/default.yaml --oneshot --output testgrid/istio-generated.yaml
```
howardjohn: this isn't in the container right?

Makefile.core.mk
```diff
@@ -1,21 +1,33 @@
...
 
 generate-config:
-	(cd prow/config/cmd; go run generate.go write)
+	@(cd prow/config/cmd; GOARCH=amd64 GOOS=linux go run generate.go write)
 
 diff-config:
-	(cd prow/config/cmd; go run generate.go diff)
+	@(cd prow/config/cmd; GOARCH=amd64 GOOS=linux go run generate.go diff)
```
howardjohn, line 33: is goarch not set by the container?

howardjohn (commented):
looks good! just a couple questions about turning container build on

Makefile.core.mk
geeknoid: All container-based builds use go modules, we turn that on container-wide.

Makefile.overrides.mk
geeknoid: This setting is only relevant for stuff you run from the command-line using make. Anything run from Bazel is unaffected, anything run form Prow is unaffected.

Makefile.core.mk
geeknoid: GOARCH is set by default to the architecture of the host within the top-level Makefile. So it needs to be overridden, otherwise this ends up produring a mac binary that it tries to run in the Linux VM.

Makefile.overrides.mk
howardjohn: I think there are one-off commands that are run though. under prow/makefile. It may not be included in the top level makefiles though, on which case this would not apply to it

geeknoid:
The lint-check job failure is transient, it'll go away after this change is merged.

Makefile.core.mk
geeknoid: Correct. What is configurator and where does it come from?

Makefile.overrides.mk
geeknoid: Yeah, I don't think that applies. This change only impacts build targets in Makefile.core.mk or anything it imports. Other makefiles are unrelated (although we'll eventually want to merge ther functionality as well so that all build activity can be automated and can be containerized)

Makefile.core.mk
geeknoid: I'm guessing this is the source for this tool:

https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/configurator

geeknoid: I've updated this to build the tool on the fly, but it's not quite working yet. Not clear why...

geeknoid: Works now. We could leave it this way, or we could put configurator into the build container. Not sure if its worth it. Thoughts?

howardjohn (approved):

Makefile.core.mk
```diff
@@ -1,21 +1,37 @@
...
 
 check-config:
-	(cd prow/config/cmd; go run generate.go check)
+	@(cd prow/config/cmd; GOARCH=amd64 GOOS=linux go run generate.go check)
```
fejta, line 36: Why do we need to force the architecture? Seems like this won't work for mac?

fejta (approved):
/hold

Makefile.core.mk
geeknoid: We're building from a container. But we have a (I expect) temporary situation where we force the architecture of stuff built within the container to be the architecture of the host system. In this case, we need to build the binary in the container and run that binary, so we need to force the OS/ARCH to what the container needs.

We also want to support building outside of the container at which point this approach will not work. We'll need a few more variables to control this, but we're not there yet. So for now, we have a few spots across the repos where we force the OS/ARCH so that things work right in the container.

PR #1784, opened 2019-09-09
Disable integration test for installer

howardjohn:
Do not merge until https://github.com/istio/installer/pull/391 is approved


fejta (approved):
/hold

PR #1790, opened 2019-09-10
Make lintmodern_cni required, and retire classic lint_cni.

geeknoid:
I'll eventually rename the job to drop "modern".

### DIFF
--- a/prow/cluster/jobs/istio/cni/istio.cni.master.gen.yaml
+++ b/prow/cluster/jobs/istio/cni/istio.cni.master.gen.yaml
@@ -36,34 +36,6 @@ postsubmits:
     branches:
     - ^master$
     decorate: true
-    name: lint_cni_postsubmit
-    path_alias: istio.io/cni
-    spec:
-      containers:
-      - command:
-        - entrypoint
-        - make
-        - lint
-        image: gcr.io/istio-testing/istio-builder:v20190829-4edc0e70
-        name: ""
-        resources:
-          limits:
-            cpu: "3"
-            memory: 24Gi
-          requests:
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/google/go-github v17.0.0+incompatible
 	github.com/hashicorp/go-multierror v0.0.0-20171204182908-b7773ae21874
 	github.com/kr/pretty v0.1.0
+	github.com/onsi/ginkgo v1.7.0 // indirect
```
fejta: This change seems unrelated?

geeknoid: The tool chain did this change. It's becauee an earlier PR from John removed this since he's testing out go 1.13. Since I'm still on 1.12, that version of the compiler wants those two imports. When we switch everything to 1.13, then we'll refresh all the go.mod files.

fejta: Then please put that into its own PR. It is unrelated to this change.

fejta: Note that bazel avoids these sorts of problems, by declaring -- and using -- a particular go version.

geeknoid:
PTAL.

howardjohn (approved):

rlenglet:
`make lint_modern` fails in the `release-1.3` branch. No PR can be merged into the `release-1.3` branch now, e.g. https://github.com/istio/cni/pull/173#issuecomment-532016382.
Can this be reverted?

rlenglet:
Or is it missing a common files update?

rlenglet:
Updating the common files doesn't fix that: https://github.com/istio/cni/pull/174.

PR #1791, opened 2019-09-10
Postsubmit for deploying boskos and updating boskos config

clarketm:
Postsubmit(s) to deploy boskos and update boskos config.

/hold
> ~~`prow-deployer-service-account` needs to be updated **before** this can be merged.~~

**Update:**
`prow-deployer-service-account` updated with `container.admin` permissions.

### DIFF
--- a/prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
+++ b/prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
@@ -29,6 +29,58 @@ postsubmits:
         secret:
           secretName: prow-deployer-service-account
 
+  - name: post-test-infra-update-boskos-config
+    cluster: test-infra-trusted
+    run_if_changed: '^boskos/resources.yaml$'
+    decorate: true
+    branches:
+      - master
+    annotations:
+      testgrid-create-test-group: "false"
+    spec:
+      containers:
+        - image: gcr.io/k8s-testimages/gcloud-in-go:v20190125-cc5d6ecff3
+          command:
+            - bash
+          args:
+            - -c
+            - |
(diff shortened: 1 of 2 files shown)


prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
```diff
@@ -29,6 +29,58 @@ postsubmits:
...
+      testgrid-create-test-group: "false"
+    spec:
+      containers:
+        - image: gcr.io/k8s-testimages/gcloud-in-go:v20190125-cc5d6ecff3
+          command:
+            - bash
+          args:
+            - -c
+            - |
+              gcloud auth activate-service-account --key-file=/creds/service-account.json && \
```
fejta, line 48: Consider trying to make the command just `make -C boskos boskos-config` (aka skip bash) and set the `env: GOOGLE_APPLICATION_CREDENTIALS=/creds/service-account.json`, which may or may not work

fejta (approved):
/hold

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
clarketm: Okay, Let me give that a shot.

clarketm: It did not seem to work. Maybe it has to do with the fact `make` runs in subshells ... not sure. I would like to get this simplified but can do this in a follow-up.

clarketm: #1794 is a follow-up attempt to clean this up.

PR #1794, opened 2019-09-11
Deploy postsubmit job cleanup / simplification

clarketm:
Cleanup and simplification of *deploy* `postsubmit` jobs:
1. Add `preset-prow-deployer-service-account` preset since its contents are used in several places now.
2. Run `activate-service-account` in Makefile.
3. Extract common gcloud deployment operations into a separate Makefile (`Makefile.gcloud.mk`)

/assign @fejta @cjwagner

### DIFF
--- a/boskos/Makefile
+++ b/boskos/Makefile
@@ -1,14 +1,11 @@
+include ../prow/Makefile.gcloud.mk
+
 PROJECT ?= istio-prow-build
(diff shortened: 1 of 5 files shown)


fejta (approved):
Nice! Long-term we should switch to use the same deployment strategy as upstream.

PR #1796, opened 2019-09-11
Add "boskos-admin" to mason rather than using defaut serviceaccount

clarketm:
Add "boskos-admin" to `mason` rather than using default serviceaccount.

/assign @sebastienvas @krzyzacy 

@sebastienvas @krzyzacy - I have not explicitly verified this. Is this a safe change or does Mason need more privileges?

### DIFF
--- a/boskos/cluster/mason-deployment.yaml
+++ b/boskos/cluster/mason-deployment.yaml
@@ -12,6 +12,7 @@ spec:
       labels:
         app: boskos-mason
     spec:
+      serviceAccountName: boskos-admin
       terminationGracePeriodSeconds: 300
       containers:
       - name: boskos-mason


sebastienvas (approved):

fejta (approved):

fejta:
FYI I expect this to change the identity it uses to access internal kubernetes resources.

I would not expect it to change the identity it uses to access GCP resources -- for this it will still use the default service account associated with the underlying VM.

clarketm:
@fejta 

The impetus for this change was to allow `mason` to perform without having to add `cluster-admin` bindings the the *default* service account in the namespace. 

One of the tasks when setting up the build cluster was to create that binding and I prefer to favor RBAC with sane default perms since we are using it on other resources. I do not desire a change in identity for accessing GCP resources; for that reason I anticipate this change to be safe.

PR #1809, opened 2019-09-14
Add a lintmodern job to istio/istio.

geeknoid:


howardjohn (approved):

howardjohn:
we should probably make this optional. can you watch to ensure it doesnt fail and revert if it does

PR #1810, opened 2019-09-16
Try to get the containers job working on the tools repo.

geeknoid:
- Also, switch most of the jobs in the tools repo to being constructed
by our simplified config model.


prow/config/jobs/tools.yaml
```diff
@@ -0,0 +1,15 @@
+org: istio
+repo: tools
+image: gcr.io/istio-testing/build-tools:2019-09-12T11-19-01
+suppress_entrypoint: true
+branches:
+  - master
+  - release-1.3
```
howardjohn: I don't think this branch exists?

geeknoid: True. I removed it.

I also changed the file permissions used by the config generator to not mark the produced yaml files as executables.

howardjohn (approved):

Issue #1762, opened 2019-09-04
question: simplify rerun_command/trigger


howardjohn:
Prow has two fields:

```
   trigger: "(?m)qux test this( please)?" # Regexp, see discussion.
   rerun_command: "qux test this please"  # String, see discussion.
```

Say we have a job `unit-tests_istio_release-1.3`

Right now a user can `/test all`, `/retest`, `/test unit-tests_istio_release-1.3` and they will all trigger it.

A better UX would just be to type `/test unit-tests`

if we set `trigger: /test unit-tests` will this give us expected outcome? 

Specifically, will it be smart enough to not trigger the `unit-tests_istio_release-1.2` (different release) or `unit-tests_OTHERREPO_release-1.3` tests? And will `retest`/`test all` still work or do we need some crazy regex for those?

clarketm:
@howardjohn - link to or create this issue in [kubernetes/test-infra](https://github.com/kubernetes/test-infra) for this. You will reach a broader audience and get more feedback.

howardjohn:
answered in https://github.com/kubernetes/test-infra/issues/14265

PR #1817, opened 2019-09-17
Simplify rerun and trigger

howardjohn:
Based on answer to https://github.com/istio/test-infra/issues/1762 this will work

### DIFF
--- a/prow/cluster/jobs/istio/bots/istio.bots.master.gen.yaml
+++ b/prow/cluster/jobs/istio/bots/istio.bots.master.gen.yaml
@@ -92,6 +92,7 @@ presubmits:
     decorate: true
     name: build_bots
     path_alias: istio.io/bots
+    rerun_command: /test build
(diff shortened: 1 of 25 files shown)


clarketm (approved):

prow/cluster/jobs/istio/bots/istio.bots.master.gen.yaml
```diff
@@ -136,6 +139,7 @@ presubmits:
           privileged: true
       nodeSelector:
         testing: test-pool
+    trigger: '(?m)^/test (?:.*? )?lint(?: .*?)?$'
```
fejta, line 142: We should delete this and rerun_command. The command for retesting the `lint_bots` context should be straightforward: `/test lint_bots`. This is the default behavior.

@howardjohn @clarketm 

ref https://github.com/istio/test-infra/issues/1833

PR #1834, opened 2019-09-23
Update the operator/istio/api repos to the latest container image.

geeknoid:


prow/config/jobs/operator-1.3.yaml
```diff
@@ -0,0 +1,21 @@
+org: istio
+repo: operator
+image: gcr.io/istio-testing/build-tools:2019-09-17T18-24-15
```
sdake, line 3: is this the correct image?

geeknoid: For 1.3, yes. It's the same image as previously. I had to split the config file in two in order to keep 1.3 on the same image as before.

howardjohn (approved):

PR #1835, opened 2019-09-23
Bump api and client-go prow build image

rcernich:


fejta (commented):
Do all these changes come from somewhere else? Where?

This doesn't seem like the type of thing oncall needs to review.

rcernich:
There were changes made to build-tools and the presubmit prow jobs are failing because they're using an older build-tools image.  This needs to be pushed so the linked PRs can be pushed.

@geeknoid, @sdake, this should fix the presubmit test issues.  It simply updates the image version used.

geeknoid:
@rcernich Couple things. First, I've updated the api repo stuff yesterday so that should be working now, no need for further updates. Second, we don't need to update the common files in this repo as part of this change, I'll do that separately at some point. So update the PR to just deal with the client-go changes and we should be all set.

rcernich:
@geeknoid, let me know if this is correct.  I left the changes for api, as the image version was slightly different (two hours newer).

geeknoid (approved):

PR #1842, opened 2019-09-26
Use the DIND image recently built

sdake:
DIND was added to the build-tools image. This should work well,
although it can be easily reverted if the CNI repository is negatively
impacted.

The DIND PR: https://github.com/istio/tools/pull/373


howardjohn (approved):

howardjohn:
Need to run `make generate-config` first, looks good though

PR #1843, opened 2019-09-26
Retire the old lint job and rename lint_modern to lint.

geeknoid:


prow/config/jobs/istio.yaml
```diff
@@ -7,10 +7,6 @@ branches:
 jobs:
   - name: unit-tests
     command: [make, -e, "T=-v", build, localTestEnv, test, binaries-test]
-  - name: lint
-    type: presubmit
-    resources: lint
```
howardjohn, line 12: we probably still want this? otherwise we are at risk of OOM?

geeknoid:
The lintmodern job has been working fine, right? I'm just retiring the old
lint job here

prow/config/jobs/istio.yaml
geeknoid: The lintmodern job has been working fine, right? I'm just retiring the old
lint job here

howardjohn: Is it working fine because we got lucky or because it actually only uses 3Gb of memory? I doubt we are under 3gb of memory, although we don't pass `-v` anymore so I can't confirm.

The problem is that when the node doesn't have a lot of load, we can go way above our memory requests (3gb) with no problem, but once we get on a loaded node we will get OOM killed. We should identify how much memory this is actually using and make sure we request that much

geeknoid: The lint job lower in the file makes the same resource request, so I think everything is groovy.

howardjohn (approved):

PR #1844, opened 2019-09-26
Use latest dind image

sdake:


howardjohn (approved):
See comments on https://github.com/istio/tools/pull/376

PR #1853, opened 2019-09-27
Use dockerd & approach with CNI

sdake:


howardjohn:
Do we want this on 1.3? I thought just master? Since this will update CNI to golang 1.13, etc that release-1.3 is not on in istio/istio?

sdake:
will resubmit a new pr.

sdake:
done correctly: https://github.com/istio/test-infra/pull/1854

PR #1861, opened 2019-09-29
Job updates.

geeknoid:
- Update all uses of the build-tools container to the latest version.

- Stop generating configs for the 1.3 branch for a few repos that don't have
such a branch.

- Change the config generation rule to delete all generated yaml files before
running generation so as to clean up any config being removed.


prow/config/jobs/release-builder.yaml
```diff
@@ -1,12 +1,12 @@
...
 branches:
-- master
+  - master
 
 jobs:
-- name: lint
-  command: [make, lint]
-- name: test
-  command: [make, test]
+  - name: lint
```
howardjohn, line 9: does yamllint support this check and we don't have it turned on? I don't mind either way but don't want commits flip flopping between indented and not (or our docs!)

howardjohn (approved):

prow/config/jobs/release-builder.yaml
geeknoid: I was surprised to find those two files not formatted in a standard way.

We do yaml linting in this repo, but we do currently have a lot of the yaml style linting options turned off. Cranking up those options is on my todo list once we're fully on the new build-tools plan everywhere

Makefile.core.mk
```diff
@@ -29,6 +29,7 @@ testgrid:
 	@GOARCH=amd64 GOOS=linux go run k8s.io/test-infra/testgrid/cmd/configurator --prow-config prow/config.yaml --prow-job-config prow/cluster/jobs --output-yaml --yaml testgrid/default.yaml --oneshot --output testgrid/istio.gen.yaml
 
 generate-config:
+	@rm -fr prow/cluster/jobs/istio/*/*.gen.yaml
```
fejta, line 32: Can you put this file in a subfolder so it doesn't require root level approval?

fejta (approved):

Makefile.core.mk
geeknoid: There's already a line in CODEOWNERS for that.

PR #1867, opened 2019-09-30
install python 3.7 for benchmark testing

carolynhu:
istio benchmark testing required python 3.7

### DIFF
--- a/docker/istio/shared/tools/install-python.sh
+++ b/docker/istio/shared/tools/install-python.sh
@@ -16,6 +16,7 @@
 
 set -eux
 
+apt-get install python3.7
 apt-get -qqy install python3-pip
 pip3 install --upgrade pip
 pip install --user yamllint


howardjohn:
Can we just not require python 3.7? That doesn't seem like a reasonable requirement considering gLinux doesn't even have 3.7

docker/istio/shared/tools/install-python.sh
```diff
@@ -16,6 +16,7 @@
 
 set -eux
 
+apt-get install python3.7
 apt-get -qqy install python3-pip
```
clarketm: Doesn't this already install python 3?

howardjohn:
I dont think we need this since we have https://github.com/istio/tools/pull/397

carolynhu:
Yes, we do not need this. I'm going to close it.

Issue #1734, opened 2019-08-28
Allow testing of job changes before changes are committed


howardjohn:
Right now it is pretty hard to test changes to jobs. We can test locally with a variety of tools like mkpj, mkpod, phaino, etc, but this is tedious and only gives an approximation of what will run. Using mkpod and running in a GKE cluster is pretty close, but even with this its pretty challenging to run all of the jobs since we have so many.

Ideally when the jobs are changed, the new jobs would run in the PR. The only way I can think of achieving this is https://github.com/kubernetes/test-infra/issues/13370, so this issue may just be essentially tracking that.

howardjohn:
One example where this would be super useful - upgrading golang version. This could break basically any test in some subtle way, so I have no choice but to run a bunch of tests locally (will take hours) or push it out and hope for the best.

howardjohn:
Another case where we wasted a few hours as a result of this https://github.com/istio/istio/pull/16428

geeknoid:
I just wasted a few hours today trying to debug why a PR in istio/istio wasn't working right. Problem is the fact the CI system was configured to using an older version of our linter, as compared to what I was trying to use in my PR. I really wish the CI system's config were co-located in the repo to make this kind of thing harder to screw up.

howardjohn:
Changes like this are impossible to do "right" since they require atomic changes between test-infra and all of our other repos: https://github.com/istio/istio/pull/17797

howardjohn:
Spent a while debugging differences  in local runs and CI in istio/operator today, could have been avoided if we had this

sdake:
> Spent a while debugging differences in local runs and CI in istio/operator today, could have been avoided if we had this

PR in question where lack of atomicity resulted in lost engineering productivity: https://github.com/istio/operator/pull/477

howardjohn:
We had an engineer go through the tools -> common files -> test-infra cycle 3 times because they ran into issues that would not exist if we did this. Granted they should have tested locally, but still would be really nice to not have to jump through hoops

howardjohn:
Another example of this - https://github.com/istio/istio/pull/19133. I need to sync a change across multiple repos.

Is there any update on the status of this?

clarketm:
[...]

PR #1869, opened 2019-09-30
Add entrypoint for installer

howardjohn:
forgot this, needed to set up dockerd


howardjohn:
Would be possible to iterate if we had https://github.com/istio/test-infra/issues/1734

geeknoid (approved):

PR #1871, opened 2019-10-01
Add Slack reporter

### LINKED ISSUE #1643: Report test failures to slack
We current send test failures to google group istio-oncall, but I don't think many look at this. I don't think testgrid supports slack, but we can set up the slack Email app to get an email that will forward to slack, where it will get noticed

We have many daily releases and no one notices

scottilee:
Per https://github.com/kubernetes/test-infra/blob/master/prow/cmd/crier/README.md#slack-reporter this reports test failures and errors to Slack.

Fixes https://github.com/istio/test-infra/issues/1643.

/assign @howardjohn

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -55,6 +55,19 @@ prowjob_namespace: default
 pod_namespace: test-pods
 log_level: debug
 
+slack_reporter:
+  job_types_to_report:
+  - presubmit
+  - postsubmit
+  - periodic
+  - batch
+  job_states_to_report:
+  - failure
+  - error
+  channel: test-failures
+  # The template shown below is the default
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -55,6 +55,19 @@ prowjob_namespace: default
 pod_namespace: test-pods
 log_level: debug
 
+slack_reporter:
+  job_types_to_report:
+  - presubmit
```
geeknoid: I don't think we want to report presubmit failures. That'd be a lot of noise.

fejta: presubmits might be okay if it messaged the person who authored the PR, but that isn't how it works.

scottilee: Removed.

howardjohn (approved):

clarketm (approved):

PR #1877, opened 2019-10-02
Add optional gencheck jobs in a number of repos.

geeknoid:
- The gencheck jobs will be used to ensure all checked-in generated artifacts are
up to date. These are optional until I go through all the repos and add the local
logic they need.

- Add support for the gencheck job in this repo.


prow/config/jobs/istio.yaml
```diff
@@ -221,6 +221,12 @@ jobs:
     resources: lint
     suppress_entrypoint: true
 
+  - name: gencheck
+    command: [make, gen-check]
```
howardjohn, line 225: this target doesn't exist? Even if it is optional it makes it so every PR "fails". If you really want to run this before the make target is merged can you add "hidden" to the modifiers so it doesn't show up in github?

howardjohn (changes requested):

prow/config/jobs/istio.yaml
geeknoid: Good idea. Done.

howardjohn (approved):

geeknoid (commented):
@howardjohn I removed the hidden flag again since the gen-check target has been added in the right places.

clarketm (approved):

PR #1881, opened 2019-10-02
Make the gencheck job required for istio/api and istio/operator

geeknoid:


geeknoid:
On hold until https://github.com/istio/operator/pull/316 is checked in.

PR #1882, opened 2019-10-02
Make gen check optional

howardjohn:
Tests are expected to be left optional for ~1 week before being marked as required. This test is not stable so should not be required: https://prow.istio.io/job-history/istio-prow/pr-logs/directory/gencheck_istio

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
@@ -1500,6 +1500,7 @@ postsubmits:
     decorate: true
     name: gencheck_istio_postsubmit
     path_alias: istio.io/istio
+    skip_report: true
     spec:
       containers:
       - command:
@@ -2658,7 +2659,9 @@ presubmits:
     - ^master$
     decorate: true
     name: gencheck_istio
+    optional: true
     path_alias: istio.io/istio
+    skip_report: true
     spec:
       containers:
       - command:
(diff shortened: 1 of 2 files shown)


geeknoid:
This is different than a flaky test.

The failures are due to sporadic crash in the protoc compiler. It's not going to be getting any better anytime soon. Hiding and making this gate optional basically relegates it to never being used. The validation performed by the gate is valuable, even if it occasionally fails and needs to be retried.

howardjohn:
Adding a test that is broken, especially one with no fix in sight, is not the right move. Flaky tests are not acceptable per code mauve goals -- one of the goals was 95% of PRs are merged without a test flake. That will never happen if we keep this test as is.

If we want to make this required it must be stable -- just like all other tests

geeknoid:
All right, I'm fine making it optional, but I don't want this to be hidden.

Hopefully the flakiness will go away with the next protoc update.

howardjohn:
> All right, I'm fine making it optional, but I don't want this to be hidden.

That is reasonable

Alternatively, we could have it retry on failure a couple times?

geeknoid:
Does prow have the ability to retry, or would this need to be done in the makefile?

howardjohn:
In the makefile, I don't think prow can

fejta:
Kubernetes has a retester job that searches for approved PRs with fail tests and drops a `/retest` comment on the oldest ones. This eliminates the need to babysit approved PRs to deal with flakiness.

This same pattern will work for istio: https://github.com/kubernetes/test-infra/blob/5316362beb67671e2a59120e9578e62ae4d98d23/config/jobs/kubernetes/test-infra/fejta-bot-periodics.yaml#L130-L178

I would also expect bazel to mitigate this kind of flakiness.

And yes, any retries you want to happen inside the container before it exits needs to be encoded into the container (aka needs to be in the Makefile)

PR #1885, opened 2019-10-03
Support client certificate in kubeconfig

mengqiy:
Added option to support creating a GKE cluster with client certificate enabled.
Support creating kubeconfig using client certificate if it is present.

### DIFF
--- a/boskos/gcp/gcloud.go
+++ b/boskos/gcp/gcloud.go
@@ -15,20 +15,97 @@
 package gcp
 
 import (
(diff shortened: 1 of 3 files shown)


fejta:
Seb, can/should this be handled upstream?

sebastienvas:
For now this mason component only exist here.

sebastienvas (approved):

clarketm (approved):

PR #1887, opened 2019-10-03
Add a bunch of optional hidden jobs using build-tools on istio/istio.

geeknoid:


prow/config/jobs/istio.yaml
```diff
@@ -221,6 +221,194 @@ jobs:
...
+  - name: codecov-new
+    modifiers: [optional, hidden]
+    image: gcr.io/istio-testing/build-tools:2019-10-02T14-57-08
+    command: [make, coverage-diff]
+
+  - name: release-test-new
+    modifiers: [optional, hidden]
+    image: gcr.io/istio-testing/build-tools:2019-10-02T14-57-08
+    type: presubmit
+    command: [prow/release-test.sh]
```
howardjohn, line 243: This one won't work, missing some stuff like `fpm`

prow/config/jobs/istio.yaml
```diff
@@ -221,6 +221,194 @@ jobs:
...
+    image: gcr.io/istio-testing/build-tools:2019-10-02T14-57-08
+    command: [prow/e2e-kind-suite.sh, --single_test, e2e_pilotv2_v1alpha3]
+    requirements: [kind]
+
+  # TODO migrate to kind once we have metric server in kind
+  - name: e2e-dashboard-new
+    modifiers: [optional, hidden]
+    image: gcr.io/istio-testing/build-tools:2019-10-02T14-57-08
+    command: [prow/e2e-dashboard.sh]
+    requirements: [gcp]
```
howardjohn, line 349: This one also won't work I think, due to the scripts calling `gcloud` to authenticate. We should try it though, I am not certain

howardjohn (commented):
Should we make them `skipped` as well? then they won't run on every pr, only if you do `/test all`. I just don't want to double the resources we require - this adds like 90 CPU requests or something per PR so its not free

geeknoid:
Can I mark them skipped but not hidden? That is, "don't run these normally, but when they do get run show their results in GitHub"

prow/config/jobs/istio.yaml
geeknoid: I'm expecting nearly every new job to fail. But having all the jobs there, we can tackle issues one by one and gradually gets chunks of jobs to work.

howardjohn:
yes they can be skipped and not hidden. but even if it's hidden you can still click on "pr history" and see all the jobs for one pr - I'm fine either way

geeknoid:
PTAL

howardjohn (approved):

PR #1889, opened 2019-10-05
Add jobs for the bt branch of istio/istio.

geeknoid:
- Instead of having skipped+hidden jobs in master, create normal
jobs that only affect the container branch.


howardjohn (approved):

howardjohn:
you can make them all presubmit only (`type: presubmit`) to probably get around length limit

PR #1895, opened 2019-10-06
Use build tools for release-builder

howardjohn:
Step 1 to unblocking the release stuff, then I think I just need to `go mod download` and things will pass


ericvn (approved):
lgtm

PR #1896, opened 2019-10-07
Remove jobs that have been successfully converted to the build-tools container

geeknoid:


howardjohn (approved):
Another thing that would be easier with https://github.com/istio/test-infra/issues/1734

PR #1900, opened 2019-10-07
Update image for 1.3 tests

richardwxn:
The release test is failing because of older golang verion

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.release-1.3.gen.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.release-1.3.gen.yaml
@@ -21,7 +21,7 @@ postsubmits:
         - localTestEnv
         - test
         - binaries-test
-        image: gcr.io/istio-testing/istio-builder:v20190829-4edc0e70
+        image: gcr.io/istio-testing/istio-builder:v20190926-b1cd57d5
         name: ""
         resources:
           limits:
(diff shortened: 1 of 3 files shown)


howardjohn:
1.3 is supposed to be on golang 1.12

howardjohn:
what is the test failure?

richardwxn:
> what is the test failure?

@howardjohn Here is the link: https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio/17642/release-test_istio_release-1.3/83.

This is also why I created this PR: https://github.com/istio/istio/pull/17657/files trying to handle it for golang 1.12.

If 1.3 supposed to use golang 1.12 and my another PR doesn't work for golang 1.12, is there other way to fix that failing test, which is blocking a 1.3 change.

howardjohn:
its because of istio/cni, someone must have merged the golang 1.13 changes
into the release-1.3 branch @sdake @geeknoid @rlenglet?

geeknoid:
The release-1.3 cni branch should not be using the build-tools container, or the recent updates to common-files. Prow is not configured to use build-tools for the release-1.3 branch.

If build system changes to master have been pulled into the release-1.3 branch, I think we need to undo those changes. release-1.3 should only receive minor patch level changes, not tool updates.

rlenglet:
I guess it's me? Cf. https://github.com/istio/cni/pull/174 which I did because https://github.com/istio/test-infra/pull/1790 broke all builds in `release-1.3`.

howardjohn:
fixed in istio, we were checking out cni at master...

PR #1909, opened 2019-10-08
Add benchmark test to periodic job

richardwxn:


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -218,3 +230,31 @@ periodics:
...
+    containers:
+      - <<: *istio_container_with_kind
+        env:
+          - name: GIT_BRANCH
+            value: master
+          - name: RELEASE_BOT
+            value: foo
+        command:
+          - entrypoint
+          - scripts/get_tools_repo.sh
```
howardjohn: get_tools_repo depends on artifacts of the daily build. We can probably just directly clone. I suggest removing this line and adding:
```
    extra_refs:
   - org: istio
      repo: tools
     base_ref: master
      path_alias: istio.io/tools
```

prow/cluster/jobs/all-periodics.yaml
```diff
@@ -218,3 +230,31 @@ periodics:
...
+          - scripts/run-test.sh
```
howardjohn: this also depends on the artifacts. You can find the latest build at https://storage.googleapis.com/istio-build/dev/latest which points to the last merged commit here: https://gcsweb.istio.io/gcs/istio-build/dev/

howardjohn (changes requested):

prow/cluster/jobs/all-periodics.yaml
richardwxn: @howardjohn thanks for pointing out. Seems like get_tools_repo and run_tests both depend on the SHA from the daily build manifests file? Looking at your suggestion, is adding ref to master of istio.io/istio and istio.io/tools enough or we need to depend on the SHA from here: https://storage.googleapis.com/istio-build/dev/latest?

howardjohn: I think we can run tools from master, but we want to test and record info about a specific PR, so fetching the SHA seems better. But really it should be the same.. master is going to be the same one pointed to by the dev/latest. If you do master you may hit a race though where its been merged but not yet published

howardjohn (commented):
From https://github.com/istio/tools/blob/e97b1eb7981216a5c4741314feadf76acc41a1a3/perf/benchmark/run_benchmark_job.sh#L66 looks like we need to get $TAG from somewhere. I suggest from the dev/latest. But we also need to make the `setup_istio` support pulling from there. I think that should be pretty simple though, just adding a new GCS url

richardwxn:
is it good to go after https://github.com/istio/tools/pull/432 and https://github.com/istio/tools/pull/433 merged?

howardjohn (approved):

PR #1928, opened 2019-10-12
Improve job specs

geeknoid:
- Remove support for suppress_entrypoint. Instead, explicitly specify the entrypoint
command when needed. This is a bit less magic, and simplifies most cases.

- Make the default image be the build-tools image rather than istio-builder, which
means upgrading to a new image requires making one central change.

- Support not specifying any branches in a job config. That defaults it to master.

- Introduce the `branch` command-line option. This will automatically create the job
configs for a release branch. It'll create new job config files which snapshot the
build image. This is paired with the new supports_release_branching field in the job
config, to indicate which sets of jobs to clone for a release. You just invoke the
generator with `go run generate.go branch 1.4" in order to get the 1.4 set of jobs.


prow/config/jobs/istio.yaml
```diff
@@ -1,179 +1,202 @@
...
 repo: istio
-image: gcr.io/istio-testing/build-tools:2019-10-11T13-37-52
-branches:
-  - master
+support_release_branching: true
 
 jobs:
   - name: unit-tests
-    command: [make, -e, "T=-v", build, localTestEnv, racetest, binaries-test]
+    command: [entrypoint, make, -e, "T=-v", build, localTestEnv, racetest, binaries-test]
```
howardjohn, line 7: rather than `entrypoint` we could make this a new `requirement: [docker]` to be more explicit *why* we have it

howardjohn (approved):

clarketm (approved):

Issue #1511, opened 2019-07-18
istio-testing failing to update some branches


fejta:
Looks like the job cannot update istio/community and istio-ecosystem/istio-coredns-plugin:

/cc @geeknoid 

```console

{"component":"branchprotector","error":"update istio: update community: update master from protected=true: get current branch protection: getting branch protection 404: Not Found","file":"prow/cmd/branchprotector/protect.go:127","func":"main.main","level":"error","msg":"0","time":"2019-07-18T18:22:49Z"}
{"component":"branchprotector","error":"update istio-ecosystem: update istio-coredns-plugin: update master from protected=true: get current branch protection: getting branch protection 404: Not Found","file":"prow/cmd/branchprotector/protect.go:127","func":"main.main","level":"error","msg":"1","time":"2019-07-18T18:22:49Z"}
{"component":"branchprotector","file":"prow/cmd/branchprotector/protect.go:129","func":"main.main","level":"fatal","msg":"Encountered 2 errors protecting branches","time":"2019-07-18T18:22:49Z"}
```

cjwagner:
The branchprotector cronjob is still failing with errors like the following:
```yaml
{
  component:  "branchprotector"   
  error:  "update istio-ecosystem: update istio-coredns-plugin: update master from protected=true: get current branch protection: getting branch protection 404: Not Found"   
  file:  "prow/cmd/branchprotector/protect.go:127"   
  func:  "main.main"   
  level:  "error"   
  msg:  "0"   
}
{
  component:  "branchprotector"   
  error:  "update istio: update community: update thisisnotapril-patch-2 from protected=false: get policy: required prow jobs require branch protection"   
  file:  "prow/cmd/branchprotector/protect.go:127"   
  func:  "main.main"   
  level:  "error"   
  msg:  "1"   
}
{
  component:  "branchprotector"   
  error:  "update istio: update community: update hklai-testrelease-group from protected=false: get policy: required prow jobs require branch protection"   
  file:  "prow/cmd/branchprotector/protect.go:127"   
  func:  "main.main"   
  level:  "error"   
  msg:  "1"   
}

geeknoid:
I deleted some stale branches just now on the istio/community repo which should eliminate the two errors coming from that repo.

geeknoid:
For the istio-ecosystem errors, I sent an invitation from istio-ecosystem to have istio-bot join that org. Once the invitation is accepted, that bot can be made an admin and thus will be able to do branch-protection stuff.

[...]

PR #1936, opened 2019-10-15
Add release-1.4 branch protection

howardjohn:
Do not merge until we create the branches

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -104,6 +104,11 @@ blocked_branches: &blocked_branches
     restrictions:
       teams:
       - release-managers-1-3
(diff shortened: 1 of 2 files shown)


howardjohn:
Release branches create, this is needed now

howardjohn:
Note to future: I think the correct move is merging this then creating the branches

clarketm (approved):
FYI, I renamed the PR and added `WIP` to ensure the `wip` label does not get removed. Rename or remove label when you are ready to merge.

---

Also, I fixed `branchprotector` to continue when it encounters an error #1511, so I also think the correct move is merging this then creating the branches.

PR #1943, opened 2019-10-15
Add jobs for trigger warnings

howardjohn:


howardjohn:
See https://github.com/istio/release-builder/pull/22

This can merge before that if we want though, since the test is optional

geeknoid (approved):

PR #1945, opened 2019-10-16
Hide gencheck for operator

howardjohn:
This test has never passed. Hiding in master to avoid confusion, and
disable in 1.4 as it will presumably never be fixed in 1.4.


prow/config/jobs/operator.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
 
   - name: gencheck
     command: [make, gen-check]
-    modifiers: [optional]
+    modifiers: [optional, hidden]
```
geeknoid: We're just about to fix this, so I'd rather not hide it. It's fine to hide it in 1.4 though.

howardjohn:
@geeknoid I left master alone instead

geeknoid (approved):

PR #1947, opened 2019-10-16
Add benchmark jobs for release 1.3, 1.4

richardwxn:


howardjohn (commented):
Lets get the master one passing then add these? Seems to fail still https://prow.istio.io/view/gcs/istio-prow/logs/daily-performance-benchmark/3

mandarjog:
https://github.com/istio/tools/pull/445 should fix the issue.

richardwxn:
@howardjohn should be good to go now?

howardjohn (approved):

howardjohn:
Looks like master passed https://prow.istio.io/view/gcs/istio-prow/logs/daily-performance-benchmark/8

Our boskos config (see `boskos/resources.yaml`) only allows 2 of the perf clusters at once though, so we should stagger the times?

howardjohn:
Also the charts look a bit messed up, looks like the x axis is undefined? https://storage.googleapis.com/istio-build/perf/benchmark_data.20191020-08.d9fd1c24a0e3c3271e75a6e5f23b6895a7f46e51/graphs/istiocpu,3000QPSover240seconds.html

Very cool though

richardwxn:
@howardjohn yep, the graph code part still has some issues so there is some mismatch conditions when many labels created. But anyway @carolynhu has her frontend code to render based on csv data only without using the graph and I would try to fix the graph code later.

ok, I would stagger the trigger time.

howardjohn:
Sounds good, once the time is staggered feel free to remove the hold.
Thanks!

richardwxn:
moved 1.3 job to run at 12:15 pm instead

PR #1949, opened 2019-10-17
Update image for client-go tests

maxbischoff:
Contains new build tools for kubernetes-1.16.2

Required, as tests for https://github.com/istio/client-go/pull/12#issuecomment-543209519 fail due to an older image being used

### DIFF
--- a/prow/config/jobs/client-go.yaml
+++ b/prow/config/jobs/client-go.yaml
@@ -1,6 +1,7 @@
 org: istio
 repo: client-go
 support_release_branching: true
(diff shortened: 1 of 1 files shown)


jasonwzm:
I think you need to update here: https://github.com/istio/test-infra/blob/master/prow/config/generate.go#L47
Please make sure the generated files are not updating 1.4 release configs.

howardjohn:
this is included in https://github.com/istio/test-infra/commit/c4f8346ccc36cda561356cbdd32252328f49dfa9

PR #1953, opened 2019-10-17
Clean up boskos resources

howardjohn:
* Kill `gke-e2e-test-latest`. We can just have every test use `gke-e2e-test`
* Remove VMs - as far as I know they are never used
* Upgrade to k8s 1.13. This is already happening from gke automatically, so having `1.10` here was deceiving 

This is marked as WIP only because (1) will require changes to istio/istio, which I will make if this is approved.

### DIFF
--- a/boskos/cluster/metrics-deployment.yaml
+++ b/boskos/cluster/metrics-deployment.yaml
@@ -18,7 +18,7 @@ spec:
       - name: metrics
         image: gcr.io/k8s-prow/boskos/metrics:v20190730-dbac84bd7
         args:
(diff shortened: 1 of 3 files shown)


sebastienvas (approved):

howardjohn:
@sebastienvas I recall from a while back we had to set `min == max` because something (resource IDs?) hadn't been implemented. Any idea what the status on that is, should we leave it as is?

sebastienvas:
Yes please leave it as is for now.

howardjohn:
Thanks, will remove the WIP on this once I get istio/istio to stop using
the `-latest` resource

clarketm (approved):

PR #1956, opened 2019-10-18
Use prowbazel:0.5.11 in istio/proxy master and release-1.4.

PiotrSikora:


PiotrSikora:
Note: that image isn't published yet... Not sure who has permission to do that.

/wait

howardjohn (approved):
building it now, it is slow though

PR #1964, opened 2019-10-18
Fix lint

howardjohn:


PiotrSikora (approved):
That's one way of doing it! Thanks :D

clarketm (approved):

PR #1969, opened 2019-10-19
Add "Crier" deployment

clarketm:
Add **Crier** deployment with slack reporter.

/assign @michelle192837 
/cc @scottilee

### DIFF
--- a/prow/cluster/crier_deployment.yaml
+++ b/prow/cluster/crier_deployment.yaml
@@ -1,0 +1,46 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: crier
+  labels:
+    app: crier
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: crier
+  template:
+    metadata:
+      labels:
+        app: crier
+    spec:
+      # serviceAccountName: crier
+      terminationGracePeriodSeconds: 30
+      containers:
+      - name: crier
+        image: gcr.io/k8s-prow/crier:v20191018-2ae771e3e
+        args:
+        - --slack-workers=1
+        - --slack-token-file=/etc/slack/token
+        - --config-path=/etc/config/config.yaml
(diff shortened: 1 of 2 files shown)


prow/cluster/crier_deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+        app: crier
+    spec:
+      # serviceAccountName: crier
+      terminationGracePeriodSeconds: 30
+      containers:
+      - name: crier
+        image: gcr.io/k8s-prow/crier:v20191018-2ae771e3e
+        args:
+        - --slack-workers=1
+        - --slack-token-file=/etc/slack/token
```
howardjohn, line 24: Is there any benefits to using crier instead of plank for github reports as well in the future? https://github.com/kubernetes/test-infra/tree/master/prow/crier#migration-from-plank-for-github-report

clarketm: While Crier has its benefits as a pluggable, generic ProwJob reporter, I am not aware of any *benefits* to using Crier over Plank for GitHub reporting ATM. Since Plank GitHub reporting will be removed in the *near future*, we should migrate. I can do this in a follow-up.

fejta: Can you file a `/kind oncall-hotlist` issue to migrate reporting off plank? Otherwise we're likely to remove this without realizing we're breaking istio.

Also, we do not want both crier and plank reporting at the same time. Can you turn off plank reporting before this turns on?

clarketm: #1976 - created issue

This implementation of Crier is specifically for slack reporting, so when this goes in it will not double [report to Github](https://github.com/kubernetes/test-infra/blob/f0f14f5bcd398eab90ec5df74c531237ef802f9c/prow/cmd/crier/README.md#github-reporter).

fejta (approved):
/hold

PR #1972, opened 2019-10-20
Build 1.4 tagged image and 1.5-dev tagged image

sdake:


howardjohn (commented):
do you want the istio release process to build this image or the operator as a post submit? We can't have both with the same tag

sdake:
If the release process overrides env, then there is no need for this PR to do so.  I am not familiar with how the tagging process works for 1.4 branch.  Could you expand?

Cheers
-steve

howardjohn:
The release process is triggered for each istio/istio commit and for the
official builds and RC builds etc. It does set the tags

sdake:
> The release process is triggered for each istio/istio commit and for the official builds and RC builds etc. It does set the tags


If that is the case, should I just close this PR?

sdake:
this looks preoblamatic: 
```
sdake@beast-01:~/go/src/istio.io/operator$ docker pull docker.io/istio/operator:1.4-dev
Error response from daemon: manifest for istio/operator:1.4-dev not found: manifest unknown: manifest unknown
```

sdake:
this looks ok though: https://console.cloud.google.com/gcr/images/istio-testing/GLOBAL/operator?gcrImageListsize=30

sdake:
@howardjohn after looking over this PR and the image generation, it doesn't look like anything needs to be done to test-infra.  May need to change the operator repo to not use `master-latest-daily` but instead use `latest`.  Thoughts?

Cheers
-steve

howardjohn:
I need to send a PR to Istio/release-builder to publish the 1.4-dev and
latest images first, then everything should work

geeknoid (commented):
Could you leave the space between the different jobs? Makes the files considerably easier to read IMO.

howardjohn:
Publishing from release-builder now

PR #1979, opened 2019-10-22
Add a postsubmit job for doc tests.

geeknoid:


howardjohn (approved):
I suspect the istio.io test is likely going to quickly grow to take a very long time. We cannot have long, required tests. For now this is fine, but in the future we need to either make it optional or postsubmit only (or I may be wrong and it doesn't take a long time)

nmittler (approved):

brian-avery:
+1 on the istio.io tests likely eventually taking some time. We have a lot of content on istio.io, we have to wait for containers to deploy, sometimes there's some delays in things like configuration filtering through Pilot....

brian-avery (approved):

PR #1993, opened 2019-10-24
Verify that the release scripts work in istio/proxy presubmits.

PiotrSikora:

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
@@ -98,6 +98,37 @@ presubmits:
       nodeSelector:
         testing: build-pool
 
+  - name: proxy-presubmit-release
+    annotations:
+      testgrid-dashboards: istio_proxy
+    context: prow/proxy-presubmit-release.sh
+    branches: *branch_spec
+    always_run: true
+    decorate: true
+    path_alias: istio.io/proxy
+    timeout: 60m
+    extra_refs:
+    - org: istio
+      repo: istio
+      base_ref: master
+      path_alias: istio.io/istio
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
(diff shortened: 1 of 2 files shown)


kyessenov:
I don't think we should be pushing artifacts in pre-submit. The pushed stuff should be only for committed changes after code review.

PiotrSikora:
> I don't think we should be pushing artifacts in pre-submit. The pushed stuff should be only for committed changes after code review.

We're not going to push artifacts in pre-submit, only test that they can be built.

PiotrSikora:
See: https://github.com/istio/proxy/compare/master...PiotrSikora:presubmit-release?expand=1

kyessenov:
Still not following. I don't think we need to test bash scripts. If they don't work, just fix them when you see a failure. We'll never have robust scripts anyways.

PiotrSikora:
We used to have it, but it was removed while retiring CircleCI jobs in istio/proxy#2475.

The release scripts broke postsubmit (i.e. pushing release artifacts) just 3 commits later, which tells me that we need this check after all...

PiotrSikora:
@kuat can we get this in? We used to have it enabled until a few days ago, and it proven to prevent breakages in post-submits, so I don't understand why are you against it now.

kyessenov (approved):
I don't have a right to approve. There is a security problem that un-authorized PRs will pollute (or override?) artifacts that needs to be addressed.

howardjohn (approved):

PiotrSikora:
Again, the presubmit job is not pushing any artifacts anywhere.

kyessenov:
OK, but the one thing that I have doubts about in the script PR is whether we can push the artifacts :D

PiotrSikora:
Presubmit jobs don't have the permissions to push anywhere, postsubmit job has.

I think that attackers could explicitly load the permissions using `gcloud auth` (i.e. I don't think it's blocked at the Prow level), but that's not new with this job, it's going to be obvious in the offending PRs and presubmit jobs won't accidentally push artifacts.

PR #1995, opened 2019-10-24
Update common files.

geeknoid:
- Also switch istio/tools and istio/test-infra to the new container.


howardjohn (approved):

clarketm (approved):

geeknoid:
@clarketm As is common when upgrading our container, we end up having to do some checkins on faith. The lint and gencheck jobs will fail until the PR goes it, since those checks need the new container that is included in the job spec in this repo.

So if everything looks good, I will force merge this change in.

clarketm:
> @clarketm As is common when upgrading our container, we end up having to do some checkins on faith. The lint and gencheck jobs will fail until the PR goes it, since those checks need the new container that is included in the job spec in this repo.
> 
> So if everything looks good, I will force merge this change in.

This might require a more robust solution ... but an idea at least to prevent the force merge: do we want to disable the lint and gencheck presubmits conditionally on PRs that change common files? Afterall, the changes themselves are validated in the `istio/common-files` repo when they are added right? With an additional test that builds the image with changes from the PR and runs the lint and gencheck tests inside it. 

https://github.com/kubernetes/test-infra/issues/13370

PR #1997, opened 2019-10-25
Drop Boskos cluster amount to 20 (from 50)

howardjohn:
Per
https://velodrome.istio.io/dashboard/db/boskos-dashboard?orgId=1&from=1567260126490&to=1571962688260

We haven't exceeded 20 in 90 days.

We are not using gke-e2e-test on any repo on `master` anymore, so all
usages are from old release branches that are unlikely to ever get an
influx of many PRs at once and go over the limit.

### DIFF
--- a/boskos/resources.yaml
+++ b/boskos/resources.yaml
@@ -1,9 +1,9 @@
 resources:
 - type: gke-e2e-test
   state: dirty
-  # Set back to 10 when mason client supports request ids.
-  min-count: 50
-  max-count: 50
+  # Set min back to 10 when mason client supports request ids.
+  min-count: 20
+  max-count: 20
   needs:
     gcp-project: 1
   config:


clarketm:
Do we want (or need) to reduce the quantity of project resources as well?

@sebastienvas 
How are these projects created and/or destroyed. Is this process automatic or manual? There were a bunch more before ~100 but we cut them from config during build cluster switch. 
https://github.com/istio/test-infra/blob/6ba6857995b2e9374b74d2a499bc62e9a7e36cf4/boskos/resources.yaml#L59-L122

clarketm (approved):

howardjohn:
I have no clue about the projects, defer to @sebastienvas

sebastienvas:
We don't need to reduce the number of projects. We might need to actually delete them from the de console if we don't need them.

sebastienvas (approved):

clarketm:
@sebastienvas
> We don't need to reduce the number of projects. We might need to actually delete them from the de console if we don't need them.

Cool, thanks. I physically need to delete the unused projects from GCP. So basically any GCP projects (e.g. `istio-boskos-60` or greater ... 61, 62, etc.) we need to delete from GCP?

It looks like they go up to ~`istio-boskos-1000`. Are these used anywhere else or are they safe to delete? (I presume they don't cost if they are unused)

PR #1998, opened 2019-10-25
All istio jobs are on kind now

howardjohn:
@geeknoid this pull in a new license file, but the license does not look right...

### DIFF
--- a/licenses/k8s.io/kubernetes/License
+++ b/licenses/k8s.io/kubernetes/License
@@ -1,0 +1,28 @@
+Copyright © 2013 Keith Rarick.
+Portions Copyright (c) 2012 The Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+   * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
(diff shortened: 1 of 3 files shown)


geeknoid:
Yeah, I'm seeing this license file being pulled down in my PR too. I can't repro locally, when I collect the licenses on my system it comes out as LICENSE.

I need to understand this...

geeknoid:
This copyright file comes from the github.com/kubernetes/third_party/forked/godep directory in version 1.13.x and 1.14.x, but not 1.15.x.

Apparently, when I build the test-infra repo through the build-tools container on my Mac, I'm not seeing a dependency on the older k8s repo, so the extra license file isn't brought in. But when building through prow, this older dependency is present and so the license file is pulled in.

It's very worrisome that the CI system is seeing a different set of dependencies than what I'm seeing on my system. The point of using the container to build was to make everything the same.

Anyway, let's go ahead and check this in. I'll figure out why I'm not seeing this dependency when building on my Mac.

geeknoid (approved):

fejta:
Mac operates oddly with mixed casing. AKA I'm pretty sure it considers `License` and `LICENSE` the same file.

Why is licenses/k8s.io/kubernetes/License copywritten to Keith Rarick (as opposed to the kubernetes authors)?

geeknoid:
I figured out the problem with that license file. There are two license files for the package, one is called LICENSE and the other is License. On the Mac, one overwrites the other due to case folding in the file system. But on Linux they are both preserved.

The solution is to change the mirroring feature of the license linter to have it preserve the full path of the original license file within the generated licenses directory. This'll require another rollout of the common files and container, probably next week.

geeknoid:
@fejta Oh, I thought I was clever and figured this all out, but you had already figured it out hours before :-)

fejta (approved):

fejta:
Battle scars! 

We had to rename all the `BUILD` files to `BUILD.bazel` because of:

https://github.com/kubernetes/kubernetes/tree/master/build
https://github.com/kubernetes/kubernetes/blob/master/BUILD.bazel

PR #2004, opened 2019-10-25
Update common files

howardjohn:


geeknoid (approved):

howardjohn:
Would be much easier with https://github.com/istio/test-infra/issues/1734

geeknoid:
This is another case where the gencheck job will fail until this PR is pushed, and then things will work for the next PR.

So @clarketm, if you approve, I'll do a force merge like yesterday.

Thanks

clarketm (approved):

Issue #1985, opened 2019-10-24
bazel cannot create symlinks


kyessenov:
```
WARNING: failed to create one or more convenience symlinks for prefix 'bazel-':
  cannot create symbolic link bazel-bin -> /home/bootstrap/.cache/bazel/_bazel_bootstrap/45ae47b4a0e12d1e81c831ece04d820d/execroot/io_istio_proxy/bazel-out/k8-fastbuild/bin:  /home/prow/go/src/istio.io/proxy/bazel-bin (Permission denied)
```

This appears due to mis-aligned permission model - prowbazel runs as user bootstrap but `~/go` is owned by root.

clarketm:
Slack convo: https://istio.slack.com/archives/C6FCV6WN4/p1571867121100600

related: istio/istio#18195

clarketm:
I surmise that changing the [`prowbazel`](https://github.com/istio/test-infra/blob/5bb1e4bfb1f5bf83c38e5eeb1caec399895ca3c3/docker/prowbazel/Dockerfile#L30) image user to `root` will resolve. I need to ensure this has no adverse side-effects before making the change.

PR #2009, opened 2019-10-25
Promote prowbazel image for istio/proxy master and release-1.4 to 0.5.13

clarketm:
Bump to `prowbazel:0.5.13` in `istio/proxy` `master`/`release-1.4` jobs.

> Image includes fix for `bazel` symlinking #1985

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
@@ -23,7 +23,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.5.12
+      - image: gcr.io/istio-testing/prowbazel:0.5.13
         command:
         - ./prow/proxy-presubmit.sh
         resources:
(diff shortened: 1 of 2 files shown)


PiotrSikora (approved):

prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
```diff
@@ -23,7 +23,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: gcr.io/istio-testing/prowbazel:0.5.12
+      - image: gcr.io/istio-testing/prowbazel:0.5.13
```
fejta, line 26: How long do these jobs take? Seems like we should use RBE to speed things up.

fejta (approved):

prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
PiotrSikora: Most presubmit jobs take ~30mins, presubmit-release and postsubmit take ~2h.

I was thinking about adding RBE to it yesterday, now that it's working again, but if you want to do it yourself and/or if there is some Prow dedicated RBE buildfarm already, then that's great!

PiotrSikora: I filled an issue, so that this wouldn't get lost: #2013

PR #2001, opened 2019-10-25
Update prowbazel to run as "root" user

clarketm:
Fixes #1985

> `bazel` symlink(s) from `root` owned directories (i.e Prow `GOPATH`) are permitted using this version of image: https://prow.istio.io/view/gcs/istio-prow/logs/proxy-postsubmit/788

@PiotrSikora - For versioning the image, I went with a patch update; let me know if you want to use another version #.

cc @kyessenov

PR #2010, opened 2019-10-25
Update prowbazel to install goimports and golangci

bianpengyuan:
This is not quite maintainable though.. @geeknoid @howardjohn have you guys thought about making istio-proxy also use build-tools image? It has some snowflake needs and could increases image size quite a lot. Maybe build prowbazel image from build-tools image?


howardjohn:
> Maybe build prowbazel image from build-tools image?

I think this is what we want ideally

clarketm:
@bianpengyuan
This change will conflict #2001. `0.5.13` is already published. This should be version `0.5.14`.

docker/prowbazel/Makefile
```diff
@@ -1,5 +1,5 @@
 PROJECT = istio-testing
-VERSION ?= 0.5.12
+VERSION ?= 0.5.13
```
clarketm: Same for the other references.
```suggestion
VERSION ?= 0.5.14
```

clarketm (changes requested):

docker/prowbazel/Makefile
bianpengyuan: Ok, I did not see 0.5.13 in the README file so I used it and I think I've already pushed the image with that tag :). Anyway I updated everything to 0.5.14 now.

clarketm: 1) Prow pushes the image on `postsubmit`. I don't think it is good practice to manually push versioned images overwriting ones that already exist. 
2) That is my fault for not updating the README, I am retroactively doing it here #2011. The version was clearly defined in `Makefile` if you were to rebase.
3) I am not sure what tags you pushed, but now `0.5.12`, `0.5.13`, and `0.5.14` all reference the same image. As long as the `0.5.14` you pushed has my changes from #2001 it should be fine ... and nothing uses those old images.
image

bianpengyuan: Oh sorry I don't realize there is a postsubmit that does image push. I thought it was always done manually. I won't push manually, thanks!

Yeah, it should be fine for this updates, only 0.5.14 should be used now and let's merge this PR to make sure the tag has all changes?

docker/prowbazel/README.md
```diff
@@ -56,3 +56,4 @@ Our dependency on this script is because it appropariately writes test job resul
 * 0.5.10: Install cmake in /usr/local/bin
 * 0.5.11: Update clang and libc++ to 9.0 and preload Bazel 1.0.0
 * 0.5.12: Install clang and libc++ from the official releases instead of packages.
+* 0.5.14: Install goimports and golangci-lint
```
fejta, line 60: I'd recommend using bazel to install and manage goimports and golangci-lint for a bazel image

fejta (approved):
/hold
Unhold at your leisure

docker/prowbazel/Makefile
fejta: Can we change the postsubmit to avoid overwriting existing images?

clarketm: @fejta #2014

clarketm: No problem @bianpengyuan . I was more worried about the image tag overwriting than anything else. Can you rebase, then we can merge and be up-to-date.

clarketm (approved):

docker/prowbazel/Makefile
bianpengyuan: Thanks! Rebased.

docker/prowbazel/README.md
bianpengyuan: Ah, thanks for the pointer! I think it makes more sense to use the bazel rules for istio/proxy. When proxy uses an image built from `build-tools`, we can switch to go-lint target from common files. 

But anyway, let's merge this to make everything here in sync with images in gcr, there is no harm to do so.

PR #2020, opened 2019-10-28
Make new installer test required, run in postsubmit

howardjohn:
The reason this was not required before is not because it was flaky, but because it was broken for two reason:
* We were running published images, so we were not even testing the PR really
* We were pulling from `HEAD` of istio/installer, so tests depended on external state that may fail

Both of these are fixed now


clarketm (approved):

howardjohn:
@geeknoid this is the default in 1.4 so I think its important to have this required

geeknoid (approved):
Sorry for the delay, this one slipped through the cracks.

PR #2027, opened 2019-10-31
Stop publishing 1.1 daily builds

howardjohn:
Also delete the job for master, which was apparently triggering PRs to
build the daily builds, but we had already deleted the job to publish
them.


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -109,62 +109,6 @@ periodics:
       - entrypoint
       - scripts/clean_release_requests.sh
 
-- cron: "15 9 * * *" # every day at 1:15 am PST
-  name: release-trigger-daily-build-master
-  branches: master
```
brian-avery: Did you mean to remove the master daily build?

jwendell:
Does it make sense to remove all cronjobs for daily builds on 1.3, 1.2? Do we have commit-based builds for them?

Also, I've noticed there are some commits on 1.1 branch, after 1.1.17: https://github.com/istio/istio/commits/release-1.1

Is 1.1 really dead?

howardjohn:
We still have daily builds for 1.2 and 1.3 so we should keep those around.

One of the commits was by me to change some test infra, I am trying to
reduce the number of dedicated clusters we have running for tests, part of
why I want to kill this job.

The other one on 1.1 - not sure, Josh is the release manager and said there
were no plans to ever release another 1.1 again, and it's already been EOLed

jwendell (approved):

PR #2028, opened 2019-10-31
Stop using the the trigger plugin's trusted_org feature.

cjwagner:
Warning message:
> trusted_org functionality is deprecated. Please ensure your configuration is updated before the end of December 2019.

ref: https://github.com/kubernetes/test-infra/issues/14875

__Effect:__ Previously repos under the `istio-ecosystem` and `istio-releases` orgs allowed any member of the `istio` org to trigger presubmit tests. Now the `istio-ecosystem` repos will only allow `istio-ecosystem` org members to trigger presubmits and similarly for `istio-releases` will only allow `istio-releases` org members.

I'm not sure what kind of impact this will have or who needs to be notified so I'll let Istio engprod/admins merge this at their leasure.
/hold
/assign @clarketm @geeknoid @howardjohn 

ref https://github.com/istio/test-infra/issues/2114

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -2,14 +2,6 @@
 # Keys: Full repo name: "org/repo".
 # Values: List of plugins to run against the repo.
 ---
-triggers:
-- repos:
-  - istio
-  - istio-ecosystem
-  - istio-releases
-  - istio-private
-  trusted_org: istio
-
 config_updater:
   maps:
     prow/config.yaml:


fejta (commented):
@geeknoid @howardjohn any concerns here?

Is it okay to require an istio-ecosystem PR to be from an istio-ecosystem member in order to start presubmits?

Right now we will automatically trust an istio member to be able to run an istio-ecosystem presubmit. After this change they will have to be an explicit org member.

Prow is considering dropping this feature, so if this seems like a hassle / problem we can bring this feedback back and discuss it further (maybe we just disable it for k8s, but leave the functionality in)

fejta (approved):
/hold

howardjohn:
https://github.com/orgs/istio-releases/people has 2 people only, we certainly need the release robot to be part of that org before we make this change

https://github.com/orgs/istio-ecosystem/people is similar, we would probably need to port everyone over from istio into that org as well, but I think that would be ok to do

fejta:
This bot is the worst

clarketm:
Please rebase.

cjwagner:
> Please rebase.

I'm not sure we want to do this. Removing the trusted_org feature is seeming like a bad idea. It is still useful in a number of places and doesn't add unmanageable complexity so I'm not sure why we want to invest effort in deprecating and removing it.

fejta:
> Removing the trusted_org feature is seeming like a bad idea.

Let's debate the existence of the feature in k/ti?

The istio org doesn't need it as peribolos makes it largely irrelevant.

cjwagner:
AFAICT peribolos hasn't been set up to mirror the org membership to `istio-ecosystem` or `istio-releases`. I'll rebase this, but we can't get rid of the trusted_org functionality unless that is addressed.

fejta:
Agree, I'm fine with closing this one until that happens

PR #2029, opened 2019-10-31
Add "genjobs" support for periodics, node selectors, and job type

clarketm:
Add `genjobs` support for:
* **periodic** jobs via the new `-t, --job-type` flag (e.g. `-t presubmits,periodics`).
* node selectors using `--selector` flag (e.g. `--selector testing=build-pool`).
* changed *default* for `-l, --labels` to be empty `[]string`.
* changed *default* for `--bucket` to be `private-build`.

> ℹ️  I will actually **generate** the `istio/istio` jobs once `hook` is configured in the `istio-private` GitHub organization. For `istio/proxy`, I will do the same once istio/proxy#2496 goes through.

### DIFF
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -11,6 +11,19 @@ PROJECT_BUILD ?= istio-prow-build
 ZONE    ?= us-west1-a
 CLUSTER ?= prow
 
+generate-istio-private-jobs:
+	go run ./genjobs \
+		--clean \
+		--mapping istio=istio-private \
+		--job-type postsubmit \
+		--input ./cluster/jobs/ \
+		--output ./cluster/jobs/ \
+		--repo-whitelist istio \
+		--job-whitelist release_istio_postsubmit \
+		--bucket istio-private-build \
+		--cluster private \
+		--selector testing=build-pool
+
 update-config-dry-run: get-cluster-credentials
 	./recreate_prow_configmaps.py \
 		--job-config-dir=cluster/jobs \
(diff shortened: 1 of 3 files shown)


prow/Makefile
```diff
@@ -11,6 +11,19 @@ PROJECT_BUILD ?= istio-prow-build
 ZONE    ?= us-west1-a
 CLUSTER ?= prow
 
+generate-istio-private-jobs:
+	go run ./genjobs \
+	--clean \
```
fejta: I'd indent this deeper than the `go run` so its clear its part of the `go run` command.

prow/Makefile
```diff
@@ -11,6 +11,19 @@ PROJECT_BUILD ?= istio-prow-build
...
+	-m istio=istio-private \
```
fejta: IMO --mapping will be more clear here than -m

prow/genjobs/cmd/genjobs/main.go
```diff
@@ -125,8 +140,8 @@ func validateOrgRepo(o options, org string, repo string) bool {
 }
 
 // validateJob validates that the job passes validation and should be converted.
-func validateJob(o options, name string, patterns []string) bool {
-	if o.jobBlacklist.Has(name) || (len(o.jobWhitelist) > 0 && !o.jobWhitelist.Has(name)) || !isMatchBranch(o, patterns) {
+func validateJob(o options, name string, patterns []string, jType string) bool {
```
fejta, line 143: I'd be inclined to make this `(o options, name, jType string, patterns ...string)`

fejta (approved):
/hold

prow/genjobs/cmd/genjobs/main.go
clarketm: Is the inclination more about having the `slice` param after the `string` param or about explicitly making the `slice` param variadic. Is this a Golang convention; what is the rationale? The argument being passed is always a `[]string` so I do not see an obvious justification for this to be variadic?  But if this is convention or more readable then I do not mind updating.

PR #2035, opened 2019-11-02
add more upgrade periodic jobs

carolynhu:
https://github.com/istio/istio/issues/17950

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -281,7 +281,7 @@ periodics:
 
 # upgrade using helm from 1.2.8 to 1.3.4
 - cron: "0 2 * * *"  # starts every day at 02:00AM UTC
-  name: istio-upgrade-using-helm
+  name: istio-upgrade-using-helm-1.2.8-1.3.4
   branches: master         # this master branch is for prow/upgrade-tests.sh
   decorate: true
   extra_refs:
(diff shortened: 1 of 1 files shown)


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -279,8 +279,58 @@ periodics:
           - entrypoint
           - prow/perf_benchmark_test.sh
 
-- cron: "0 18 * * *"  # starts every day at 18:00PM UTC
-  name: istio-upgrade-helm-chart
+# upgrade using helm from 1.2.8 to 1.3.4
+- cron: "0 2 * * *"  # starts every day at 02:00AM UTC
+  name: istio-upgrade-using-helm
```
howardjohn: job names need to be unique, should probably include from and to in the name

prow/cluster/jobs/all-periodics.yaml
```diff
@@ -279,8 +279,58 @@ periodics:
...
+    testgrid-dashboards: istio_release-pipeline
+    testgrid-alert-email: istio-oncall@googlegroups.com
+    testgrid-num-failures-to-alert: '1'
+  labels:
+    preset-service-account: "true"
+  spec:
+    containers:
+      - <<: *istio_container_with_kind
+        env:
+          - name: SOURCE_TAG
```
howardjohn, line 302: I think we should just always set TARGET_TAG, to be more explicit

howardjohn (commented):
lgtm other than the names

prow/cluster/jobs/all-periodics.yaml
carolynhu: ok, I see, I will add TARGET_TAG for every job

prow/cluster/jobs/all-periodics.yaml
carolynhu: oh, I see.

howardjohn (approved):

PR #2037, opened 2019-11-04
Private Deck instance 

clarketm:
Before merge:
- [x] Create an A record for `prow-private.istio.io`.
- [x] Add `clientID`, `clientSecret`, and `cookieSecret` to cluster.
- [x] ~Upload `private-config` ConfigMap to cluster~ using `--hidden-only` flag.

---

After merge:
- [x] ~Add *postsubmit* job to update `private-config` on changes.~ using `--hidden-only` flag.
- [x] Update SSL cert to include `prow-private.istio.io` #2052 ~#2047~
- [x] Perform a security audit. 
---

After deployed and stable:
- [x] Enable `spyglass`. #2061 
- [x] Enable job `rerun`, will not implement.

### DIFF
--- a/prow/cluster/deck_private_deployment.yaml
+++ b/prow/cluster/deck_private_deployment.yaml
@@ -1,0 +1,94 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deck-private
+  labels:
+    app: deck-private
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: deck-private
+  template:
+    metadata:
+      labels:
+        app: deck-private
+    spec:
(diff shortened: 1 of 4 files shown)


prow/cluster/deck_private_deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: deck-private
+  template:
+    metadata:
+      labels:
+        app: deck-private
+    spec:
+      # serviceAccountName: deck-private
```
cjwagner: Uncomment this.

cjwagner (approved):

PR #2052, opened 2019-11-06
Use GKE managed certs for "prow-private.istio.io"

clarketm:
Since this is an ingress change (which is currently not auto-deployed) and the certificate itself take time to be issued, I went ahead and preconfigured this in the cluster to ensure it was a *safe* change and that it would not cause downtime: 

https://prow-private.istio.io/

Step 1 towards: #1984

/assign @cjwagner

### DIFF
--- a/prow/cluster/cert-manager/tls-ing.yaml
+++ b/prow/cluster/cert-manager/tls-ing.yaml
@@ -9,13 +9,13 @@ metadata:
     certmanager.k8s.io/acme-challenge-type: dns01
     certmanager.k8s.io/acme-dns01-provider: clouddns
     certmanager.k8s.io/cluster-issuer: letsencrypt-prod
+    networking.gke.io/managed-certificates: prow-private-istio-io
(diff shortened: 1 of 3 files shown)


Katharine:
The probe change feels pretty tangential here? Otherwise lgtm, though.

fejta (approved):
/hold

clarketm:
@Katharine 
> The probe change feels pretty tangential here?

The hope was to use the probe as the health check for the L7 GCE load balancer to prevent **502** (due to the fact that `/` on oauth proxy returns non-`200` status code) but the [GKE ingress does not respect this](https://stackoverflow.com/questions/42967763/kubernetes-ingress-gce-keeps-returning-502-error/42971328), so it ends up being tangential although still nice to have 😄. I instead configured this [directly in GCP](https://stackoverflow.com/a/42971328/4279440).

clarketm:
/hold cancel
I will monitor the Prow auto-deployment.

PR #2054, opened 2019-11-06
Use GKE managed cert for "prow.istio.io"

clarketm:
Preparation for transition from cert-manager to GKE cert. By being not being defined in `ingress` this will not cause any complication with TLS or drop webhooks. 

Step 2 towards: #1984

/assign @cjwagner

### DIFF
--- a/prow/cluster/prow-istio-io_managedcertificate.yaml
+++ b/prow/cluster/prow-istio-io_managedcertificate.yaml
@@ -1,0 +1,7 @@
+apiVersion: networking.gke.io/v1beta1
+kind: ManagedCertificate
+metadata:
+  name: prow-istio-io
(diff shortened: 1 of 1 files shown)


Katharine (approved):

Katharine:
/hold

because I'm only about 90% certain this won't somehow break anything even when unused.

oncall should unhold, probably

fejta (approved):

clarketm:
> because I'm only about 90% certain this won't somehow break anything even when unused.

I verified it does *nothing* (also including not creating the cert itself) if not also configured in `annotations` on the `ingress` object. Still agree it should be monitored. 

> **Note:** this is not needed yet. The cert is still good until December #1984

PR #2074, opened 2019-11-09
Genjobs improvements and istio-private job generation

clarketm:
`genjobs` improvements:
- Add `--dry-run` flag to enable dry run mode.
- Add `--ssh-clone` flag to enable ssh cloning.
- Add `--extra-refs` flag to force enable extra refs translation.
- Support presubmit, postsubmit, and periodic *merging*.

Job changes:
- Translate "all" `istio/istio` and `istio/proxy` jobs to `istio-private` org.
- Add `preset-enable-ssh` preset to enable the use of `ssh` in *any* job.
- Add utility script `gen-private-jobs.sh` with job translation instructions.


clarketm:
Need to wait for #2075 to fix `checkconfig` test.

prow/genjobs/cmd/genjobs/main.go
```diff
@@ -403,6 +469,10 @@ func Main() {
 		cleanOutPath(o, o.output)
 	}
 
+	if o.dryRun {
+		return
```
fejta, line 473: Should this just skip `writeOutFile` but otherwise walk through the files? Why are we aborting here?

prow/gen-private-jobs.sh
```diff
@@ -0,0 +1,66 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+COMMON_OPTS=$(
```
fejta:
```bash
common_opts=(
  --mapping=istio=istio-private
  --channel=-
  # ...
)

go run ./genjobs "${COMMON_OPTS[@]}" --env DOCKER_HUB=whatever
```

prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.yaml
```diff
@@ -0,0 +1,211 @@
...
+      ssh_key_secrets:
+      - ssh-key-secret
+    extra_refs:
+    - base_ref: master
+      clone_uri: git@github.com:istio-private/istio.git
+      org: istio-private
+      path_alias: istio.io/istio
+      repo: istio
+    labels:
+      preset-service-account: "true"
```
fejta, line 22: Not needed for this PR but if we're generating the jobs please do not use presets.

clarketm: > Not needed for this PR

Meaning this job does not need a service account? AFAIK it does [activate a SA](https://storage.googleapis.com/istio-prow/logs/proxy-postsubmit/847/build-log.txt) and [uploads build artifacts to GCS](https://github.com/istio/proxy/blob/c5b7d3418847103059acd63858266d769003f522/prow/proxy-postsubmit.sh#L40)

> if we're generating the jobs please do not use presets.

It would still be nice to keep these as human-readable as possible and not perform too many transformations. Do you recommend I expand the `labels` in the job spec? What benefits does this approach have over `presets`?

prow/gen-private-jobs.sh
clarketm: That is much cleaner. Thanks!

prow/genjobs/cmd/genjobs/main.go
clarketm: A true `--dry-run` was not what I was going for here. I can rename this to `--clean-only` and implement a more robust `--dry-run` in a follow-up? My [current need](https://github.com/istio/test-infra/blob/31d7f49abbfd2d794b0ff1cb4a0aa8f75b564581/prow/gen-private-jobs.sh#L34) is that I want to use the `--clean` flag without writing - and I am optimizing time a bit by returning before walking the filesystem (instead of after).

prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.yaml
fejta: It is more explicit, readable and encapsulated in exchange for being more verbose. The verbosity is not a problem given that it is generated.

fejta (approved):

PR #2075, opened 2019-11-09
Disable "long-job-names" warning in checkconfig

clarketm:
This warning is too restrictive.

### DIFF
--- a/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.yaml
+++ b/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.yaml
@@ -19,6 +19,7 @@ presubmits:
         - --exclude-warning=mismatched-tide
         - --exclude-warning=unknown-fields
         - --exclude-warning=non-decorated-jobs
+        - --exclude-warning=long-job-names
 
   - name: flakey-run
     branches:


fejta:
Why do we want/need long job names? What is an example of a job that cannot be reduced in length?

clarketm:
@fejta
This [job](https://github.com/istio/test-infra/blob/046abbeb333e291fc90d20a78964cfcfbd0aae93/prow/cluster/jobs/istio/istio/istio.istio.release-1.4.gen.yaml#L247 ) is at or near the character limit and if I want to add a modifier to the name for a *private* version of the job I cannot.

howardjohn:
We can rename that one if needed

fejta (approved):
/hold
IMO this is unnecessarily long: `e2e-bookInfoTests-envoyv2-v1alpha3_istio_release-1.4_postsubmit`

clarketm:
> IMO this is unnecessarily long: `e2e-bookInfoTests-envoyv2-v1alpha3_istio_release-1.4_postsubmit`

I agree that this job name is *overly verbose*. Conversely, IMO I do not think that a *long* job name(s) should be presubmit blocking unless there is an *actual problem* introduced into the codebase.

fejta:
Either the name is too long and we want a presubmit to enforce this norm or it isn't too long and it's safe to disable the check.

clarketm:
> Either the name is too long and we want a presubmit to enforce this norm or it isn't too long and it's safe to disable the check.

I see value in linting, but unless the linting rule(s) (i.e. character limit) is configurable by the user, then it is difficult to appease all consumer style preferences.

fejta:
Right but we're in agreement here that the name is too long. 

The reason we set it to the length we do is because kubernetes limits the length of labels, which makes jobs with excessively long names harder to find.

clarketm:
> The reason we set it to the length we do is because kubernetes limits the length of labels, which makes jobs with excessively long names harder to find.

I didn’t realize 63 characters was the [limit for `label` names](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). So the intention is error prevention and not linting. 😞 We should revert. 

@howardjohn - FYI, I am going to follow-up with a PR to shorten some of the longer job names so we can stay well within this 63 character hard limit (even after transformation).

PR #2077, opened 2019-11-10
Revert "Disable "long-job-names" warning in checkconfig"

clarketm:
Reverts istio/test-infra#2075


fejta (approved):
/hold

We don't need this check if you don't want it, I was just trying to make the case for why its there.

fejta:
AKA we added it because we had some outages around excessively long job names.

clarketm:
> AKA we added it because we had some outages around excessively long job names.

Knowing this can prevent fatal errors in Kubernetes with Pod creation, it makes this an *attractive* check. Thanks for providing the background and history.

fejta:
We hopefully by now we've already addressed all the issues -- its why we now put the job name in both a label as well as annotation (the annotation has a longer/no limit), but you never know when something will regress, hence the check

PR #2078, opened 2019-11-10
Genjobs: Add "--modifier" flag and truncate job name at 63 characters

clarketm:
Essential `genjobs` tool tweaks per #2075:
- Add `--modifier` flag to customize job/file name modifier.
- Truncate job names at `63` characters. (to prevent Kubernetes errors)


prow/genjobs/cmd/genjobs/main.go
```diff
@@ -295,7 +298,12 @@ func updateEnvs(o options, job *config.JobBase) {
 
 // updateJobBase updates the jobs JobBase fields based on provided inputs to work with private repositories.
 func updateJobBase(o options, job *config.JobBase, orgrepo string) {
-	job.Name = job.Name + jobnameSeparator + modifier
+	maxNameLen := maxLabelLen - len(jobnameSeparator+o.modifier)
+	if len(job.Name) > maxNameLen {
+		job.Name = job.Name[:maxNameLen]
```
fejta, line 303: a) An issue with this is that we might create duplicates, right? Would we want to truncate enough such that we can append a hash of the whole thing?

b) Can we shrink the name of some things? (Maybe priv instead of private, etc)

fejta (approved):
/hold

howardjohn:
We can just rename the jobs to be shorter

prow/genjobs/cmd/genjobs/main.go
clarketm: > a) An issue with this is that we might create duplicates, right? Would we want to truncate enough such that we can append a hash of the whole thing?

Yes, technically we could create duplicates but it is much less likely since I am truncating the *original* job name instead of the *job name + modifier*.

**e.g.**   
```diff
- e2e-bookInfoTests-envoyv2-v1alpha3_istio_release-1.4_postsubmit
+ - e2e-bookInfoTests-envoyv2-v1alpha3_istio_release-1.4_posts_priv
```
So as long as the modifier suffix is unique to the *original* job name there will not be duplicates. That being said, I do like the idea of a hash since it comes with a higher guarantee of uniqueness.

> Can we shrink the name of some things? (Maybe priv instead of private, etc)

While `private` is the *default*, I am setting this on the command-line to be `priv`: 
https://github.com/istio/test-infra/blob/d4ae93fe482a034117479b3590f8c770532edbe3/prow/gen-private-jobs.sh#L30

I am onboard with shortening job names as well (at least for the really long ones). I was hesitant to do that first because job history will be lost (i.e. disassociated with new job name). I think I just need to do a `gsutil` command to remedy this though.

fejta: `gsutil -m` and/or `gsutil rsync` are your friends here.

I've also just had the old and new copies of the jobs for a week/month so there's sufficient history that no one is likely to notice

fejta (approved):

prow/genjobs/cmd/genjobs/main.go
clarketm: Okay, thanks. I have the job rename PR here: #2079. I will script the follow-up to rename the log directory for shortly after that goes in.  I will look into hashing as well.

PR #2079, opened 2019-11-11
Shorten job names

clarketm:
Shorten *overly* verbose job names by a few character:
- e2e-bookInfoTests...~envoyv2-v1alpha3~...

To preserve [job history](https://prow.istio.io/job-history/istio-prow/pr-logs/directory/e2e-bookInfoTests-envoyv2-v1alpha3_istio) for past jobs (before rename), we will need to copy (i.e. `gsutil cp`) from *old* dir to *new* dir. @fejta - any other measure I need to take besides repairing job history so this goes smoothly?

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -371,7 +371,7 @@ postsubmits:
         bucket: istio-private-build
       ssh_key_secrets:
       - ssh-key-secret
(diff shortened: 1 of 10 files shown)


howardjohn (commented):
you can drop the `envoyv2-v1alpha3` part. There hasn't been anything else for over a year

fejta (approved):

clarketm:
We have some logic built into the [`genjobs`](https://github.com/istio/test-infra/tree/master/prow/genjobs) to handle this. If we still want to rename jobs we can reopen.

PR #2081, opened 2019-11-11
Add direct as fallback for GOPROXY

howardjohn:
See https://github.com/golang/go/issues/35516


clarketm (approved):
FYI, obviously being explicit is paramount since *defaults* change, but just so you are aware - this is the *default* value for `GOPROXY` in go1.13.

PR #2082, opened 2019-11-12
Revert "Add direct as fallback for GOPROXY"

howardjohn:
Reverts istio/test-infra#2081

This breaks proxy which is using go 1.12.

for now just revert


PiotrSikora (approved):
Thanks!

clarketm (approved):

PR #2083, opened 2019-11-12
Add a new job to try out proxy-build-tools

bianpengyuan:
add a new optional presubmit job to istio proxy to use the new proxy-builder-tools image. This is just a temporary job to experiment the new docker image. I don't want to block proxy changes on this since it might need some iterations to pass.

cc @PiotrSikora

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.master.yaml
@@ -36,6 +36,38 @@ presubmits:
       nodeSelector:
         testing: build-pool
 
+  - name: proxy-presubmit-new-builder
+    annotations:
+      testgrid-dashboards: istio_proxy
+    context: prow/proxy-presubmit.sh
+    branches: *branch_spec
+    always_run: true
+    decorate: true
+    path_alias: istio.io/proxy
+    timeout: 60m
+    optional: true
+    extra_refs:
+    - org: istio
+      repo: istio
+      base_ref: master
+      path_alias: istio.io/istio
+    labels:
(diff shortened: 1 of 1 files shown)


PiotrSikora (approved):
Change looks good, thanks! But what's the difference vs prowbazel (i.e. why are we doing this in the first place)?

bianpengyuan:
`prowbazel` has lots of stuff in common as the `build-tools` image used by other repos, like golang, gcloud etc. The new proxy builder image is built on top of the build-tools image, so that we don't need to maintain two completely different docker images, and meanwhile we could take advantage of the common build targets like golang-lint, yaml lint etc included in common files (https://github.com/istio/common-files).

howardjohn:
And because it is a superset of the other build tools we can build istio/proxy and istio/istio with the same image

howardjohn (approved):

bianpengyuan:
Looks like the job does not run: https://github.com/istio/proxy/pull/2551 @howardjohn Do I need to add some other config to add the job?

howardjohn:
@bianpengyuan the prow job config is pretty confusing. (almost) every other repo is using this job config generator so the config for a job is as simple as:

```
  - name: istio_e2e_cloudfoundry
    command: [entrypoint, make, e2e_cloudfoundry]
```

We didn't move the proxy over since it had some differences at some point, but we probably can now. 


Anyhow I think the problem is the `context` is overlapping with another test, so both are reported to github as the same name and one of them wins. You can see the job is actually running: https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_proxy/2551/proxy-presubmit-new-builder/1

PR #2091, opened 2019-11-13
Configurable envoy repo in private istio/proxy jobs

clarketm:
Corresponding change to `istio/proxy` is: https://github.com/istio/proxy/pull/2555

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.4.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.4.gen.yaml
@@ -11,6 +11,47 @@ postsubmits:
         bucket: istio-private-build
       ssh_key_secrets:
       - ssh-key-secret
+    labels:
+      preset-enable-ssh: "true"
+      preset-service-account: "true"
+    name: release_istio_release-1.4_postsubmit_priv
+    path_alias: istio.io/istio
+    reporter_config:
+      slack:
+        channel: '-'
+    spec:
+      containers:
+      - command:
+        - entrypoint
+        - prow/release-commit.sh
+        env:
+        - name: DOCKER_HUB
+          value: gcr.io/istio-prow-build
+        - name: GCS_BUCKET
+          value: istio-private-build/dev
+        image: gcr.io/istio-testing/build-tools:release-1.4-2019-11-05T22-47-16
+        name: ""
+        resources:
+          limits:
+            cpu: "8"
+            memory: 24Gi
+          requests:
+            cpu: "5"
+            memory: 3Gi
+        securityContext:
+          privileged: true
+      nodeSelector:
+        testing: test-pool
+  - branches:
+    - ^release-1.4$
+    clone_uri: git@github.com:istio-private/istio.git
+    cluster: private
+    decorate: true
+    decoration_config:
+      gcs_configuration:
+        bucket: istio-private-build
+      ssh_key_secrets:
+      - ssh-key-secret
     name: unit-tests_istio_release-1.4_postsubmit_priv
     path_alias: istio.io/istio
     reporter_config:
@@ -61,42 +102,6 @@ postsubmits:
         - entrypoint
         - make
         - coverage-diff
-        image: gcr.io/istio-testing/build-tools:release-1.4-2019-11-05T22-47-16
-        name: ""
-        resources:
-          limits:
-            cpu: "8"
-            memory: 24Gi
-          requests:
-            cpu: "5"
-            memory: 3Gi
-        securityContext:
-          privileged: true
-      nodeSelector:
-        testing: test-pool
-  - branches:
-    - ^release-1.4$
-    clone_uri: git@github.com:istio-private/istio.git
-    cluster: private
-    decorate: true
-    decoration_config:
-      gcs_configuration:
-        bucket: istio-private-build
-      ssh_key_secrets:
-      - ssh-key-secret
-    labels:
-      preset-service-account: "true"
-    max_concurrency: 5
-    name: release_istio_release-1.4_postsubmit_priv
-    path_alias: istio.io/istio
-    reporter_config:
-      slack:
-        channel: '-'
-    spec:
-      containers:
-      - command:
-        - entrypoint
-        - prow/release-commit.sh
         image: gcr.io/istio-testing/build-tools:release-1.4-2019-11-05T22-47-16
         name: ""
         resources:
@@ -230,7 +235,6 @@ postsubmits:
       - ssh-key-secret
     labels:
       preset-service-account: "true"
-    max_concurrency: 5
     name: e2e-dashboard_istio_release-1.4_postsubmit_priv
     path_alias: istio.io/istio
     reporter_config:
@@ -742,7 +746,6 @@ postsubmits:
       - ssh-key-secret
     labels:
       preset-service-account: "true"
-    max_concurrency: 5
     name: pilot-multicluster-e2e_istio_release-1.4_postsubmit_priv
     path_alias: istio.io/istio
     reporter_config:
@@ -1793,7 +1796,6 @@ presubmits:
       - ssh-key-secret
     labels:
       preset-service-account: "true"
-    max_concurrency: 5
     name: release-test_istio_release-1.4_priv
     path_alias: istio.io/istio
     reporter_config:
@@ -2569,7 +2571,6 @@ presubmits:
       - ssh-key-secret
     labels:
       preset-service-account: "true"
-    max_concurrency: 5
     name: e2e-dashboard_istio_release-1.4_priv
     path_alias: istio.io/istio
     reporter_config:
@@ -2920,7 +2921,6 @@ presubmits:
       - ssh-key-secret
     labels:
       preset-service-account: "true"
-    max_concurrency: 5
     name: pilot-multicluster-e2e_istio_release-1.4_priv
     optional: true
     path_alias: istio.io/istio
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -409,6 +409,20 @@ presets:
       - key: secret
         path: id_rsa
         mode: 0600
+- labels:
+    preset-enable-netrc: "true"
+  volumeMounts:
+  - mountPath: /root
+    name: netrc
+    readOnly: true
+  volumes:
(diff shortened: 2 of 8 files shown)


prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.4.gen.yaml
```diff
@@ -11,7 +11,11 @@ postsubmits:
         bucket: istio-private-build
       ssh_key_secrets:
       - ssh-key-secret
-    name: unit-tests_istio_release-1.4_postsubmit_priv
+    labels:
+      preset-enable-ssh: "true"
+      preset-service-account: "true"
+    max_concurrency: 5
+    name: release_istio_release-1.4_postsubmit_priv
```
howardjohn, line 17: the diff here is interesting, is it not deterministic?

prow/gen-private-jobs.sh
```diff
@@ -41,20 +41,31 @@ go run ./genjobs \
...
 
-# istio/proxy build jobs(s) - postsubmit(s)
+# istio/proxy master build jobs(s) - postsubmit(s)
 go run ./genjobs \
   "${COMMON_OPTS[@]}" \
+  --branches master \
+  --labels preset-enable-netrc=true \
   --job-type postsubmit \
-  --env GCS_BUILD_BUCKET=istio-private-build,GCS_ARTIFACTS_BUCKET=istio-private-artifacts,DOCKER_REPOSITORY=istio-prow-build/envoy \
+  --env GCS_BUILD_BUCKET=istio-private-build,GCS_ARTIFACTS_BUCKET=istio-private-artifacts,DOCKER_REPOSITORY=istio-prow-build/envoy,ENVOY_REPOSITORY=https://github.com/envoyproxy/envoy-wasm,ENVOY_SHA=bc93450483712189f22b2225f76039d5fe1f8ff9,ENVOY_PREFIX=envoy-wasm \
```
howardjohn: in a case where we are updating envoy for  security patch (or even normal sync) the flow would be:

- update istio/envoy
- update sha here
- then trigger istio/proxy update with a dummy change?

howardjohn: when we send a patch to envoy we will get some new SHA which will eventually be the same one as OSS, so we DO actually want to update the istio/proxy directly I would think rather than through here. I think what we want to override is just the source for where we pull that DHA from?

prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.4.gen.yaml
clarketm: I noticed that as well. I can likely sort the array of jobs before writing them to fix this.

prow/gen-private-jobs.sh
clarketm: > when we send a patch to envoy we will get some new SHA which will eventually be the same one as OSS, so we DO actually want to update the istio/proxy directly I would think rather than through here. I think what we want to override is just the source for where we pull that DHA from?

I am not sure exactly what the patch flow is for `envoy`. Since we are now mounting a `.netrc` file with credentials to clone private repos, we can actually specify the archive, sha, etc. here via *environment vars* OR directly in the `WORKSPACE` of the fork. Both are possible. 

https://github.com/istio/test-infra/blob/b36a8c1d76f15ba41b9337ae4cdb77280cc86ea9/prow/gen-private-jobs.sh#L66

howardjohn: I think we can leave the SHA part to the proxy repo, and just override the repo or whatever that contains that Sha. Would that make sense @piotrsikora?

PiotrSikora: Well, we're cloning the repository here, so we need to provide some commit to checkout.

But hardcoding it here feels a bit weird, would we need to update definitions of jobs in public(?) for each update?

howardjohn: We aren't cloning the Envoy repo anywhere here are we? I thought that was done by bazel? So the SHA to checkout would be determined by the WORKSPACE file like today, and we just pass in the override for the repo?

PiotrSikora: No, we're cloning it outside of Bazel, and then pointing Bazel at local checkout.
See: istio/proxy#2555.

Also, the private security fixes committed to istio-private will have a new SHA by definition, so retaining SHA from the `WORKSPACE` wouldn't work anyway

clarketm: @howardjohn 
Here is my proposed change in `istio/proxy`: https://github.com/istio/proxy/pull/2555/files. We pass in `url` (i.e. repo) and `sha`.

howardjohn: I see, I missed that part. What I was thinking is in oss:
1. make changes to Istio/envoy
2. update sha in proxy workspace

so in private it should also be
1. make changes to istio-private/envoy
2. update sha in proxy workspace

that way when everything is done we can directly push all the changes up stream.

Can we have the script read the sha from the workspace?

clarketm: ~I was under the impression that patches to `envoy` would be applied to a private repo (e.g. `istio-private/envoy`) first. Is that not the case? Meaning the change originate in private then makes its way to OSS. If that is not the case then your approach is 👍 .~ I see what you mean. That approach works. we still need t mount the `.netrc`.

PiotrSikora: That's a good point, you're right.

howardjohn: I updated my comment above to be more clear hopefully. Everything should be done in the private org then once embargo lifts we push everything from private to public

PiotrSikora: That's the case, but once embargo is lifted, those commits are going to be pushed to OSS (retaining history and commit SHAs).

clarketm: Updated mine as well. Like I mentioned before, both approaches are possible. I will update the jobs - that seems cleaner than piping env vars (since saving SHA in `WORKSPACE` is going to be pushed upstream to public eventually anyway).

clarketm: One downside (or thing to remember) is that while `SHA` may be the same the `url` (i.e. repo) will need to be changed back to *public* value before merging back to *public* repo (which is not particularly clean). Using env vars does allow us to point to any arbitrary source without code modification. Let me think on this unless you guys have suggestions. I would prefer fully dynamic/automated since it is less error-prone.

howardjohn: If every SHA referenced by istio-private/proxy exists in istio-private/envoy we do not need to change it right? I think that would be the case because we would sync the repos? Unless I am misunderstanding when we get into a situation when we point to something that doesn't exist

clarketm: The `sha` will be the same but the `url` will be different, right?

Instead of:
`https://github.com/istio/envoy/archive/`

It will be:
`https://github.com/istio-private/envoy/archive/`

in `WORKSPACE`. So when we merge back to *public* a human need to amend the workspace commit and make sure the `url` is the *public* one (is what I was getting at).

https://github.com/istio/proxy/blob/0758acc9ec49ae48193a43ff6d86572968ae00e7/WORKSPACE#L51

clarketm: It should not be difficult to remedy. Let's sync up tomorrow and briefly discuss to ensure we are on the same page.

howardjohn: My suggestion was that we keep everything in this PR except the SHA part (`ENVOY_SHA=429ee52b1f45b57420db15f434a36ad9efd81828`) but we still pass the stuff like `ENVOY_REPOSITORY=https://github.com/istio-private/envoy`

Then we keep everything from https://github.com/istio/proxy/pull/2555 except instead of reading `ENVOY_SHA` as an environment variable we grab it from https://github.com/istio/proxy/blob/0758acc9ec49ae48193a43ff6d86572968ae00e7/WORKSPACE#L41

That way everything done in the istio-private/proxy repo will be identical to what we will eventually push to public

PiotrSikora: `WORKSPACE` will always point to the public repo, i.e. `https://github.com/istio/envoy/archive/` and we're going to use `ENVOY_REPOSITORY` to override that in private builds, but the SHA is going to be the same in both repos, so we don't need to override it.

clarketm: Okay, I think we are on the same page. As it stands now with https://github.com/istio/proxy/pull/2555 all of `ENVOY_REPOSITORY, ENVOY_SHA, ENVOY_PREFIX` need to be set for it to work properly. 

What you are suggesting is that we *only* override `ENVOY_REPOSITORY` and `ENVOY_PREFIX` *here* and add custom logic to the [proxy-common.inc](https://github.com/istio/proxy/blob/master/prow/proxy-common.inc) script to read `ENVOY_SHA` value from the `WORKSPACE`? That is doable.

howardjohn: :+1: sounds great

clarketm: https://github.com/istio/proxy/pull/2556

prow/config.yaml
```diff
@@ -409,6 +409,20 @@ presets:
       - key: secret
         path: id_rsa
         mode: 0600
+- labels:
+    preset-enable-netrc: "true"
```
fejta, line 413: We should not generate jobs with presets, please add the secrets and mounts directly to each pod.

clarketm: That is a good suggestion! I will add an option and support for this (e.g. `--expand-presets`). This functionality should be user-specified as to whether to use presets or expand them. This tool is used for Istio GOB (and potentially other places in the future) - places where job *readability* is beneficial and the convenience of a *preset* is justified.

clarketm: Will add this in a follow-up:
issue: https://github.com/istio/test-infra/issues/2096
PR: #2099

howardjohn (approved):

prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.4.gen.yaml
clarketm: Added sorting for env vars: https://blog.golang.org/go-maps-in-action#TOC_7.

fejta (approved):

PR #2092, opened 2019-11-13
Move proxy to prow config generation

howardjohn:
This PR moves the proxy to the new prow config generation. I made this
enabled only for the "prow-stage" branch so we can do send a dummy PR to
validate I didn't break anything.

Benefits:
* Proxy follows same conventions of other repos (in terms of naming,
testgrid, branching, etc)
* Config should be much simpler to understand
* Resolves some issues with duplicate names, confusing context, etc

The only effective changes after this is the change to the job names. I
am open to renaming, the reason I didn't keep the job names as before is
we would have ended up with names like
proxy-postsubmit_proxy_postsubmit, due to the naming convention of
test_repo_jobtype.


howardjohn:
I made this a draft to let the proxy maintainers take a look first

prow/config/jobs/proxy.yaml
```diff
@@ -0,0 +1,43 @@
...
+support_release_branching: true
+image: gcr.io/istio-testing/prowbazel:0.5.16
+node_selector:
+  testing: build-pool
+branches: [prow-stage]
+
+jobs:
+- name: test
+  type: presubmit
+  repos: [istio/istio@master]
```
bianpengyuan: Should these be istio/proxy?

bianpengyuan:
This looks great. Thanks!

prow/config/jobs/proxy.yaml
howardjohn: `repos` are extra repos to checkout, proxy is implicitly checked out at the PR commit for presubmit and the commited PR for for postsubmit.

The proxy jobs all explicitly configured this before, so I assumed they depend on something in istio/istio, is that not needed?

bianpengyuan: I don't think it is necessary. Maybe @PiotrSikora would have better idea. I think it is fine to remove it.

howardjohn: Well this is just for a dummy test branch anyways, so I'll just remove it then add it back later if we need it. No risk of breaking something

bianpengyuan: sg.

geeknoid (approved):

PR #2097, opened 2019-11-14
Remove check-config

howardjohn:
This is no longer needed, we just use gen-check now

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -38,7 +38,4 @@ generate-config:
 diff-config:
 	@(cd prow/config/cmd; GOARCH=amd64 GOOS=linux go run generate.go diff)
 
-check-config:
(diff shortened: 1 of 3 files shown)


clarketm (approved):

howardjohn:
Ah I probably need to split this up/force merge since the job is not deleted until this is merged (ref https://github.com/istio/test-infra/issues/1734)

clarketm:
> Ah I probably need to split this up/force merge since the job is not deleted until this is merged (ref #1734)

If you remove the [actual job](https://github.com/istio/test-infra/blob/1692e583791b20494f4fcd36744a832cd31c7627/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml#L166-L191) first, then once it merges update the code it calls and generates (i.e. this PR) it should work.

geeknoid (approved):

fejta (approved):

PR #2099, opened 2019-11-14
Genjobs support for preset resolution.

### LINKED ISSUE #2096: Support configurable preset expansion in genjobs
Add an option (e.g. `--expand-presets`) and the underlying support for *preset* `label` expansion in `genjobs` to allow *presets* to be resolves at job-generation time. 

https://github.com/istio/test-infra/pull/2091#discussion_r345930987

clarketm:
> This is branched from #2091 and thus should merge after that  PR.

New flags:
- `-p, --presets strings`: Path to file(s) containing additional presets.
- `--resolve`: Resolve and expand values for presets in generated job(s).

New *defaults*:
- `--bucket`: `"private-build"` -> `""`
- `--ssh-key-secret`: `"ssh-key-secret"` -> `""`
- `--cluster`: `"private"` -> `""`
> This new *defaults* scheme allows the tool to be "more generic" and not assume generated jobs are for usage in another *private* cluster. This also allows these values to be set in a global `decoration_config` fields.  

New behavior:
- `--dry-run` now actually prints to stdout the output path and # of jobs (that would have been created).

fixes #2096


howardjohn (approved):

clarketm:
> @clarketm: The following test **failed**, say `/retest` to rerun them all:
> 
> Test name	Commit	Details	Rerun command
> pull-test-infra-prow-checkconfig	[56be3cb](https://github.com/istio/test-infra/commit/56be3cbf7348ea4235c53fdf00d8b85bf82cd803)	[link](https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_test-infra/2099/pull-test-infra-prow-checkconfig/3122)	`/test pull-test-infra-prow-checkconfig`

Same issue as: https://github.com/istio/test-infra/pull/2102#issuecomment-554688287

prow/genjobs/cmd/genjobs/main.go
```diff
@@ -50,66 +51,70 @@ type options struct {
...
+	flag.StringVar(&o.sshKeySecret, "ssh-key-secret", "", "GKE cluster secrets containing the Github ssh private key.")
 	flag.StringVar(&o.modifier, "modifier", defaultModifier, "Modifier to apply to generated file and job name(s).")
-	flag.StringToStringVarP(&o.labels, "labels", "l", map[string]string{}, "Prow labels to apply to the job(s).")
-	flag.StringToStringVarP(&o.env, "env", "e", map[string]string{}, "Environment variables to set for the job(s).")
 	flag.StringVarP(&o.input, "input", "i", ".", "Input directory containing job(s) to convert.")
 	flag.StringVarP(&o.output, "output", "o", ".", "Output directory to write generated job(s).")
-	flag.StringSliceVarP(&_repoWhitelist, "repo-whitelist", "w", []string{}, "Repositories to whitelist in generation process.")
-	flag.StringSliceVarP(&_repoBlacklist, "repo-blacklist", "b", []string{}, "Repositories to blacklist in generation process.")
+	flag.StringSliceVar(&o.branches, "branches", []string{}, "Branch(es) to generate job(s) for.")
+	flag.StringSliceVarP(&o.presets, "presets", "p", []string{}, "Path to file(s) containing additional presets.")
```
cjwagner, line 94: "files or directories" is important to note since this should always point to `config.yaml` and the job config directory.

prow/genjobs/cmd/genjobs/main.go
```diff
@@ -194,6 +203,106 @@ func convertOrgRepoStr(o options, s string) string {
...
+				}
+			}
+
+			job.Spec.Containers[i].VolumeMounts = append(job.Spec.Containers[i].VolumeMounts, volm)
+		}
+	}
+}
+
+// resolvePresets resolves all preset for a particular job Spec based on defined labels.
+func resolvePresets(o options, labels map[string]string, job *config.JobBase, presets []config.Preset) {
```
cjwagner, line 277: Can we reuse this? https://github.com/kubernetes/test-infra/blob/1e74b899028a9f18076f30138846175fd3112600/prow/config/config.go#L1565-L1575

prow/genjobs/pkg/util/os.go
```diff
@@ -25,3 +26,8 @@ func RenameFile(pat string, src string, repl string) string {
 	s := regexp.MustCompile(pat).ReplaceAllString(src, repl)
 	return regexp.MustCompile(`^[^\w\d]`).ReplaceAllString(s, "")
 }
+
+// HasExtension checks if a file's extension matches a pattern.
+func HasExtension(path string, pat string) bool {
+	return regexp.MustCompile(pat).MatchString(filepath.Ext(path))
```
cjwagner, line 32: This will cause a runtime panic if `pat` is not a valid regexp. Please either compile the regexp at init time or return an error.

prow/genjobs/cmd/genjobs/main.go
clarketm: Originally, I was going to create a PR to make that upstream [`resolvePresets(..)`](https://github.com/kubernetes/test-infra/blob/1e74b899028a9f18076f30138846175fd3112600/prow/config/config.go#L1565-L1575) method *public*, and use it. 

Upon starting to make that change, I realized that the version of [`mergePreset(..)`](https://github.com/kubernetes/test-infra/blob/1e74b899028a9f18076f30138846175fd3112600/prow/config/jobs.go#L44:6) that is called does not support *overriding* `env`, `volumes`, of `volumeMounts` values, but instead the signature returns an `error` to handle conflict resolution.

So instead of modifying upstream logic for a specific downstream use-case, I decided to adapt the code. To maximize the robustness for this `genjobs` tool, it needs the ability to mutate and *override* values.

prow/genjobs/cmd/genjobs/main.go
clarketm: yes, `file(s)` is the correct description here.

For `--presets`, this is an arbitrary list of file(s). We explicitly perform an [extension check](https://github.com/istio/test-infra/pull/2099/files#diff-590584282c8f4a9811ba7e4ed678ccebR137-R140) to ensure these are files. Note, this is not limited to a `config.yaml`. A user can add presets to the generated jobs from another file (in another directory or repo) as long as it passes [`ReadJobConfig(path)`](https://github.com/istio/test-infra/pull/2099/files#diff-590584282c8f4a9811ba7e4ed678ccebR215) and has a `yaml` extension. Of course, if defining a `preset` from an external source, `--resolve` will need to be supplied as well, otherwise Prow will not be able to resolve the value of the `preset` at job-creation time.

prow/genjobs/pkg/util/os.go
clarketm: I am using `regexp.MustCompile(..)` in *several* other places, so I will need to refactor the code heavily. I will tackle this issue with a robust solution in a follow-up PR #2110; this diff is already quite large? For now this should be okay because the only pattern being passed in (i.e. [`yamlExt`](https://github.com/istio/test-infra/blob/5602529713f948e6f81f118dc8a845543210b2b7/prow/genjobs/cmd/genjobs/main.go#L44)) for all the current usages is a constant, compilable pattern.

prow/genjobs/cmd/genjobs/main.go
```diff
@@ -50,66 +51,70 @@ type options struct {
...
 	flag.StringSliceVar(&_jobBlacklist, "job-blacklist", []string{}, "Job(s) to blacklist in generation process.")
+	flag.StringSliceVarP(&_repoWhitelist, "repo-whitelist", "w", []string{}, "Repositories to whitelist in generation process.")
+	flag.StringSliceVarP(&_repoBlacklist, "repo-blacklist", "b", []string{}, "Repositories to blacklist in generation process.")
 	flag.StringSliceVarP(&_jobType, "job-type", "t", []string{"presubmit", "postsubmit", "periodic"},
 		"Job type(s) to process (e.g. presubmit, postsubmit. periodic).")
-	flag.StringToStringVarP(&o.orgMap, "mapping", "m", map[string]string{}, "Mapping between public and private Github organization(s).")
+	flag.BoolVar(&o.clean, "clean", false, "Clean output directory before job(s) generation.")
+	flag.BoolVar(&o.dryRun, "dry-run", false, "Run in dry run mode.")
+	flag.BoolVar(&o.extraRefs, "extra-refs", false, "Apply translation to all extra refs regardless of mapping.")
+	flag.BoolVar(&o.resolve, "resolve", false, "Resolve and expand values for presets in generated job(s).")
```
fejta, line 108: Ideally we flip this so that we resolve by default (do we really need an option to not resolve?)

fejta (approved):
/hold

prow/genjobs/gen-private-jobs.sh
```diff
@@ -14,9 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o errexit
```
fejta: This was better (more readable and obvious)

fejta (approved):
/hold

prow/genjobs/gen-private-jobs.sh
clarketm: I have a keyboard macro that generates this using *short opt* flags. If you prefer *long opts* format then I will revise.

clarketm: Updated

prow/genjobs/cmd/genjobs/main.go
clarketm: It may make more sense in the future (as more users adopt this tool) and this ends up being expected behavior to default `--resolve` to `true`. ATM, I prefer to have this as `false` by default for *job succinctness and readability*.

fejta: Yes, IMO this should be the default (or only) behavior. Agree that this is the correct flag name while the default is to not do this.

fejta (approved):

prow/genjobs/cmd/genjobs/main.go
clarketm: > while the default is to not do this.

Does this refer to `--resolve` flag and if so are we in accord? or have I misunderstood and should I change this to `true`?

PR #2106, opened 2019-11-16
Authentikos

clarketm:
`authentikos` is a service used to create/refresh a Google oauth token with a set of scopes and store the value in a Kubernetes secret.


authentikos/README.md
```diff
@@ -0,0 +1,30 @@
+# Authentikos ([αὐθεντικός](https://en.wikipedia.org/wiki/Authentication))
+
+## Description
+
+`authentikos` is a service used to create/refresh a Google oauth token and store the value in a Kubernetes secret.
```
fejta, line 5: a) Why not use https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity?
b) Why not put it here rather than in k/test-infra?

authentikos/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine:3.10.3
```
fejta: A distroless image seems more appropriate here, why do we need alpine?

authentikos/README.md
clarketm: > Why not use https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity?

We actually need to physically pass the oauth token down to the container code. There is a `make` recipe that [performs a `curl` to `https://storage.googleapis.com/`](https://github.com/istio/istio/blob/53d1eefbdd927e723bc4ec4f17065d15c599af80/bin/init.sh#L95) and an access token needs to be made available via the environment as an `Authorization` header. I don't think [workload-identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) can help in this case?

> Why not put it here rather than in k/test-infra?

This service is specifically for Istio jobs that need an oauth token to access storage, so I built it here.

authentikos/Dockerfile
clarketm: > A distroless image seems more appropriate here, why do we need alpine?

What is a "distroless" image?  I went with `alpine` because I wanted something small (to run the binary) and it is only 5MB.

fejta: https://github.com/GoogleContainerTools/distroless

clarketm: 👍 will check that out!

clarketm: 👍 will check that out!

authentikos/README.md
fejta: Please use workload identity to get the access token, rather than relying on a json file that can be stolen.

https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata#verify-concealed

https://cloud.google.com/compute/docs/storing-retrieving-metadata
https://cloud.google.com/compute/docs/instances/verifying-instance-identity#request_signature

clarketm: I do not recall this being a supported auth option when using the Google Api Go client (https://godoc.org/google.golang.org/api). I can look more into it.

fejta: It's the first one: `All the clients in sub-packages support authentication via Google Application Default Credentials`

https://godoc.org/google.golang.org/api#hdr-Authentication_and_Authorization

fejta: If you enable workload identity in your pod the client will automatically detect and use those credentials.

clarketm: Currently, the way this works it that if a `--creds` file is **not** supplied then it will use *Google application default credentials*. That is how I have it [setup](https://github.com/istio/test-infra/pull/2106/files#diff-b7414ef890d889f22c3913f1c18fee2dR166-R186) (or rather what [`transport.Creds(...)`](https://github.com/googleapis/google-api-go-client/blob/006ab4afe25e899e4ec545ce1cbfdaed675d15d8/internal/creds.go#L20:6) does under the hood). It sounds like I just need to enable this mechanism in the nodes; I do not think there is an additional code change to enable this, but I can confirm.

fejta: Sounds great!

authentikos/Dockerfile
clarketm: I am now using a *distroless* docker image!

fejta (approved):
Nice
/hold

IMO this would be better upstreamed to kubernetes/test-infra (now or later). This is essentially the same thing grandmatriarch needs to do, for example. I think it will be generally useful.

authentikos/go.mod
```diff
@@ -0,0 +1,13 @@
+module istio.io/test-infra/authentikos
```
fejta, line 1: Why do we need/want this to be its own module?

authentikos/README.md
clarketm: I was able to verify that in a cluster with **workload-identity** enabled, the Google creds client will pick up the credentials from the metadata server if `--creds` is omitted. 🍾

authentikos/go.mod
clarketm: I added this to optimize the docker build, so that I can both build the binary in the container and so the image only pulls in the necessary dependencies specifically for this code.

clarketm:
> IMO this would be better upstreamed to kubernetes/test-infra (now or later). This is essentially the same thing grandmatriarch needs to do, for example. I think it will be generally useful.

I have no problem porting to k/test-infra if we can reuse. I kinda just pieced this together for an immediate need in Istio. I think there are some more tweaks (i.e. engineering effort and UX) to make this fully reusable, like for `grandmatriarch`. For example, instead of using a `--format formatString` flag it would be great to be able to supply a Go *template* string/file so we can variablize all the dynamic portions needed in the cookie file for `grandmatriarch`. 

As is it designed, it *almost* works as a replacement for `grandmatriarch` but not quite since we cannot variablize the oauth token *more that once* or provide an outside variable like a cookie expiration time. With the necessary changes to make this templating robust, I think many consumers will find this generally useful.

Another potential improvement is multicluster support via `--kubeconfig`. I have not investigated this option yet.

I will create an issue to (a) support more robust formatting (b) migrate this to k/test-infra and (c) replace `grandmatriarch`.

fejta (approved):

PR #2118, opened 2019-11-19
disable RBE in private builds until configured and update .netrc

clarketm:
disable RBE in private builds until configured and update netrc preset to use appropriate `.netrc` path for `prowbazel` image.


howardjohn (approved):

fejta:
Job changes shouldn't involve oncall. Can we update the codeowners file so this doesn't happen?

We only need oncall to approve changes to the services running in the cluster. Job changes should be self service.

fejta (approved):

PR #2119, opened 2019-11-19
update CODEOWNERS so changes to genjobs does not require oncall

clarketm:
Giving `istio/wg-test-and-release-maintainers` and myself ownership over `genjobs` to make all job creation self-service


clarketm:
@fejta - updating `CODEOWNERS` as suggested so job generation can be self-service in the future.

fejta (approved):

PR #2121, opened 2019-11-20
set proper env variables for "istio-private/proxy" presubmits

clarketm:


howardjohn (approved):

clarketm:
@fejta 
Can we get a quick review from you? Also, should/can we update the `CODEOWNERS` so `prow/config.yaml` is also owned by ` istio/wg-test-and-release-maintainers` (or should we move the `presets` to a new file)?

fejta (approved):

PR #2124, opened 2019-11-20
disable "istio-private" slack reporting

clarketm:
disable "istio-private" slack reporting until access can be better controlled.


howardjohn (approved):

clarketm:
I originally created this as a safety net (in case removal was needed). Consumers appear happy with security and utility of the Slack integration. We can always reopen if that is no longer the case

PR #2131, opened 2019-11-22
be more explict with yaml anchor merge

clarketm:
#2129 did not achieve its intended purpose. This is attempt 2.


fejta (approved):
Long term I would like to eliminate the anchors -- we're already generating yaml elsewhere, so there's no need to use yaml footguns.

PR #2132, opened 2019-11-22
Authentikos: add more robust templating support.

### LINKED ISSUE #2127: Authentikos: add more robust templating support.
It would be great to be able to supply a full-featured *template file/string* (e.g. Go template) in Athentikos so we can variables *all* the dynamic portions needed to make this service generic and reusable for more applications.

summary: https://github.com/istio/test-infra/pull/2106#issuecomment-556528394

clarketm:
fix #2127

@fejta - the tempting is now powerful enough such that we can trial this in-place of a `grandmatriarch`. Will follow-up with more unit tests an deployment examples, so that we can introduce this to one of our clusters.

### DIFF
--- a/authentikos/Makefile
+++ b/authentikos/Makefile
@@ -14,7 +14,7 @@
 
 PROJECT = istio-testing
 HUB = gcr.io
-VERSION ?= 0.0.1
+VERSION ?= 0.0.2
 
 .PHONY: deploy
 deploy: image push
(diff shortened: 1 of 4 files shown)


fejta (approved):
How about some unit tests?

clarketm:
> How about some unit tests?

I am working on unit test coverage :smile:. 

---
@fejta - Authentikos is in *prerelease* and has no downstream users. Before it is deployed or advertised, a robust suite of unit tests will be added: #2139.

PR #2136, opened 2019-11-22
enable tide merging for "istio-private"

### LINKED ISSUE #2135: Enable Tide for "istio-private"
Enable Tide for the `istio-private` GitHub organization.

clarketm:
fixes #2135

/hold for verification of `hidden_repos`.

cc @howardjohn

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -306,17 +306,21 @@ tide:
     - istio/release-builder
     labels:
     - "cla: yes"
-    missingLabels:
+    missingLabels: &istio_tide_missing_labels
     - do-not-merge
     - do-not-merge/hold
     - do-not-merge/work-in-progress
+    reviewApprovedRequired: true
+  - orgs:
+      - istio-private
+    missingLabels: *istio_tide_missing_labels
     reviewApprovedRequired: true
   merge_method:
     istio: squash
     istio-releases/pipeline: squash
     istio-ecosystem/authservice: squash
+    istio-private: squash
   target_url: https://prow.istio.io/tide
-
   context_options:
     from-branch-protection: true
     skip-unknown-contexts: true
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -236,7 +236,9 @@ branch-protection:
             master:
               protect: true
     istio-private:
-      required_pull_request_reviews: *protection_required_pull_request_reviews
+      required_pull_request_reviews:
+        <<: *protection_required_pull_request_reviews
+        require_code_owner_reviews: false
       restrictions: *protection_restrictions
       repos:
         <<: *protection_repos
@@ -312,7 +314,7 @@ tide:
     - do-not-merge/work-in-progress
     reviewApprovedRequired: true
   - orgs:
-      - istio-private
+    - istio-private
     missingLabels: *istio_tide_missing_labels
     reviewApprovedRequired: true
   merge_method:


clarketm:
After looking at the code, speaking with @cjwagner, learning of accounts of this in production, and performing some *functional* tests is an isolated environment, this is a *safe* change.

- [x] I suggest we sync Istio GitHub labels to `istio-private` before merging, otherwise it will be difficult/impossible for a user to block a merge with one of the specified labels.

cjwagner:
> /hold for verification of `hidden_repos`.

@clarketm FYI, the slash commands provided by Prow plugins must be specified at the start of their own line. Only parameters should be on the same line as the command for commands that accept parameters.

clarketm:
> > /hold for verification of `hidden_repos`.
> 
> @clarketm FYI, the slash commands provided by Prow plugins must be specified at the start of their own line. Only parameters should be on the same line as the command for commands that accept parameters.


Got it, thanks! This was more for semantics in this case, as I added the label manually. But it is good info to know!

howardjohn (approved):

prow/config.yaml
```diff
@@ -306,17 +306,20 @@ tide:
...
     - "cla: yes"
-    missingLabels:
+    missingLabels: &istio_tide_missing_labels
     - do-not-merge
     - do-not-merge/hold
     - do-not-merge/work-in-progress
     reviewApprovedRequired: true
+  - orgs:
+      - istio-private
+    missingLabels: *istio_tide_missing_labels
```
cjwagner, line 318: This config will merge any PR in the org as soon as the tests are passing if there isn't a `do-not-merge/*` label.  In other words Tide won't wait for code review.

If you include `reviewApprovedRequired: true`, Tide will wait for an approve review before merging. However, Tide uses the search token `review: approved` to search for approved PRs. I don't think that it matches PRs that have only been approved by non-CODEOWNERS though so until we deal with the non-applicable CODEOWNERS file we probably can't use that.

If we don't want to worry about the CODEOWNERS problem it might be easiest to use the `lgtm` plugin configured to respect github approve reviews and have Tide require that label?

cjwagner (changes requested):

prow/config.yaml
clarketm: > This config will merge any PR in the org as soon as the tests are passing if there isn't a `do-not-merge/*` label. In other words Tide won't wait for code review. If you include `reviewApprovedRequired: true`, ...

Good catch, I updated to include `reviewApprovedRequired: true`.

> Tide will wait for an approve review before merging. However, Tide uses the search token review: approved to search for approved PRs. I don't think that it matches PRs that have only been approved by non-CODEOWNERS though so until we deal with the non-applicable CODEOWNERS file we probably can't use that.

> If we don't want to worry about the CODEOWNERS problem it might be easiest to use the lgtm plugin configured to respect github approve reviews and have Tide require that label?

The current situation is, in fact, a *non-applicable CODEOWNERS file* scenario. I am looking into the GitHub query (at past, approved PRs) to see what happens in the case of an approval from a non-CODEOWNER. If needed I will take your suggestion and add the `lgtm` plugin (and label), Thanks!

cjwagner (approved):
Tide probably won't merge anything until the CODEOWNERS thing is addressed, but this should now be safe to merge during oncall hours.

prow/config.yaml
howardjohn: I think you can actually turn off "require codeowners approval". Then anyone can approve even if they aren't a codeowners, which is probably relevant here

clarketm: > I think you can actually turn off "require codeowners approval". Then anyone can approve even if they aren't a codeowners, which is probably relevant here

Good idea. We should try this first. Let me change that in the `branch-protection` settings and update the PR.

clarketm:
> Tide probably won't merge anything until the CODEOWNERS thing is addressed, but this should now be safe to merge during oncall hours.

I updated the PR to disable CODEOWNERS in branch priotection setting (i.e. `require_code_owner_reviews: false`). I also verified that `review:approved` GitHub query matches the relevant PRs that are approved.

cjwagner (approved):

fejta (approved):

PR #2138, opened 2019-11-23
autodeploy private cluster resources

clarketm:
Enable the auto-deployment of *private* cluster resources.

### DIFF
--- a/prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
+++ b/prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
@@ -3,7 +3,7 @@ postsubmits:
   istio/test-infra:
   - name: post-test-infra-deploy-prow
     cluster: test-infra-trusted
-    run_if_changed: '^prow/cluster/(?:build/)?[^/]+\.yaml$'
+    run_if_changed: '^prow/cluster/(?:build/|private/|[^/]+\.yaml$)'
     decorate: true
     branches:
     - master
@@ -21,6 +21,7 @@ postsubmits:
         - prow
         - deploy
         - deploy-build
+        - deploy-private
 
   - name: post-test-infra-update-boskos-config
     cluster: test-infra-trusted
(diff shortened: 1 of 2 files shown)


clarketm:
/hold
For verification that the `prow-deployer` SA has the access it needs.

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
```diff
@@ -3,7 +3,7 @@ postsubmits:
   istio/test-infra:
   - name: post-test-infra-deploy-prow
     cluster: test-infra-trusted
-    run_if_changed: '^prow/cluster/(?:build/)?[^/]+\.yaml$'
+    run_if_changed: '^prow/cluster/(?:build/|private/|[^/]+\.yaml$)'
```
cjwagner, line 6: nit: This will run for changes to non-YAML files in the `build` and `private` directories.

cjwagner: Should the regexp match the Makefile? Maybe we want this to run when anything in these directories changes regardless of extension?

cjwagner (approved):
Just some nits and questions feel free to ignore.
/hold

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
clarketm: > nit: This will run for changes to non-YAML files in the build and private directories.

The goal is to reserve the `build/` and `private/` as dedicated directories for Kubernetes resources. At least that is how we are using those directories currently.

> Should the regexp match the Makefile? 

Because the Makefile has the `gen-private-jobs` target, I am worried that this will trigger unexpected deployments.

> Maybe we want this to run when anything in these directories changes regardless of extension?

The [jobs](https://github.com/istio/test-infra/tree/master/prow/cluster/jobs) are in this directory too; we will want to *exclude* those.

PR #2141, opened 2019-11-24
Add presubmit docker check

howardjohn:
This verifies that the build won't be broken

Also move the postsubmit over to the gen code.

Depends on https://github.com/istio/tools/pull/572


sdake (commented):
LGTM once dependent PR merges.

geeknoid (approved):

PR #2142, opened 2019-11-24
Genjobs: fix incorrect behavior with --extra-refs

clarketm:
- Fix incorrect behavior with `--extra-refs` (before unknown orgs mappings would be converted to *empty* string).
- Add supporting unit tests.


clarketm:
@howardjohn @jwendell @fejta - PTaL. This is mostly tests with a minor bug fix.

howardjohn (approved):

Issue #2140, opened 2019-11-23
Add TestGrid Presubmit Tests


chases2:
A reference to a non-existent dashboard [link](https://github.com/istio/test-infra/pull/2137) caused the transfigure job to [fail](https://prow.istio.io/view/gcs/istio-prow/logs/post-test-infra-upload-testgrid-config/1190349260122492995).

Instead, a presubmit similar to [kubernetes](https://github.com/kubernetes/test-infra/tree/master/config/tests/testgrids) should prevent this from merging.

chases2:
Known issue: `post-test-infra-upload-testgrid-config` is failing. This is because the postsubmit now explicitly fails when the presubmit would fail, instead of pushing a bad PR.

Presubmit is in place (as an optional test, since it's failing) and running. [First example](https://github.com/istio/test-infra/pull/2189)

I'm working on making the necessary config changes to make this test pass.

chases2:
Test now passes. Postsubmit isn't pushing to k8s; looking into that. Last push was [2 days ago](https://github.com/kubernetes/test-infra/pull/15148), despite the job running since then (see postsubmits of #2190)

chases2:
Postsubmit has been working pretty well recently; most recent push was [4 hours ago](https://github.com/kubernetes/test-infra/pull/15554)

PR #2150, opened 2019-11-26
Addition of io_k8s_repos and removal of vendor deps

chases2:
Allows use of @io_k8s_repo_infra//hack commands in bazel
Also runs automatic cleanup/update command
@io_k8s_repo_infra//hack:update-deps

Useful cleanup towards #2140


michelle192837:
This looks good to me (once it passes tests), but will leave approval to people who contribute more often.

testgrid/istio.gen.yaml
```diff
@@ -1,5509 +0,0 @@
-dashboard_groups:
```
fejta: Is this intentional?

chases2: No; this change should be in a different PR. This is what I was working on before

chases2:
Update: ran `make gen`
Keeping as different commits to illustrate the difference between `make gen` and `bazel run @io_k8s_repo_infra//hack:update-deps`

fejta:
Do we still need this?

chases2:
I don't think we _need_ this, but the next person who comes by to try and use Bazel will certainly solve the same problem.

Giving Bazel a ref to the `io_k8s_repo_infra` makes using Bazel much easier, and I don't think it costs other workflows (ie. make) anything.

chases2:
TL;DR I would still like to merge this

fejta:
Will this pass unit tests? aka `bazel test //...` works?

chases2:
No, although `bazel test //...` didn't pass before, either.

Both with and without changes, `bazel test //...` is failing with similar errors:
Before:
```
ERROR: /usr/local/google/home/slchase/repositories/istio-test-infra/prow/config/BUILD.bazel:29:1: no such package 'vendor/github.com/ghodss/yaml': BUILD file not found on package path and referenced by '//prow/config:go_default_library'
ERROR: Analysis of target '//prow/config:go_default_library' failed; build aborted: no such package 'vendor/github.com/ghodss/yaml': BUILD file not found on package path
```
After:
```
ERROR: /usr/local/google/home/slchase/repositories/istio-test-infra/prow/config/BUILD.bazel:8:1: no such package 'prow/cluster': BUILD file not found in any of the following directories.
 - /usr/local/google/home/slchase/repositories/istio-test-infra/prow/cluster and referenced by '//prow/config:go_default_test'
ERROR: Analysis of target '//prow/config:go_default_test' failed; build aborted: no such package 'prow/cluster': BUILD file not found in any of the following directories.
```

chases2:
What this does let us do is run `@io_k8s_repo_infra//hack:update-deps`, `@io_k8s_repo_infra//hack:update-bazel`, etc.

fejta (approved):
/hold
/assign @clarketm @howardjohn 
fyi

PR #2159, opened 2019-11-27
update "release-builder" jobs w/ env overrides

clarketm:
Dependent on changes to`istio/release-builder` https://github.com/istio/release-builder/pull/85

> **Note:** I have a follow-up `sort` PR (https://github.com/istio/test-infra/pull/2160) that will ensure the job order is deterministic regardless of execution order.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -479,6 +479,14 @@ presets:
       secretKeyRef:
         name: authentikos-token
         key: token
+- labels:
+    preset-override-deps: "release-1.4"
+  env:
+  - name: DEPENDENCIES
+    valueFrom:
+      configMapKeyRef:
+        name: release-1.4-deps
+        key: dependencies
 # Enable GOPROXY by default
 - env:
   - name: GOPROXY
--- a/prow/genjobs/gen-private-jobs.sh
+++ b/prow/genjobs/gen-private-jobs.sh
@@ -91,15 +91,37 @@ go run ./genjobs \
   --env BAZEL_BUILD_RBE_JOBS=0,GCS_BUILD_BUCKET=istio-private-build,GCS_ARTIFACTS_BUCKET=istio-private-artifacts,DOCKER_REPOSITORY=istio-prow-build/envoy,ENVOY_REPOSITORY=https://github.com/istio-private/envoy,ENVOY_PREFIX=envoy \
   --repo-whitelist proxy
 
+# istio/release-builder master test jobs(s) - pre/postsubmit(s)
+go run ./genjobs \
+  "${COMMON_OPTS[@]}" \
+  --branches=master \
+  --job-type presubmit,postsubmit \
+  --repo-whitelist release-builder \
+  --job-whitelist lint_release-builder,lint_release-builder_postsubmit,test_release-builder,test_release-builder_postsubmit,gencheck_release-builder,gencheck_release-builder_postsubmit
+
+# istio/release-builder release-1.4 test jobs(s) - pre/postsubmit(s)
+go run ./genjobs \
+  "${COMMON_OPTS[@]}" \
+  --branches=release-1.4 \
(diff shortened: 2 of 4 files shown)


prow/config.yaml
```diff
@@ -479,6 +479,14 @@ presets:
...
         name: authentikos-token
         key: token
+- labels:
+    preset-override-deps: "release-1.4"
+  env:
+  - name: DEPENDENCIES
+    valueFrom:
+      configMapKeyRef:
+        name: release-1.4-deps
+        key: dependencies
```
howardjohn, line 489: Do we have this configmap defined somewhere?

clarketm: Yes, it is  in the private cluster `test-pods` namespace.

prow/genjobs/gen-private-jobs.sh
```diff
@@ -91,15 +91,29 @@ go run ./genjobs \
   --env BAZEL_BUILD_RBE_JOBS=0,GCS_BUILD_BUCKET=istio-private-build,GCS_ARTIFACTS_BUCKET=istio-private-artifacts,DOCKER_REPOSITORY=istio-prow-build/envoy,ENVOY_REPOSITORY=https://github.com/istio-private/envoy,ENVOY_PREFIX=envoy \
   --repo-whitelist proxy
 
+# istio/release-builder test jobs(s) - pre/postsubmit(s)
+go run ./genjobs \
+  "${COMMON_OPTS[@]}" \
+  --branches=release-1.4,master \
+  --job-type presubmit,postsubmit \
+  --repo-whitelist release-builder \
+  --job-blacklist build-release_release-builder_release-1.4_postsubmit,build-release_release-builder_postsubmit,dry-run_release-builder_release-1.4,dry-run_release-builder_release-1.4_postsubmit,dry-run_release-builder,dry-run_release-builder_postsubmit,publish-release_release-builder_release-1.4_postsubmit,publish-release_release-builder_postsubmit
```
howardjohn: Should we use a whitelist here? Its not clear to me what is actual remaining?

prow/genjobs/gen-private-jobs.sh
```diff
@@ -91,15 +91,29 @@ go run ./genjobs \
...
+  --job-whitelist build-release_release-builder_release-1.4_postsubmit,build-release_release-builder_postsubmit
 
-# istio/release-builder test jobs(s) - presubmit(s)
+# istio/release-builder publish jobs(s) - postsubmit(s)
 go run ./genjobs \
   "${COMMON_OPTS[@]}" \
-  --job-type presubmit \
-  --repo-whitelist release-builder
+  --branches=release-1.4,master \
+  --env SOURCE_GCS_BUCKET=istio-private-prerelease/prerelease \
```
howardjohn: I think we don't need this job confusingly. The build step does a build + publish to staging repo. The publish just pulls a release from staging, and re-publishes it to the final output (including github this time).

In private we only need to build and push to staging, then we will do the github publish in public -- otherwise we have tons of complexity around getting all of the public credentials in private

clarketm: Okay, that makes. I update and removed this job.

prow/genjobs/gen-private-jobs.sh
clarketm: Sure, I can switch this to a whitelist, so that it is more clear.

clarketm: Switched to an explicit whitelist.

howardjohn (approved):

prow/genjobs/gen-private-jobs.sh
clarketm: I also split up `master` and `release-1.4` strictly for readability.

cjwagner (approved):

PR #2160, opened 2019-11-27
Genjobs: support lexicographical job sorting

clarketm:
Add `--sort` flag for lexicographical job sorting (ascending and descending). This is primarily for guaranteeing deterministic job order regardless of execution order, rather than aesthetics.


fejta (commented):
Are you familiar with https://github.com/fvbommel/util?

https://github.com/GoogleCloudPlatform/testgrid/blob/73df06c5d674235686b22b34f8f456e00cce7ae0/cmd/updater/main.go#L706

Testgrid uses it

clarketm:
> Are you familiar with https://github.com/fvbommel/util?
> 
> https://github.com/GoogleCloudPlatform/testgrid/blob/73df06c5d674235686b22b34f8f456e00cce7ae0/cmd/updater/main.go#L706
> 
> Testgrid uses it

I am now 😄. I will check it out, thanks.

fejta (approved):
/hold

PR #2163, opened 2019-11-28
Genjobs: update docs, create Dockerfile, autodeploy

clarketm:
1. update installation and usage documentation.
1. create Dockerfile and start publishing images.
1. ~add `go.mod` and make submodule for smaller images and modular build.~
1. add autodeploy jobs to publish docker images.


prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
```diff
@@ -85,14 +86,56 @@ postsubmits:
...
+      - image: gcr.io/istio-testing/build-tools:2019-10-18T22-40-12
+        command:
+        - entrypoint
+        args:
+        - make
+        - -C
+        - authentikos
+        - deploy
+        securityContext:
+          privileged: true
```
fejta: Thoughts on using GCB? I don't like seeing privileged containers in the trusted cluster.

prow/config.yaml
```diff
@@ -488,6 +488,14 @@ presets:
       configMapKeyRef:
         name: release-1.4-deps
         key: dependencies
+- labels:
+    preset-enable-dind: "true"
```
fejta: Please remove this preset, create the volume and mounts in the containers that need them.

prow/genjobs/README.md
```diff
@@ -6,16 +6,39 @@
...
+
 ```console
-$ go get -u istio.io/test-infra/prow/genjobs
+$ docker pull gcr.io/istio-testing/genjobs:latest
+```
+
+Install from source:
+
+```console
+$ git clone --depth=1 https://github.com/istio/test-infra.git
```
fejta: This will be easier for users to consume if you get rid of the `$`

prow/genjobs/go.mod
```diff
@@ -0,0 +1,14 @@
+module istio.io/test-infra/prow/genjobs
```
fejta: I'm concerned that we have so many submodules in this repo. Why do we need/want them?

https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
clarketm: I am not that familiar with GCB. I am open to it but how can we leverage it to avoid using `privileged` containers; or is GCB for satisfying the general image publishing need?

prow/genjobs/go.mod
clarketm: Originally this was intended to resolve the `go get ...` issue and any k8s dependency conflicts with the root `go.mod`. 

Since this uses more-or-less the same deps as the root (i.e. no reduced image size benefit) and does not solve the `go get ...` dilemma (i.e. b/c it depends on `k8s/test-infra` directly which uses `replace` directives (e.g. https://github.com/kubernetes/test-infra/issues/14070)) maybe submodule for this particular tool does **not** make sense after all.

clarketm: Conversely, for tools like `authentikos`, I think there is a clear benefit to using submodules for modularity, image size reduction, separate `go` versioning, and usability (UX).

prow/config.yaml
clarketm: Isn't the intended purpose of a `preset` to reduce duplication for commonly reused configuration like this?

clarketm: Isn't the intended purpose of a `preset` to reduce duplication for commonly reused configuration?

prow/genjobs/README.md
clarketm: As a user, it is clear that this is a terminal command when I see the `$` prefix? But I see what you mean; if someone were to come along and try to copy-and-paste. Updated!

prow/config.yaml
fejta: In general DAMP is more useful than DRY for this sort of thing https://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests -- readability is more important than avoiding [redundancy].

When the repetition is so onerous, istio has genjobs to address this problem. Therefore I do not see a use for presets in istio.

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
fejta: GCB jobs run in their cluster, not ours. So our cluster remains secure and figuring out how to deal with this problem becomes GCB's problem. Ask Katharine how we're doing this for k8s

prow/config.yaml
clarketm: Sounds good; thanks for the clarification. I will remove this preset and move these jobs to the job generator.

clarketm: Added some updates to the job generator and moved these jobs over to it.

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
clarketm: I moved these jobs over to the generator for now. I will investigate GCB as an alternative once I become more familiar with it.

prow/genjobs/README.md
clarketm: Updated

prow/config.yaml
clarketm: It will port the other, related jobs to the generator in a folllow-up. That change will mutate the job name (which requires additional care) and I want to keep this particular PR atomic and manageable in size.

clarketm:
@fejta 
PTAL. I have resolved all the suggestions and will look more into GCB as an alternative for these types of jobs #2211.

prow/config/config_test.go
```diff
@@ -334,7 +335,7 @@ func TestConfig(t *testing.T) {
 
 func TestTrustedJobs(t *testing.T) {
 	const trusted = "test-infra-trusted"
-	trustedPath := path.Join(*jobConfigPath, "istio", "test-infra", "istio.test-infra.trusted.master.yaml")
+	trustedPath := path.Join(*jobConfigPath, "istio", "test-infra")
```
clarketm, line 338: FYI, I am permitting anything in `istio/test-infra` directory to be "trusted".

fejta (approved):
/hold

PR #2167, opened 2019-12-02
increase perf timeout for running more tests

carolynhu:
This PR can only be merged after this: https://github.com/istio/tools/pull/591

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -159,9 +159,9 @@ periodics:
   name: daily-performance-benchmark
   branches: master
   decorate: true
-  timeout: 12h
+  timeout: 14h
   decoration_config:
-    timeout: 12h0m0s
+    timeout: 14h0m0s
   extra_refs:
     - org: istio
       repo: tools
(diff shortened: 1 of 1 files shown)


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -209,7 +209,32 @@ periodics:
...
         command:
           - entrypoint
-          - perf/benchmark/run_benchmark_job.sh
+          - perf/benchmark/run_istio_benchmark_job.sh
+
+- cron: "0 8 * * *" # starts every day at 08:00AM UTC
+  name: daily-linkerd-performance-benchmark
+  branches: master
+  decorate: true
+  timeout: 12h
```
howardjohn: does this one take 12hr as well?

carolynhu: it should be like around 3 hr actually. I just put it here in case we have more tests configurations.

howardjohn: How does it become 3hrs? it looks like the test is just running at 6 different# of connections for 4min each shouldn't that be 24min?

carolynhu: sorry, it should be  4min * 6 * 3= 72 min

mandarjog (approved):
We should just have captive clusters now.
12 hours is already a lot, and we want to eliminate variation between tests.

mandarjog:
john is it ok to increase timeouts to these levels? is there a scheduling issue here with successive tests?

howardjohn (approved):
Should be fine. We have two clusters for this, so as long as the tests don't go over 24hrs (realistically 20hrs or so to give us a buffer) we should be fine? SEe https://velodrome.istio.io/dashboard/db/boskos-dashboard?orgId=1&from=now-30d&to=now the bottom panel

Issue #2161, opened 2019-11-27
Genjobs: additionally support a yaml configuration file 


clarketm:
As the number of jobs generated by [`genjobs`](https://github.com/istio/test-infra/tree/master/prow/genjobs) grow, so does the complexity in the command-line invocations and their respective arguments (e.g. [`gen-private-jobs.sh`](https://github.com/istio/test-infra/blob/master/prow/genjobs/gen-private-jobs.sh)). 

While having a CLI interface is *great* for simple use-cases, additionally providing a configuration file as input (similar to [`generate.go`](https://github.com/istio/test-infra/blob/master/prow/config/generate.go)) can greatly reduce complexity and maximize utility of the tool.

Issue #1627, opened 2019-08-08
Create a job to update common files across repos


geeknoid:
Istio has the istio/common-files repo which contains files which should appear in every development repo in the org. When a push occurs in istio/common-files, we want a job to iterate through all repos in the org. For each repo, check out the repo and run "make updatecommon" and create a PR to push all changes.

howardjohn:
Related to https://github.com/istio/istio/issues/16381

clarketm:
@geeknoid - based on our discussion, I will take this off your plate (unless it is already in progress).

geeknoid:
Excellent, thanks.

@jasonwzm FYI

clarketm:
/assign @Biwwie

As discussed, the rough process should be something like:

Create `postsubmit` job(s) with 2 possible *triggers*:
1. https://github.com/istio/tools/tree/master/docker/build-tools dir changes
1. https://github.com/istio/common-files repo changes

With the following actions:
1. If a change to `build-tools`, then update the image in all relevant jobs (e.g. https://github.com/istio/test-infra/blob/5fa628ecc5a0794f6a3cd727e8fa7118f3019edd/prow/cluster/jobs/istio/api/istio.api.master.gen.yaml#L18)
1. If a change to common files, propagate **new** files to repo *consumers* (e.g. `istio/istio`, `istio/test-infra`)
    1. If and only if the apply does **not** break current presubmits
1. Create a PR for the above changes.

---
**Aside:** 
If we can simplify this flow with *actual* git submodule(s), then now would be an *ideal* time to make that change.

howardjohn:
More info about a broader issue: https://github.com/istio/istio/issues/16381

clarketm:
> More info about a broader issue: [istio/istio#16381](https://github.com/istio/istio/issues/16381)

Great; I was not aware of this being a broader issue. It sounds like, making the approach canonical and extensible, so that it can be leveraged in other capacities, is a key requirements.

jasonwzm:
> 2. If a change to common files, propagate **new** files to repo _consumers_ (e.g. `istio/istio`, `istio/test-infra`)

One thing to note that, the branch does matter here. `istio/test-infra` only has one branch, but the jobs it contains will depend on different images.

clarketm:
> One thing to note that, the branch does matter here. `istio/test-infra` only has one branch, but the jobs it contains will depend on different images.

[...]

PR #2172, opened 2019-12-04
Automate `istio/api` cross-repo file generation

### LINKED ISSUE #2103: Automate istio/api cross-repo file generation and propagation
If a change is done in a branch of `istio/api`, we should trigger an update in the *same* branch of `istio/client-go`, followed by an update in the *same* branch of `istio/istio`.

**e.g.**
1. A change occurs (e.g. PR merged in `istio/api`)
2. In response to this *initial* change in `istio/api`, in `istio/client-go` we will run:
```sh
# update `istio/api` dependency
go get "istio.io/api@<branch>"

# regenerate files
make gen

# Push change to a PR
git push ...
```

3. In response to the changes in `istio/api` `istio/client-go`, in `istio/istio` we will run:
```sh
# update `istio/api` dependency
go get "istio.io/api@<branch>"

# update `istio/client-go` dependency
go get "istio.io/client-go@<branch>"

# regenerate files
make gen

# Push change to a PR
git push ...
```

---

The trigger for 1 -> 2 should be a standard `postsubmit` job run after *any* change to `istio/api` that creates a PR to `istio/client-go` with changes. The trigger from 2 -> 3 also will be `postsubmit` job run after *any* change that will create a PR to `istio/istio` with changes. 

If we want to tweak the "after *any*" part we can get more sophisticated with trigger.

clarketm:
This runs `make gen` and generates a PR to `istio/client-go` and `istio/istio` on *postsubmit* of `istio/api` for branches **master** ~and **release-1.4**~.

fix #2103


tools/automator/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM gcr.io/k8s-prow/pr-creator:latest-root as deps
+FROM gcr.io/istio-testing/build-tools:master-latest
```
clarketm: Whilst this works, I am not fond of using `build-tools:master-latest` for all branches. 

Maybe we can use cloud build to generate images for all branches or use `build-tools` and pull in the dependencies in the script ~or set `BUILD_WITH_CONTAINER` to `"1"` and run in container.~?

clarketm: @howardjohn - when you have a moment can I get your thoughts?

howardjohn: Would it make sense to just add these to the build-tools containers? Then we completely avoid this issue.

Also I think we really only need this on master. We shouldn't be changing the API after a release ideally. At least not for 1.4, in the future maybe it would be useful for 1.5

clarketm: > Would it make sense to just add these to the build-tools containers? Then we completely avoid this issue.

I am cool with that. The `pr-creator` is an all-round useful tool so it would be beneficial to add to the `build-tools` image. For this script itself, I can clone it in through `extra_refs`.

> Also I think we really only need this on master. We shouldn't be changing the API after a release ideally. At least not for 1.4, in the future maybe it would be useful for 1.5

Okay, I will dial it back to only `master` until/if it makes sense to extend to future `release` branches. Also, if we are going to use the `build-tools` image directly, then we can leverage your job generator and configs as well which is a plus for maintainability.

howardjohn: For what it's worth the job generated supports a custom image per job as well so I think you could use it anyways, although we may need to add support for adding the GitHub mounts

clarketm: Nice, that is good to know. On second thought, I did see that the `image` field could be specified in the config. 

As an aside, at some point we should consider coalescing the logic of the 2 job generators, possibly exposed as different subcommands. Yours is an *actual* generator (i.e. `generate`), while mine is more of a translator (i.e. `translate`). At the point when I can get rolling and make progress on https://github.com/istio/test-infra/issues/2161 (since we probably want a unified schema) might be a good time to evaluate.

tools/automator/automator.sh
```diff
@@ -0,0 +1,185 @@
...
+    git show --shortstat
+    git push --force "https://$user:$token@github.com/$user/$repo.git" "HEAD:$branch-automator"
+
+    create_pr
+  fi
+
+  popd || print_error_and_exit "invalid dir: $tmp_dir"
+}
+
+main "$@"
```
fejta, line 185: Can we make this a go program? There are quite a lot of flags for a go program.

An example: https://github.com/kubernetes/test-infra/pull/15508

prow/cluster/jobs/istio/api/istio.api.all.yaml
```diff
@@ -0,0 +1,57 @@
...
+    decorate: true
+    path_alias: istio.io/api
+    spec:
+      containers:
+      - image: gcr.io/istio-testing/automator:0.0.1
+        command:
+        - /usr/bin/automator
+        args:
+        - --org=istio
+        - --repo=client-go
```
fejta: Why is automator cloning the repo vs having prow do it?

tools/automator/automator.sh
clarketm: Interestingly, I actually have a stashed branch locally with an *incomplete* Go version of this. Since this is still simple maybe `bash` makes since for this POC? I plan to iterate on the Go version either way since I will need something more robust for https://github.com/istio/test-infra/issues/1627.

fejta: IMO when you start defining flags is generally a good signal that its time to upgrade to python or golang.

prow/cluster/jobs/istio/api/istio.api.all.yaml
clarketm: This is to decouple it from Prow. This makes it easier to run and test outside of Prow.

tools/automator/automator.sh
clarketm: For arg parsing in `bash`, using `getopts` is simple and intuitive IMO especially for strings and filepaths. The decision point for me (in choosing a higher-level programming language) is overall code complexity and maintainability.

prow/cluster/jobs/istio/api/istio.api.all.yaml
fejta: If the person runs it outside of prow, can't we/shouldn't we assume they already have the repo checked out and at the commit they want?

clarketm: The purpose of the `org`, `repo`, and `branch` args to the script is so that the user does not have to make any assumptions about local filesystem state (or worry about what directory they are in, what commit is checked out, whether the working tree is dirty, etc.). This makes it stateless and allows `git` remote to be the source of truth.

tools/automator/automator.sh
clarketm: I will continue to iterate on the Go version of this and update.

clarketm: I will tackle this as a follow-up (https://github.com/istio/test-infra/issues/2212) and consider this `bash` code as a POC.

prow/config/jobs/api.yaml
```diff
@@ -1,11 +1,33 @@
...
   - name: build
     command: [make, presubmit]
 
   - name: gencheck
     command: [make, gen-check]
+
+  - name: gen_client-go
+    type: postsubmit
+    command:
+    - /home/prow/go/src/istio/test-infra/tools/automator/automator.sh
```
howardjohn: I have no opinion here: is it better to have the full path, or `../test-infra`?

howardjohn (approved):

prow/config/jobs/api.yaml
clarketm: I went with *absolute* to be explicit but *relative* path should be equally as good. I will update to relative since it is cleaner and safe from any prow root dir changes (look like there is a typo either way since `istio/test-infra` should be `istio.io/test-infra` - I didn't realize the path alias was automatically added).

clarketm: Updated. As far as best practices ... I do not think there is one. Whatever is the most intuitive is best and on second thought that is probably relative.

howardjohn: Like I said before I don't mind either way, but I could see an argument that relative is preferred because /home/prow/go could hypothetically change. In practice, it won't

clarketm:
- [x] I will manually trigger one of these ProwJobs to verify it still functions after the recent image and job changes. (https://prow.istio.io/log?id=2&job=gen_client-go_api_postsubmit)

prow/config/jobs/api.yaml
clarketm: That is a good point. Yes, while we do try to remain backwards-compatible there is no guarantee the root go path will not change. So relative is the safer bet.

clarketm:
@fejta - PTAL. I created a separate issue to point this over to Golang, which should be simple to do as a follow-up once it is proven this automation is effective.

fejta (approved):
/hold

PR #2176, opened 2019-12-04
Authentikos: use SA until workload identity can be properly configured

clarketm:
use service account (SA) until workload identity can be properly configured

### DIFF
--- a/prow/cluster/private/authentikos_deployment.yaml
+++ b/prow/cluster/private/authentikos_deployment.yaml
@@ -41,6 +41,16 @@ spec:
         args:
         - --verbose
         - --secret=authentikos-token
+        - --creds=/etc/creds/service-account.json
         - --namespace=test-pods
         - --scopes=https://www.googleapis.com/auth/devstorage.full_control
         - "--format=Authorization: Bearer %v"
+        volumeMounts:
+        - name: creds
+          mountPath: /etc/creds
+          readOnly: true
+      volumes:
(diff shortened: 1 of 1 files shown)


clarketm:
One of the things we need to fix `build-release_release-builder_release-1.4_postsubmit_priv`. I can always switch back to workload identity once it has been properly configured in all clusters.

prow/cluster/private/authentikos_deployment.yaml
```diff
@@ -41,6 +41,16 @@ spec:
...
         - "--format=Authorization: Bearer %v"
+        volumeMounts:
+        - name: creds
+          mountPath: /etc/creds
+          readOnly: true
+      volumes:
+      - name: creds
+        secret:
+          defaultMode: 0644
+          secretName: service-account
```
cjwagner, line 56: We should probably be using a separate service account and/or token instead of using the `service-account` secret for everything. If this is supposed to be temporary I guess it doesn't matter much.
/shrug

cjwagner (approved):
/hold

prow/cluster/private/authentikos_deployment.yaml
clarketm: Yes, it is temporary otherwise I would change it. Also this runs in the private cluster where there is one SA tuned specifically for storage, so not crucial for separation concerns, especially since this impermanent.

PR #2186, opened 2019-12-05
Remove build-pool

howardjohn:
Currently we have two node pools: build-pool (20 nodes) used for
istio/proxy, and test-pool (60 nodes) used for everything else. 20 nodes
is probably way to many for proxy, but rather than trying to optimize
this we should just merge the two so we can bin-pack more efficiently


nmittler (approved):

clarketm (approved):
I will enable scaling as well, like I did with the *private* cluster. It seems to be working great to reduce cost. I agree thought that a bunch of separate node pools is bewildering.

PR #2190, opened 2019-12-06
Add "Override Annotations" feature to genjobs

chases2:
Allows a passed argument to inject a set of annotations to every generated job.
Default behavior is to delete all annotations, as before.

Also implements this feature to prevent all generated private jobs from making unused TestGrid groups.

xref #2140

/assign @clarketm @howardjohn


clarketm (approved):

clarketm:
@chases2 
Need to run `make gen` to regenerate the jobs, otherwise lgtm. Thanks!

chases2:
Update: `make gen`-generated code added in separate commit.

chases2:
Hey, that list of failures is way shorter! :tada:

One more commit to address those (unrelated) failures

fejta (approved):
/hold

PR #2197, opened 2019-12-09
add istio/bots policybot deployment postsubmit

clarketm:
Dependent on:
- [x] updates to `istio/bots` `policybot` deployment: https://github.com/istio/bots/pull/112; and 
- [x] Prow version: [`v20191211-ff6c35575`](https://github.com/kubernetes/test-infra/pull/15571) or newer.

### DIFF
--- a/prow/cluster/jobs/istio/bots/istio.bots.master.gen.yaml
+++ b/prow/cluster/jobs/istio/bots/istio.bots.master.gen.yaml
@@ -109,6 +109,81 @@ postsubmits:
           privileged: true
       nodeSelector:
         testing: test-pool
+  - annotations:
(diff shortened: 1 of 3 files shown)


clarketm:
/hold
- [x] Due to the risk, this warrants creating a manual ProwJob for verification which I will do shortly. This will ensure cluster access is sufficient given the service account, etc. (https://prow.istio.io/view/gcs/istio-prow/logs/deploy-policybot_bots_postsubmit/7)

Since this is a `postsubmit`, we can leverage gitops for both deployments and rollbacks. For example, if a `policybot` rollback is desired, then we can simply do a revert in GitHub and when the PR merges a rollback deployment will occur.

howardjohn (approved):

PR #2203, opened 2019-12-10
Fix 1.3 proxy branch spec

howardjohn:


howardjohn:
this is matching `release-1.3.6-patch` which we don't want

clarketm (approved):

brian-avery (approved):

PR #2215, opened 2019-12-12
fix broken recreate_prow_configmaps.py

clarketm:
This is used as a local tool for emergency configmap updates. The gzip is needed, otherwise there is a risk of `job-config` configmap failure when size over 1mb.

### DIFF
--- a/prow/recreate_prow_configmaps.py
+++ b/prow/recreate_prow_configmaps.py
@@ -26,9 +26,10 @@
 
 from __future__ import print_function
 
-from argparse import ArgumentParser
 import os
 import sys
+
+import argparse
 import subprocess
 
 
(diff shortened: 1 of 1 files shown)


prow/recreate_prow_configmaps.py
```diff
@@ -26,19 +26,20 @@
 
 from __future__ import print_function
 
-from argparse import ArgumentParser
 import os
 import sys
+
 import subprocess
+from argparse import ArgumentParser
```
fejta: Please import modules (argparse) not methods/classes (ArgumentParser)

http://google.github.io/styleguide/pyguide.html#22-imports

fejta (approved):
/hold

PR #2216, opened 2019-12-13
Align perf cluster with standard per machine type.

mandarjog:
Align perf cluster with n1-standard-32 machine type which is standard in performance tests.

### DIFF
--- a/boskos/resources.yaml
+++ b/boskos/resources.yaml
@@ -42,8 +42,8 @@ resources:
     content: |
       gcp-perf-test:
         - clusters:
-          - machinetype: n1-highmem-8
-            numnodes: 7
+          - machinetype: n1-standard-32
+            numnodes: 4
             version: 1.13
             zone: us-central1-f
             scopes:


boskos/resources.yaml
```diff
@@ -42,8 +42,8 @@ resources:
     content: |
       gcp-perf-test:
         - clusters:
-          - machinetype: n1-highmem-8
-            numnodes: 7
+          - machinetype: n1-standard-32
+            numnodes: 4
```
carolynhu, line 46: You want to allocate more CPU and Memory for perf cluster?

howardjohn: agree, do we need this much for just running two pods? Seems a bit overkill unless we are going to do the servicegraph testing.

If its for servce graph testing, do we want 2 of these clusters?

howardjohn: To clarify, gcp-perf-test currently has 2 clusters, and it doesn't autoscale

mandarjog: @carolynhu the machine types are/may be different and I have never run anything on n1-highmem. 
We can reduce it to 2 nodes I already reduced it to 4. 
@howardjohn what change are you looking for?

howardjohn: Just cost saving. Do we need 128 cores when we are only using a small amount? We are just running two pods right?

mandarjog:
I am going to create dedicated clusters for perf.

howardjohn:
these sort of are dedicated clusters with built in logic to ensure only one
test is using them at once, and handles cleaning up etc. What is the
benefit?

PR #2221, opened 2019-12-16
Authentikos: add grandmatriarch example

clarketm:


authentikos/examples/authentikos-grandmatriarch-deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+          source.developers.google.com FALSE  {{.Now | .TimeToUnix | .Add 3600}} {{.Token}}
+        volumeMounts:
+        - name: creds
+          mountPath: /etc/creds
+          readOnly: true
+      volumes:
+      - name: creds
+        secret:
+          defaultMode: 0644
+          secretName: service-account
```
cjwagner, line 86: We should use a separate secret specifically for this.

cjwagner (approved):

authentikos/examples/authentikos-grandmatriarch-deployment.yaml
clarketm: I can update this. I don’t have any preference on naming. Do you prefer a particular name (or naming convention) or is something like *authentikos-service-account* reasonable?

cjwagner: That sounds great, thanks!

chases2 (approved):

cjwagner:
/hold cancel
We can iterate on this if needed, but I'd like to start trying it out.

howardjohn (approved):

PR #2223, opened 2019-12-17
Update common-files on postsubmit

clarketm:
This covers **part 2** (`postsubmit` on https://github.com/istio/common-files repo changes) of: https://github.com/istio/test-infra/issues/1627#issuecomment-552665234


howardjohn (approved):

clarketm:
@fejta can you please take a look.

fejta (approved):
Why is this one coming our way?
/hold

clarketm:
> Why is this one coming our way?

In this case, it is because we own the `tools/` directory where `automator.sh` script lives. The `tools/` dir does not have an explicit codeowner so falls back to the wildcard (*).

fejta:
https://github.com/istio/test-infra/blob/master/CODEOWNERS#L1 -> points to the oncall team, anything specific to istio should not go to prow oncall.

Do we consider this common tooling? If so can we move it into kubernetes/test-infra?

clarketm:
> https://github.com/istio/test-infra/blob/master/CODEOWNERS#L1 -> points to the oncall team, anything specific to istio should not go to prow oncall.
> 
> Do we consider this common tooling? If so can we move it into kubernetes/test-infra?

These are general tools but specific to Istio needs ATM. I will follow up and change the CODEOWNERS. To help me understand desired ownership, only Prow Kubernetes resources and Prow configuration should be owned by our team; all else owned by Istio?

PR #2225, opened 2019-12-17
update and use gcs.istio.io hostname

### LINKED ISSUE #2222: regression: test logs no longer accessible without GCP account
Clicking artifacts used to lead to a gcsweb page that was public, now it leads directly to GCS which requires a GCP account (not sure if its public after that, as I have access - but others may not)

Copying more details from slack:

Erick Fejta Dec 10th at 11:41 AM
:bump: https://github.com/istio/test-infra/pull/2202
istio-testingistio-testing

John Howard  5 days ago
@fejta did this (or another recent change) change the "Artifacts" button? It does not seem to work as before. See https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio.io/6059/lint_istio.io/3107 for reference.
Before it would link to istio gcsweb. Now its linking to k8s gcsweb, which redirects me to pantheon. I suspect this means for people without a google account this won't work

Erick Fejta  5 days ago
Possibly, let me look in a bit

Erick Fejta  5 days ago
Any chance this is a redirection bug or config bug? That's my first guess. 
Also I'm not sure I've ever seen a istio gcsweb... How recently have you seen it?

John Howard  5 days ago
It definitely used to go to a gcsweb not gcloud. I am not 100% sure it was ever istio's gcsweb. We do have one though, but maybe it wasn't used for this. let me check, I link to it in issues often

John Howard  5 days ago
I can only find myself linking to the spyglass page (eg https://prow.istio.io/view/gcs/istio-prow/logs/integ-istioctl-k8s-tests_istio_postsubmit/133) or directly to a file (eg https://storage.googleapis.com/istio-prow/logs/integ-istioctl-k8s-tests_istio_postsubmit/133/artifacts/kind/istio-testing-control-plane/pods/istio-system_istio-ingressgateway-776f79dc9c-xxh6l_2f57099a-1852-4f26-b4e7-1e4939754580/istio-proxy/0.log). Never to the "artifacts" page

John Howard  5 days ago
But I am confident it was a gcsweb, just not sure if it was k8s.io or istio.io

Erick Fejta  5 days ago
https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L31
config/prow/config.yaml:31

Erick Fejta  5 days ago
https://github.com/istio/test-infra/blob/master/prow/config.yaml#L30
prow/config.yaml:30

John Howard  5 days ago
Hm ok so I guess the behavior of gcsweb.k8s.io may have changed possibly?

clarketm:
fix #2222

cc @howardjohn

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -27,7 +27,7 @@ sinker:
 deck:
   spyglass:
     size_limit: 500000000 # 500MB
-    gcs_browser_prefix: https://gcsweb.k8s.io/gcs/
+    gcs_browser_prefix: https://gcsweb.istio.io/gcs/
     testgrid_config: gs://k8s-testgrid/config
     testgrid_root: https://testgrid.k8s.io/
     viewers:
--- a/sites/istio.io/gcsweb/deployment.yaml
+++ b/sites/istio.io/gcsweb/deployment.yaml
@@ -23,6 +23,7 @@ spec:
         - name: gcsweb
           image: k8s.gcr.io/gcsweb-amd64:v1.0.6
           args:
+            - -b=istio-prow
             - -b=istio-build
             - -b=istio-artifacts
             - -b=istio-release


howardjohn (approved):

cjwagner (approved):

howardjohn:
@chases2 can you take a look? This is impacting us, as some can't view test artifacts

chases2 (approved):

PR #2226, opened 2019-12-17
migrate gcsweb

### LINKED ISSUE #2227: gcsweb TLS cert not auto-renewing
Same core issue as #2207 but since I am going to resolve these issues separately I am breaking this out.

clarketm:
Migrate `gcsweb` to google managed certs. 

fix #2227

/hold 
> Will be making an announcement in #istio Slack and performing the switch off-peak hours as there is a small amount of downtime required for the ingress and load balancer changes. 

---
The better (and safer) migration path is to make the switch via the DNS A record. The IP is `35.201.76.230`. 

**Before** merge:
- [x] Update A record for `gcsweb`
- [x] Re-provisioned TLS cert
- [x] Use a separate NS (e.g. `gcs`) for resources so not to conflict with exiting resources.
- [x] Configure default BE health check to be `/healthz` ([xref](https://github.com/kubernetes/ingress-gce/blob/master/docs/faq/gce.md#can-i-configure-gce-health-checks-through-the-ingress)) or set `readinessProbe` ([xref](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#health_checks))

**After** merge:
- [x] Cleanup old static IP, NS and associated resources. 
> all resources save to purge have been.


howardjohn (approved):

cjwagner (approved):
/hold

fejta (approved):

PR #2228, opened 2019-12-17
common-files update master-2019-12-15T16-17-48

sdake:
Update to common-files with image master-2019-12-15T16-17-48


howardjohn (changes requested):
somethign went wrong here

sdake:
@clarketm fyi - this was a result of running the automation I handed off to you. Just thought you should see it.

I am not sure what the root cause is - possibly a change in my local branch of test-infra on master.

Cheers
-steve

clarketm:
Thanks for the info. FWIW, in the job that I created (adapted from your script) that automates this when there are *not any conflicts*, it will perform a fresh shallow clone of the destination repo (ref: https://github.com/istio/test-infra/pull/2223), so I think this may be an issue with the local repo source on your filesystem.

Issue #1988, opened 2019-10-24
Prow cluster resource leak


howardjohn:
Background: https://github.com/kubernetes-sigs/kind/issues/759

When we started using Kind, we found all of the sudden our tests started flaking often. It turns out this was due to a resource leak by not cleaning up kind clusters at all. I am not 100% what was leaking, but something related to hostPath mount to `/lib/modules` or `/sys/fs/cgroup`. Once we added cleanup configuration, things went back to normal.

The best reference we had was the "resting CPU" rate of the nodes. In the original case, after a 2 week period was at between 30-90% with no tests running.

Now, months later, the same problem has returned :slightly_frowning_face: 

![2019-24-10_09-39-14](https://user-images.githubusercontent.com/623453/67506582-30e37c00-f642-11e9-85a0-1d59cb168908.png)

This time it it much more subtle though, only 20% max usage after 2 weeks.

My suspicion is that we still have some leaks from kind.

Action items:
* [x] Investigate stronger cleanup, as recommended by @bentheelder

> The docker in docker runner / wrapper script we use in test-infra / prow.k8s.io also terminates all containers in an exit handler, amongst other things, redundantly to the cluster deletion we do in the kind specific scripts.

* [x] Problem seemed to start on October 1st. Was there any changes there?

* [ ] Can we identify that we have a leak and clean up proactively as a periodic job?

BenTheElder:
You should be able to check the cgroups on a host node for if there's actually something leaking, there _shouldn't_ be. I've done fairly extensive testing of that with kind itself locally, but it's possible something else is up in CI.

xref: https://github.com/kubernetes-sigs/kind/issues/421 (note that the "memory cgroups leak" is actually just a kernel bug in flushing them after deletion, we'e also seen this just running containers without any nesting etc.)

howardjohn:
Likely root cause 01e1bf6b, adds kind to a new repo. This repo does have the `trap cleanup` stuff but for some reason its not running.. When I run locally it cleans up, but in CI I don't see the cleanup occuring: https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_installer/457/base_installer/96

BenTheElder:
is this trap in the image or in the repo? where are the image sources? 👀

howardjohn:
[...]

PR #2243, opened 2019-12-20
Make cgroup mount readonly

howardjohn:
For https://github.com/istio/test-infra/issues/1988


howardjohn:
After some local testing I am about 80% sure this will work :slightly_smiling_face: . If not we can always revert

nmittler (approved):

PR #2245, opened 2019-12-26
Initialize string pointer in struct

JimmyCYJ:
Some circleci tests fail due to initialization failures
https://circleci.com/gh/istio/istio/463030?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


howardjohn (changes requested):
This is not the right fix (or one that works). The issue is 1.2 branch is using an old golang version where `go get` is not behaving as we want, and is updating the github library to latest rather than what is in the go.mod file.

PR #2247, opened 2020-01-01
Use istioctl install for standard integ tests

howardjohn:
This is tested in https://github.com/istio/istio/pull/19867 showing that
all tests pass with this enabled (except security, due to known issues,
which is intentionlly left out here).

Helm is still covered by other targets in postsubmit


howardjohn:
For https://github.com/istio/istio/issues/17221

nmittler (approved):

PR #2249, opened 2020-01-02
Require TestGrid Presubmit Test

chases2:
It may be worth it to make this test required, so that the TestGrid config can't fall out of sync and stop working.


chases2:
xref #2140

howardjohn (approved):

cjwagner (approved):

PR #2254, opened 2020-01-06
migrate GitHub reporting from plank to crier

### LINKED ISSUE #1976: Migrate ProwJob GitHub reporting from Plank to Crier.
Migrate ProwJob GitHub reporting from Plank to Crier. 

https://github.com/kubernetes/test-infra/blob/f0f14f5bcd398eab90ec5df74c531237ef802f9c/prow/cmd/crier/README.md

clarketm:
close https://github.com/istio/test-infra/issues/1976

### DIFF
--- a/prow/cluster/crier_deployment.yaml
+++ b/prow/cluster/crier_deployment.yaml
@@ -20,11 +20,18 @@ spec:
       - name: crier
         image: gcr.io/k8s-prow/crier:v20191217-e7571bd25
         args:
+        - --github-workers=1
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
+        - --github-token-path=/etc/github/oauth
         - --slack-workers=1
         - --slack-token-file=/etc/slack/token
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
         volumeMounts:
+        - name: oauth
+          mountPath: /etc/github
+          readOnly: true
         - name: slack
           mountPath: /etc/slack
           readOnly: true
@@ -35,6 +42,9 @@ spec:
           mountPath: /etc/job-config
           readOnly: true
       volumes:
+      - name: oauth
+        secret:
+          secretName: oauth-token
       - name: slack
         secret:
           secretName: slack-token
--- a/prow/cluster/plank_deployment.yaml
+++ b/prow/cluster/plank_deployment.yaml
@@ -15,17 +15,14 @@ spec:
       - name: plank
         image: gcr.io/k8s-prow/plank:v20191217-e7571bd25
         args:
+        - --github-token-path=""
         - --config-path=/etc/config/config.yaml
         - --dry-run=false
         - --job-config-path=/etc/job-config
         - --tot-url=http://tot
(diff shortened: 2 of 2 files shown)


prow/cluster/plank_deployment.yaml
```diff
@@ -19,13 +19,9 @@ spec:
         - --dry-run=false
         - --job-config-path=/etc/job-config
         - --tot-url=http://tot
-        - --github-endpoint=http://ghproxy
-        - --github-endpoint=https://api.github.com
         - --kubeconfig=/etc/kubeconfig/istio-config
+        - --skip-report=true
```
cjwagner, line 24: Due to [this nonsense](https://github.com/kubernetes/test-infra/blob/ca8fda0be1599c631fcdba7d9a659f1dd0d5b8b4/prow/flagutil/github.go#L71-L75) you need to specify `--github-token-path=""`

prow/cluster/crier_deployment.yaml
```diff
@@ -20,11 +20,17 @@ spec:
       - name: crier
         image: gcr.io/k8s-prow/crier:v20191217-e7571bd25
         args:
+        - --github-workers=1
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
         - --slack-workers=1
         - --slack-token-file=/etc/slack/token
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
```
cjwagner, line 30: Please specify `--github-token-path` even if the automagic default happens to be correct.

prow/cluster/plank_deployment.yaml
clarketm: Are you certain this is needed? I noticed kubernetes/test-infra [plank deployment](https://github.com/kubernetes/test-infra/blob/ca8fda0be1599c631fcdba7d9a659f1dd0d5b8b4/prow/cluster/plank_deployment.yaml#L38-L45) does not explicitly set this flag either. Do we need to continue to mount the **oauth** secret as well because of this? ([example](https://github.com/kubernetes/test-infra/blob/ca8fda0be1599c631fcdba7d9a659f1dd0d5b8b4/prow/cluster/plank_deployment.yaml#L64-L66))

prow/cluster/crier_deployment.yaml
clarketm: Now explicitly setting `--github-token-path`.

prow/cluster/plank_deployment.yaml
clarketm: Updated, I understand now why this is needed. 

If we wanted to clean this up could we just not parse GitHub options in `plank` if `skip-report=true`?

cjwagner: Yeah that would work. Alternatively we could switch plank to not default the token path (after a warning period). It does seem like `skip-report` and `--github-token-path` should be mutually exclusive though.

cjwagner (approved):

fejta (approved):

clarketm:
I will take on the deployment and monitoring of the change.
/assign

PR #2256, opened 2020-01-07
Use multiple comma-delimited dirs in COMPONENT_FILE_DIR

clarketm:
Use multiple directories for `COMPONENT_FILE_DIR`.

> Dependent on deployment of https://github.com/kubernetes/test-infra/pull/15800

/hold


fejta (approved):
I wish we were using explicitly passed flags rather than implicitly inherited environment variables.

PR #2265, opened 2020-01-08
Fix peribolos command

howardjohn:
there is no `peribolos` binary in the image


brian-avery (approved):

clarketm:
> there is no `peribolos` binary in the image

FYI, using `/peribolos` as the entrypoint is also an alternative.

PR #2266, opened 2020-01-08
Use "default_decoration_configs" for istio-private

### LINKED ISSUE #2261: Use `default_decoration_configs` for istio-private org decoration configuration.

clarketm:
Use `default_decoration_configs` rather than job-level `decoration_config` overrides for **istio-private**.

> Also fix `--input` and `--output` paths which were incorrect for private job gen.

fix #2261


howardjohn (approved):
lgtm as long as we keep support for this in the genjobs tool

clarketm:
> lgtm as long as we keep support for this in the genjobs tool

I will not be removing the functionality (or any of the defaults for these fields) from `genjobs`

fejta (approved):

PR #2286, opened 2020-01-20
Update gke version for perf cluster

richardwxn:
1.5 perf test fails at installation step because of similar issue mentioned here: https://github.com/istio/istio/issues/20290


howardjohn (approved):
This is a good idea but a bad reason to do it, I don't think we should drop 1.13 support yet

PR #2292, opened 2020-01-23
Add conformance tests to presubmit

### LINKED ISSUE #19924: run conformance test on presubmit

howardjohn:
Fixes https://github.com/istio/istio/issues/19924

These are quick tests and break often, we should run in presubmit


howardjohn:
This was broken then subsequently fixed by https://github.com/istio/istio/commit/12fd6dde7dc985da46ef6a30787b1be128aca64c. We break this test all the time, and its really quick and stable so there is no reason to not make it default

nmittler (approved):

PR #2310, opened 2020-01-28
Cleanup release-1.5 automator jobs and fallback to direct GOPROXY

clarketm:
1.  Ensure `istio/test-infra@master` if added as an extraref for `release-1.5` jobs mounting it.
2. [Fallback](https://golang.org/doc/go1.13#modules) to a **direct** connection for `GOPROXY`


prow/config.yaml
```diff
@@ -515,9 +515,9 @@ presets:
       configMapKeyRef:
         name: release-1.4-deps
         key: dependencies
-# Enable GOPROXY by default
+# Enable GOPROXY by default and fallback to direct connection.
 - env:
   - name: GOPROXY
-    value: "https://proxy.golang.org"
+    value: "https://proxy.golang.org,direct"
```
howardjohn, line 521: This is already the default. I don't think its a fallback here. We do get `@master`, which exists, its just stale. The fallback would be if we did `@non-existent`

clarketm: It is my understanding that in Go@1.13 that `https://proxy.golang.org,direct` is the **default** when the `GOPROXY` variable is *unspecified* ([ref](https://golang.org/doc/go1.13#modules)). Here it was being explicitly set to `https://proxy.golang.org`. Regardless, I see what you mean regarding staleness and the `direct` not actually being reached. I will update to installing from SHA as well ... but I think we still want to fallback to `direct` here in case there is a **404** from proxy?

howardjohn: Ah yes, good point. I forgot about this. sgtm

howardjohn (approved):

prow/config.yaml
clarketm: I just pushed additional changes to do `go get istio.io/api@SHA` rather than branch.  The hope is that this, combined with falling back to *direct*, will fix the race 🤞. Thanks for suggesting this.

cjwagner (approved):

rlenglet:
@clarketm please revert this PR. It's breaking `release-1.3` tests with errors like:
```
invalid $GOPROXY setting: cannot have comma
```

PR #2319, opened 2020-01-30
Fix GOPROXY for proxy 1.4 repo

howardjohn:
They are on golang 1.12 which doesn't support fallback goproxy

ref https://github.com/istio/proxy/pull/2642

### DIFF
--- a/prow/cluster/jobs/istio/proxy/istio.proxy.release-1.4.yaml
+++ b/prow/cluster/jobs/istio/proxy/istio.proxy.release-1.4.yaml
@@ -26,6 +26,9 @@ presubmits:
       - image: gcr.io/istio-testing/prowbazel:0.5.16
         command:
         - ./prow/proxy-presubmit.sh
+        env:
+        - name: GOPROXY
+          value: "https://proxy.golang.org"
         resources:
           requests:
             memory: "8Gi"
@@ -57,6 +60,9 @@ presubmits:
       - image: gcr.io/istio-testing/prowbazel:0.5.16
         command:
         - ./prow/proxy-presubmit-asan.sh
+        env:
+        - name: GOPROXY
+          value: "https://proxy.golang.org"
         resources:
           requests:
             memory: "8Gi"
@@ -88,6 +94,9 @@ presubmits:
       - image: gcr.io/istio-testing/prowbazel:0.5.16
         command:
         - ./prow/proxy-presubmit-tsan.sh
+        env:
+        - name: GOPROXY
+          value: "https://proxy.golang.org"
         resources:
           requests:
             memory: "8Gi"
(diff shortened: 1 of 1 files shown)


clarketm (approved):

clarketm (approved):

howardjohn:
@clarketm is there no way to avoid "global" preset conflicts? eg

```
Could not load config: [job proxy-presubmit failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY, job proxy-presubmit-asan failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY, job proxy-presubmit-tsan failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY, job proxy-presubmit-release failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY]
```

clarketm:
> @clarketm is there no way to avoid "global" preset conflicts? eg
> 
> ```
> Could not load config: [job proxy-presubmit failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY, job proxy-presubmit-asan failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY, job proxy-presubmit-tsan failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY, job proxy-presubmit-release failed to merge presets for podspec: env var duplicated in pod spec: GOPROXY]
> ```

Here is the [code](https://github.com/kubernetes/test-infra/blob/a933a9650ced73be2b585d95cfa271de249f0d34/prow/config/jobs.go#L48-L83) in Prow and I am more-or-less certain that because of this `env` overrides are not allowed when defined both globally and at job-level. As an aside, I actually have a [reference implementation](https://github.com/istio/test-infra/blob/1b1c1c9d48b364b8edcf7d258b26afd4634827f0/prow/genjobs/cmd/genjobs/main.go#L244) of how to deal with overrides but because it is *potentially* breaking we may need to consider an alternative for now.

@cjwagner can you please confirm that this assumption is correct that a global preset cannot be overridden at the job-level?

howardjohn:
given goproxy is default now maybe we just remove this entirely

clarketm:
> given goproxy is default now maybe we just remove this entirely

I was thinking that as well. As long as we are okay with not using the proxy for jobs using go 1.12 or older since it does not have a *default* for `GOPROXY` AFAIK.

howardjohn:
At this point I think its fine, there are very very few jobs on go 1.12
(not sure which ones they are, unfortunately).

clarketm:
> At this point I think its fine, there are very very few jobs on go 1.12 (not sure which ones they are, unfortunately). 

👍 I agree. Not using a proxy is not a major impact for a small subset of jobs. https://github.com/istio/test-infra/pull/2321

PR #2340, opened 2020-02-03
Add retries to kubernetes postsubmit tests

howardjohn:
This is normally a really bad idea I think,
and just hides the problem we see. However, I think there is one use
case for it. In postsubmit, we rerun tests with different variants (ie
Kubernetes v1.13). These are the exact same tests running ~4x in
postsubmit. These tests exist to catch bugs in different environments -
it is highly unlikely that we have test flakes in different k8s versions
rather than just complete failures. Therefor, I think its appropriate to
add retries only for these redundant tests. Otherwise, we are just
making ourselves 4x more likely to have a test flake

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -1084,8 +1084,11 @@ postsubmits:
         - entrypoint
         - prow/integ-suite-kind.sh
         - --node-image
-        - kindest/node:v1.14.9
+        - kindest/node:v1.14.10
         - test.integration.kube.presubmit
+        env:
+        - name: INTEGRATION_TEST_FLAGS
+          value: ' --istio.test.retries=1 '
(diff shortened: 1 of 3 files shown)


prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
```diff
@@ -1196,8 +1202,11 @@ postsubmits:
         - entrypoint
         - prow/integ-suite-kind.sh
         - --node-image
-        - kindest/node:v1.16.3
+        - kindest/node:v1.16.4
         - test.integration.kube.presubmit
+        env:
+        - name: INTEGRATION_TEST_FLAGS
+          value: ' --istio.test.retries=1 '
```
ericvn, line 1209: Minor nit. It seems you use a single quote in some areas, and a double quote in others.

howardjohn: Its auto generated, I have no choice here. this is how the yaml library handles things to properly escape

ericvn (approved):

PR #2277, opened 2020-01-13
Remove installer master from tide

sdake:

PR #2344, opened 2020-02-05
Remove broken testrid and generated config

clarketm:
Once this change: https://github.com/kubernetes/test-infra/pull/16128#event-3008124032 goes through and the generated config ordering is deterministic, then we can add this `testgrid` target back if desired (or we can just wait and update the sha when the change lands). 

But since this target is broken and because the [testgrid/istio.gen.yaml](https://github.com/istio/test-infra/blob/master/testgrid/istio.gen.yaml) generated file is legacy and no longer used directly (explained here: https://github.com/istio/test-infra/pull/2277#issuecomment-582531899) it is not necessary and can be safely removed.

cc @sdake @howardjohn


sdake (approved):
LGTM.

howardjohn (approved):

PR #2350, opened 2020-02-06
Support custom "match-title" and revert "api" jobs to branch title

### LINKED ISSUE #20534: automator fails gen-check
https://github.com/istio/istio/pull/20533

I didn't look into if gen-check is calling more than gen, or if automated is not properly calling gen

clarketm:
1. Add support for supplying a custom **match title** (e.g. `--match-title`) *defaulting* to the value of **title** if unspecified.
2. Assign **title**, **match_title** and **body** in a child process to ensure only *exported* variables are evaluated. 
3. Revert **title** of istio/api dependency PRs to include **branch** rather than **sha**. This is for title matching on the PR - so the PR is *updated* rather of *created* on successive runs (since the sha changes on every commit and old PRs at older commits should be updated since they are no longer relevant). This also fixes the error with *PR exists conflict*: https://storage.googleapis.com/istio-prow/logs/gen_istio_api_postsubmit/47/build-log.txt

close https://github.com/istio/istio/issues/20534


fejta (approved):
/hold

PR #2354, opened 2020-02-07
add job for release-1.5 benchmark job

richardwxn:


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -155,6 +155,34 @@ periodics:
...
+        env:
+          - name: GIT_BRANCH
+            value: release-1.5
+        command:
+          - entrypoint
+          - perf/benchmark/run_benchmark_job.sh
+
 - cron: "0 16 * * *"  # starts every day at 16:00PM UTC
   name: daily-performance-benchmark-release-1.4
   branches: release-1.4
```
carolynhu, line 188: I think maybe we can get rid of release-1.4.

richardwxn: we probably should keep it as we need to compare data between two releases.

ericvn (approved):

PR #2359, opened 2020-02-11
Upgrade gke-e2e-test GKE version

clarketm:
@carolynhu / @richardwxn / @howardjohn

The *default* for new clusters is now **1.14**. Can one of you please confirm the test(s) using these boskos resources do not explicitly require k8s features exclusive to **1.13** (and approve if safe).


howardjohn (approved):
there is only one way to find out

howardjohn:
I am about 50% sure we can completely remove these resources within the next 2 weeks so not terribly worried about this change

PR #2366, opened 2020-02-12
Add sa to GCR deploy jobs

clarketm:
Fix auth when pushing images to GCR by adding serviceaccount (SA).

e.g.
https://prow.istio.io/view/gcs/istio-prow/logs/push-genjobs_test-infra_postsubmit/1

### DIFF
--- a/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
+++ b/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
@@ -127,6 +127,8 @@ postsubmits:
     - ^master$
     cluster: test-infra-trusted
     decorate: true
+    labels:
+      preset-service-account: "true"
     max_concurrency: 1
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
```diff
@@ -127,6 +127,8 @@ postsubmits:
     - ^master$
     cluster: test-infra-trusted
     decorate: true
+    labels:
+      preset-service-account: "true"
```
fejta, line 131: I disagree with using presets, please just add the concrete fields to the pods.

clarketm: I will update the generation logic. It might be more appropriate to break this out into another PR, since the change will affect several jobs unrelated to this particular change.

cjwagner (approved):

PR #2375, opened 2020-02-13
Add update commonfiles job for istio/proxy

clarketm:
The `update-common` target needs an update in **istio/proxy**. A `lint` target checks that the *proxy build-tools* and *build-tools* are the same, so we need to ensure that the *proxy build-tools* tag gets updated when `update-common` is executed.

https://github.com/istio/proxy/pull/2611
https://github.com/istio/proxy/pull/2674

Also, we may also want to run `update-common-protos` as well?

cc @bianpengyuan  

/hold

### DIFF
--- a/prow/cluster/jobs/istio/common-files/istio.common-files.master.gen.yaml
+++ b/prow/cluster/jobs/istio/common-files/istio.common-files.master.gen.yaml
@@ -118,6 +118,51 @@ postsubmits:
       - name: github
         secret:
           secretName: oauth-token
+  - annotations:
+      testgrid-alert-email: istio-oncall@googlegroups.com
+      testgrid-dashboards: istio_common-files_postsubmit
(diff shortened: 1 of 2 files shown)


bianpengyuan:
Yeah, I could not find a clean way to make `update-common` special for proxy..

clarketm:
> Yeah, I could not find a clean way to make `update-common` special for proxy.

I will look into it. Is there a GH issue/PR to track the attempt to make this work? Ideally, we would not make it special to proxy but rather generically support image overrides.

bianpengyuan:
No, I've not yet created an issue for it.

clarketm:
> No, I've not yet created an issue for it.

np. I will create one then see what we can do to get `update-common` working (properly) for proxy.

clarketm:
This change was included with #2374. This job will fail until `update-common` target is resolved.

PR #2391, opened 2020-02-18
Add istio.io tests to prow

nmittler:


prow/config/jobs/istio.io.yaml
```diff
@@ -6,3 +6,15 @@ image: gcr.io/istio-testing/build-tools:master-2020-02-14T13-09-14
 jobs:
   - name: lint
     command: [make, lint]
+
+  - name: k8s-tests
+    type: presubmit
+    command: [entrypoint, prow/integ-suite-kind.sh, test.kube.presubmit]
+    requirements: [kind]
+    modifiers: [optional]
```
howardjohn: Given that I don't think this will work out of the box (see some comments on the PR, I think the plan was to follow up on them later -- correct me if I am wrong), we may want to mark this as "hidden" as well. That way it will run but not show up in github UI. you can still see test results on prow.istio.io. The benefit is you don't cause confusion for developers when suddenly their PRs are failing tests and they can't fix it. To do so:

```suggestion
    modifiers: [optional, hidden]
```

nmittler: I think the hope is that it won't be long until they're working. The plan is to follow this up with an istio.io PR to get things green (or at least intermittently green :)). Having the link available on the PR is pretty handy to this end. Most Istio devs are reasonably use to seeing failing tests that are marked optional. Granted, istio.io has a slightly different audience but like I said, it should (hopefully) be fixed soon.

howardjohn: > Most Istio devs are reasonably use to seeing failing tests that are marked optional

I personally don't think this is a good practice. It adds a lot of noise, and I haven't seen many people understand the difference between required vs optional tests. I'll approve this with a hold though if you disagree. My thought process is the only person who cares about the test (until its working) is probably you, so its not too onerous to search out the test manually

howardjohn (approved):
/hold

feel free to remove

nmittler:
@howardjohn PTAL

prow/config/jobs/istio.io.yaml
```diff
@@ -6,3 +6,31 @@ image: gcr.io/istio-testing/build-tools:master-2020-02-14T13-09-14
...
+
+resources:
+  default:
+    requests:
+      memory: "3Gi"
+      cpu: "5000m"
+    limits:
+      memory: "24Gi"
+      cpu: "8000m"
+  lint:
```
howardjohn, line 30: this isn't used, since no job references `resources: lint`. So only the default mentioned above will apply to the two new jobs you added as well as lint. Seems reasonable to have just the default

howardjohn (approved):

PR #2396, opened 2020-02-20
Make istio.io tests required (on istio.io repo)

nmittler:


howardjohn (approved):

howardjohn:
approved assuming we keep the hold until it merges+pass postsubmit

PR #2406, opened 2020-02-21
Removing defunct E2E tests from CNI

stewartbutler:
These E2E tests were broken by upstream changes, and are duplicated in upstream istio/istio tests anyways. Fixing them would require some significant changes in 1.5 and the helm chart they are testing will be removed shortly in master (1.6).


howardjohn (approved):

howardjohn:
these have testing upstream so we will catch errors before we ship, so disabling is not too bad

PR #2408, opened 2020-02-21
Revert "Upgrade boskos"

clarketm:
Reverts istio/test-infra#2358. This version is invalid.


carolynhu (commented):
This PR is to revert boskos changes. Since after we set 3 perf clusters, we still facing same issue:https://storage.googleapis.com/istio-prow/logs/daily-performance-benchmark-release-1.5/13/build-log.txt  `current-context` is not set.

howardjohn (approved):

PR #2409, opened 2020-02-21
Enable branch protection envoy

clarketm:
fix https://prow.istio.io/view/gcs/istio-prow/logs/ci-test-infra-branchprotector/1168990827033141260

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -230,6 +230,8 @@ branch-protection:
             endpoints-v1.2.x: *blocking_merge
         test-infra:
           branches: *blocked_branches
+        envoy:
+          branches: *blocked_branches
     istio-releases:
       repos:
         pipeline:


howardjohn (approved):

fejta (approved):

fejta:
How about we just protect everything except whitelisted things?

clarketm:
> How about we just protect everything except whitelisted things?

What do you mean, exactly? Like the inverse of this configuration?

PiotrSikora:
Can we remove protection for the `master` branch? We sync/push `master` directly from [envoyproxy/envoy-wasm](https://github.com/envoyproxy/envoy-wasm) to [istio/envoy](https://github.com/istio/envoy), since we just mirror that branch, and only do PRs for the released branches.

Alternatively, automatically mirroring the `master` branch from there would also work.

clarketm:
> Can we remove protection for the `master` branch? We sync/push `master` directly from [envoyproxy/envoy-wasm](https://github.com/envoyproxy/envoy-wasm) to [istio/envoy](https://github.com/istio/envoy), since we just mirror that branch, and only do PRs for the released branches.

Sure, I will update then send out PR.

> Alternatively, automatically mirroring the `master` branch from there would also work.

This sounds like a good idea going forward. Creating a job for this should be straightforward; I will look into it.

PR #2411, opened 2020-02-22
Increase cpu/mem for istio/{proxy,envoy}

clarketm:
Higher requests for **istio/proxy** and **istio/envoy** so it gets exclusive access to node.

> Note: this is the current max until quota is increase and node pool is migrated

### DIFF
--- a/prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.gen.yaml
@@ -40,8 +40,8 @@ postsubmits:
             cpu: "16"
             memory: 60Gi
           requests:
-            cpu: "8"
-            memory: 8Gi
+            cpu: "15"
+            memory: 55Gi
         securityContext:
           privileged: true
         volumeMounts:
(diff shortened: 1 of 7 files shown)


howardjohn (commented):
if you request 16 core and the node is 16 cores, I think it may never schedule since there are some daemonsets potentially? I may be wrong here, but if not 15 CPUs should do the trick?

PiotrSikora (approved):
Thanks!

clarketm:
> if you request 16 core and the node is 16 cores, I think it may never schedule since there are some daemonsets potentially? I may be wrong here, but if not 15 CPUs should do the trick?

I was looking for docs for confirmation but couldn't find any. I will change to `15` to be safe. Thanks. Is this also the case for memory? I will lower that to `55Gi` to be safe as well.

We should know fairly quickly if this is insufficient if the pod is stuck in `pending` state.

howardjohn (approved):

PiotrSikora:
It doesn't look that those jobs are started anymore, so we probably need to put more than 1 vCPU aside.

clarketm:
> It doesn't look that those jobs are started anymore, so we probably need to put more than 1 vCPU aside.

Yep, the pod is saying insufficient cpu and memory: https://github.com/istio/test-infra/pull/2413

Issue #2385, opened 2020-02-14
Enable presubmits for istio/envoy


clarketm:
cc @PiotrSikora @howardjohn

PiotrSikora:
This should be as easy as calling:
```
bazel build --config=libc++ //source/exe:envoy
bazel test --config=libc++ //test/...
```

clarketm:
> This should be as easy as calling:
> 
> ```
> bazel build --config=libc++ //source/exe:envoy
> bazel test --config=libc++ //test/...
> ```

@PiotrSikora 

Rather than performing a direct `bazel` build and test, can we leverage some of the *existing* test scripts/targets in [do_ci.sh](https://github.com/istio/envoy/blob/master/ci/do_ci.sh) or is that specific to upstream envoy or Azure CI? 

Also, what are the environmental requirements? I see a reference to the `envoyproxy/envoy-build` docker image in the [ci/README.md](https://github.com/istio/envoy/blob/master/ci/README.md#build-image-base-and-compiler-versions).

PiotrSikora:
Hmm, you could use that indeed (or better: [`do_circle_ci.sh`](https://github.com/istio/envoy/blob/master/ci/do_circle_ci.sh) wrapper around it).

PiotrSikora:
Thanks, @clarketm, this is great!

clarketm:
I will add tests for `bazel.tsan` and `bazel.release` as well.

PiotrSikora:
Actually, any chance that you could bump those tests to 16 or more vCPUs? It's incredibly slow with 8 (effectively 6) vCPUs.

PiotrSikora:
Also, ASan didn't finish yet, but I suspect it might run out of disk space before it's done.

PiotrSikora:
Yeah, ASan builds needs a little over 300GB (and growing).

clarketm:
@PiotrSikora 

We have full capability to refine the resources as needed for **istio/envoy** whether that be: 
* adjust the kubernetes resource request; 
* modify bazel `BAZEL_BUILD_EXTRA_OPTIONS `;
* run in a separate node pool;
* or a combination of the aforementioned.

Let me know what is needed for the jobs and I can make the changes.

PiotrSikora:
We definitely need more disk space for `istio/envoy`. I'd go with 512GB SSD disk, but I suspect that ~400GB should also work.

[...]

PR #2414, opened 2020-02-22
Disable some istio/envoy tests

howardjohn:


howardjohn:
for https://github.com/istio/test-infra/issues/2385

howardjohn:
force merge since I think this is impact other jobs -- these tests took 2x longer than normal

PiotrSikora:
Thanks!

PR #2419, opened 2020-02-23
Increase istio/{envoy,proxy} cpu/memory resources

clarketm:
I reshaped the node pool to [n1-standard-32](https://cloud.google.com/compute/docs/machine-types): 32 vCPUs; 120GB Memory; 512GB Disk.  Let me know if the following job(s) need additional tweaks or if we want even more resources.

ref: https://github.com/istio/test-infra/issues/2385

### DIFF
--- a/prow/config/jobs/envoy.yaml
+++ b/prow/config/jobs/envoy.yaml
@@ -9,7 +9,7 @@ jobs:
 #   type: presubmit
 #   env:
 #   - name: BAZEL_BUILD_EXTRA_OPTIONS
-#     value: "--local_ram_resources=32768 --local_cpu_resources=12 --test_env=ENVOY_IP_TEST_VERSIONS=v4only"
+#     value: "--local_ram_resources=65536 --local_cpu_resources=28 --test_env=ENVOY_IP_TEST_VERSIONS=v4only"
 #   - name: ENVOY_SRCDIR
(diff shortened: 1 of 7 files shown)


prow/config/jobs/envoy.yaml
```diff
@@ -36,8 +36,8 @@ jobs:
...
-      memory: "50Gi"
-      cpu: "14000m"
+      memory: "110Gi"
+      cpu: "30000m"
     limits:
-      memory: "60Gi"
-      cpu: "16000m"
\ No newline at end of file
+      memory: "120Gi"
+      cpu: "32000m"
```
howardjohn, line 43: Outside the scope of this PR: I am wondering if we should reconsider how we use limits. See https://mcrthr.com/kubernetes-cpu-limits (and we do have an affected kernel version - https://cloud.google.com/container-optimized-os/docs/release-notes/m69#cos-69-10895-348-0). I suspect if we just make the requests what we want to ensure jobs get, then limits are just needlessly throttling in most or all cases

howardjohn (approved):

PiotrSikora:
Those looks good, thanks! We can enable the disabled jobs with this, if you want, since they should all pass now.

PiotrSikora (approved):

PiotrSikora:
Merging this now, I'll send a separate PR to re-add other test targets. Thanks!

PR #2422, opened 2020-02-24
Try out removing CPU limits on istio/istio

howardjohn:
Some context:
https://github.com/istio/test-infra/pull/2419#discussion_r383018567

The goal with our configuration is just to guarantee that each job has
at least 5 CPUs (which is itself a fairly arbitrary value - may benefit
from tuning). Adding a limit just introduces needless throttling due to
kernel bugs and/or due to artificial limits. If the pod has exclusive
access to the node, I see no strong reason to not let it use all 32
cores.

If this ends up being a horrible idea we can always revert


clarketm (approved):
/hold

PR #2446, opened 2020-02-28
Genjobs: support reading from a yaml configuration file(s)

### LINKED ISSUE #2161: Genjobs: additionally support a yaml configuration file

clarketm:
fix https://github.com/istio/test-infra/issues/2161 and also enabled RBE on rest of `istio-private/proxy` jobs.

> In the near future, `genjobs` will be subsumed into the `transform` subcommand of the [ProwJob CLI](https://docs.google.com/document/d/1bUNo7mpERzwMerDkqwB2dA2PMThwURgLyoIr1Awse5w/edit?usp=sharing) (draft).


prow/config/istio-private_jobs/00-clean.yaml
```diff
@@ -0,0 +1,13 @@
+common_opts: &common_opts
```
howardjohn, line 1: why do we need this if its only used one time?

prow/config/istio-private_jobs/istio.yaml
```diff
@@ -0,0 +1,37 @@
+common_jobs_mods: &common_jobs_mods
```
howardjohn, line 1: if we want to avoid yaml magic, you could add a native `defaults` field that does this. Up to you

howardjohn (approved):

prow/config/istio-private_jobs/00-clean.yaml
clarketm: You are right, it is not needed. I originally had all the transforms in a single file. I missed that this particular anchor was used only once.

prow/config/istio-private_jobs/istio.yaml
clarketm: I like the idea. I am currently doing research into various differents yaml processors, mergers, and templating tools (e.g. kustomize, helm, quicktemplate, ytt, aviator, jsonnet, yq, spruce, etc.) that may help simply the configuration.

Issue #2425, opened 2020-02-24
Start using a dependency and/or build cache in CI.


clarketm:
Start using a shared dependency and/or build cache for `bazel` and `go` to improve build and test efficiency in CI.

howardjohn:
Testing go.mod and build caching:

go mod download
* No cache: 1min
* Cold cache: 2m48s (slower because the disk is slower)
* warm cache: 2s


For build
* no cache: 54s
* cold cache: 58s
* warm cache: 20s

cost: base is $200/month for 1TB at standard speeds. We can also do $750 for premium which is 2.5TB and quite a bit faster

One build uses 500mb in go build cache, 1.3gb module caching. 

> The go command periodically deletes cached data that has not been used recently.

I don't think there is automatic module cache cleanup.

alternative for go modules is we can pre-warm the module cache inside the docker image itself, at the cost of +1.3gb to the image


This is all done with Google Filestore which is a managed NFS server. Test yamls here:

```yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: shell
spec:
  selector:
    matchLabels:
      app: shell
  template:
    metadata:
      labels:
        app: shell
    spec:
      containers:
      - name: shell
        imagePullPolicy: Always
        image: gcr.io/istio-testing/build-tools:master-2020-02-14T13-09-14
        args:
          - /bin/sleep
          - infinity
        volumeMounts:
        - mountPath: /go/pkg
          subPath: gomod
          name: mypvc
        - mountPath: /gocache
          subPath: buildcache
          name: mypvc
      volumes:
      - name: mypvc
        persistentVolumeClaim:
          claimName: fileserver-claim
          readOnly: false
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: fileserver
spec:
  capacity:
    storage: 1T
  accessModes:
  - ReadWriteMany
  nfs:
    path: /gomodtest1
    server: IP-OF-FILESTORE
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: fileserver-claim
spec:
  accessModes:
  - ReadWriteMany
  storageClassName: ""
  volumeName: fileserver
  resources:
    requests:
      storage: 5Gi
```


Basically we can have one filestore, and then have subdirectories for each cache we want (go mod, go build, bazel, ...)

howardjohn:
[...]

PR #2450, opened 2020-03-02
Add experimental build cache

howardjohn:
@clarketm do we automatically deploy these or will I have to manually run this once merged?

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -82,6 +82,53 @@ postsubmits:
       preset-enable-ssh: "true"
       preset-override-deps: release-1.4-istio
       preset-override-envoy: "true"
+    name: cache-experiment_istio_postsubmit_priv
+    path_alias: istio.io/istio
+    skip_report: true
+    spec:
+      containers:
+      - command:
+        - entrypoint
+        - make
+        - -e
+        - T=-v
+        - build
+        - racetest
+        - binaries-test
+        image: gcr.io/istio-testing/build-tools:master-2020-02-14T13-09-14
+        name: ""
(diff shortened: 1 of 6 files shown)


howardjohn:
for https://github.com/istio/test-infra/issues/2425

clarketm:
> @clarketm do we automatically deploy these or will I have to manually run this once merged?

This will autodeploy.

prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
```diff
@@ -71,6 +71,53 @@ postsubmits:
           privileged: true
       nodeSelector:
         testing: test-pool
+  - annotations:
```
clarketm: Since this will also be used in the _private_ prow cluster, you probably want to have a symbolic link from `prow/cluster/build/buildcache.yaml` -> `prow/cluster/private/buildcache.yaml` to ensure the PV and PVC are deployed in the _private_ cluster as well (or blacklist this job in [istio.yaml](https://github.com/istio/test-infra/blob/master/prow/config/istio-private_jobs/istio.yaml) if we do not want it there yet).

clarketm (changes requested):

prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
howardjohn: Can we just drop this in the private one? it will make things much easier, since it will allow other prow instances to utilize it without a cache as well. For the short term we can blacklist the job, but long term this will be used for most jobs

howardjohn: Ill blacklist for now

clarketm (approved):

PR #2453, opened 2020-03-02
Fix namespace for cache PV/PVC

howardjohn:


clarketm (approved):

howardjohn:
Do we need to remove the existing PV/PVC?

clarketm:
> Do we need to remove the existing PV/PVC?

Yes, deletes are manual. I just deleted them.

PR #2455, opened 2020-03-02
Prow: support cluster object pruning on apply

clarketm:


clarketm:
`--prune` is incomplete and does not work without explicit whitelist: https://github.com/kubernetes/kubernetes/pull/35220. Will either find an alternative or update with a comprehensive whitelist.

---
**Update:**

Use `kubectl` to dynamically retrieve api resources and provide the output to the `apply` as `--prune-whitelist` args.

fejta (approved):
/hold

PR #2456, opened 2020-03-02
Fix gopkg path in cache

howardjohn:


clarketm (dismissed):

howardjohn:
@clarketm any idea how to workaround the mount conflicts? Seems go does not support changing the location of pkg cache

clarketm:
> @clarketm any idea how to workaround the mount conflicts? Seems go does not support changing the location of pkg cache

The validation failure is a result of [validatePodSpec](https://github.com/kubernetes/test-infra/blob/ca9526d747fd5d2866ba6a13956c82e7a6bd3c8a/prow/config/config.go#L1807-L1811). Assuming it will not cause problems to mount at a nested path we can consider modifying this logic. I understand rejecting identical mount paths but parent/child paths should be fine? There may be reasons, so I would like to get weigh-in from @cjwagner.

prow/config/generate.go
```diff
@@ -581,7 +581,7 @@ func applyRequirements(job *config.JobBase, requirements []string) {
 			)
 			job.Spec.Containers[0].VolumeMounts = append(job.Spec.Containers[0].VolumeMounts,
 				v1.VolumeMount{
-					MountPath: "/go/pkg",
+					MountPath: "/home/prow/go/pkg",
 					Name:      "build-cache-pvc",
 					SubPath:   "gomod",
```
clarketm, line 586: I missed this before, but should this be `mod` instead of `gomod`?

howardjohn: this part doesn't matter, its just the path on the actual NFS drive. as long as its consistent its fine, the mountPath is what matters

cjwagner:
>  Assuming it will not cause problems to mount at a nested path we can consider modifying this logic.

If nested volumes are supported by kubernetes I think that change would be fine, but we should ensure that if users accidentally specify a mount that conflicts with Prow's pod utilities the problem is obvious.
What if there is a directory in the parent mount that conflicts with the child mount?

clarketm:
~WIP~ PR: https://github.com/kubernetes/test-infra/pull/16580

> If nested volumes are supported by kubernetes I think that change would be fine, 

Nested volume mount paths are supported, while [duplicate volume mount paths](https://github.com/kubernetes/kubernetes/blob/79e1ad2f4bbd05b1e56b7b57b63b2c1d67b90156/pkg/apis/core/validation/validation.go#L2349-L2351) are prohibited. The latter validation will be retained.

> but we should ensure that if users accidentally specify a mount that conflicts with Prow's pod utilities the problem is obvious.

Would logging a warning to the Prow logs be a reasonable way to achieve this? Or a flag to checkconfig to inform the user?

> What if there is a directory in the parent mount that conflicts with the child mount?

The behavior depends on the type of volume(s) involved and their characteristics (e.g. read-only, etc.). In the scenarios I ran using various RW volumes, the writes persisted on the *more specific* volume mount path (i.e. volume backing the child mount path) even when run bidirectionally to rule out one volume type mounting faster than another. ~I am not certain if volume mount order is deterministic though (https://github.com/kubernetes/kubernetes/issues/68249)~. Indeed, it looks like the volume mount order is based on path depth for both the Docker and Containerd container runtimes (ref: https://github.com/containerd/cri/pull/892), which is *desired* behavior since we care about persistence to the child `PV` (i.e filestore) but not to the parent `emptyDir` (which is ephemeral anyway).

clarketm (approved):
dependent on https://github.com/kubernetes/test-infra/pull/16580

howardjohn:
Do we need to update go.mod dependency as well?

clarketm:
@howardjohn - tests passing. unhold at your convenience.

howardjohn:
/hold cancel

Thanks for your help pushing this through!

PR #2459, opened 2020-03-04
Generate: support generation of Periodics

clarketm:
And generate the `istio.io`/`update-ref-docs` job.


prow/config/generate.go
```diff
@@ -173,7 +178,22 @@ func ValidateJobConfig(jobConfig JobConfig) {
 				err = multierror.Append(err, e)
 			}
 		}
-		if e := validate(job.Type, []string{TypePostsubmit, TypePresubmit, ""}, "type"); e != nil {
+		if job.Type == TypePeriodic {
```
clarketm: We should consider importing [this](https://github.com/kubernetes/test-infra/blob/8b3bfc8b7a2f4d0a2dbcc236689497cb439deec3/prow/config/config.go#L1293-L1309) (along with other) validation logic from Prow if feasible.

howardjohn (approved):

PR #2471, opened 2020-03-06
add timeout for benchmark job

richardwxn:
now since we have more configs including stackdriver filters, we need more time for the job. However at the same time I will try to investigate how to parallel the different configs to reduce running time.

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -35,9 +35,9 @@ periodics:
   name: daily-performance-benchmark
   branches: master
   decorate: true
-  timeout: 17h
+  timeout: 24h
   decoration_config:
-    timeout: 17h0m0s
+    timeout: 24h0m0s
   extra_refs:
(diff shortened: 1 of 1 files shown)


clarketm (approved):

prow/cluster/jobs/all-periodics.yaml
```diff
@@ -63,9 +63,9 @@ periodics:
   name: daily-performance-benchmark-release-1.5
   branches: release-1.5
   decorate: true
-  timeout: 17h
+  timeout: 24h
   decoration_config:
-    timeout: 17h0m0s
+    timeout: 24h0m0s
```
carolynhu, line 68: Yeah, as we add more tests, we need to extend the timeouts. Well, since we are now only running two perf jobs, i.e. release-1.5 and master and we have three clusters https://github.com/istio/test-infra/blob/master/boskos/resources.yaml#L48. We should  still fine at current stage. But, yeah, nice to do some optimizations later on.

PR #2473, opened 2020-03-06
Genjobs: support local and global defaults

clarketm:
see: https://github.com/istio/test-infra/pull/2446#discussion_r386035795

### DIFF
--- a/prow/config/istio-private_jobs/istio_1.4.yaml
+++ b/prow/config/istio-private_jobs/istio_1.4.yaml
@@ -1,24 +1,12 @@
-common_jobs_mods: &common_jobs_mods
-  mapping:
-    istio: istio-private
-  input: ./prow/cluster/jobs/
-  output: ./prow/cluster/jobs/
+defaults:
   branches: [release-1.4]
   repo-whitelist: [istio]
-  ssh-clone: true
-  extra-refs: true
-  cluster: private
-  modifier: priv
-  annotations:
-    testgrid-create-test-group: "false"
-  verbose: false
 
 
 transforms:
 
 # istio/istio master build job(s) - postsubmit(s)
-- <<: *common_jobs_mods
-  env:
+- env:
     DOCKER_HUB: gcr.io/istio-prow-build
     GCS_BUCKET: istio-private-build/dev
   labels:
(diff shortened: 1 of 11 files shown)


prow/config/istio-private_jobs/istio_1.4.yaml
```diff
@@ -1,24 +1,12 @@
-common_jobs_mods: &common_jobs_mods
-  mapping:
-    istio: istio-private
-  input: ./prow/cluster/jobs/
-  output: ./prow/cluster/jobs/
+defaults:
```
fejta, line 1: If we have defaults can we start eliminating the presets?

clarketm: No, the `defaults` keys are the same as those for the `transforms`. This is merely a mechanism to prevent duplication at file-level not eliminate need for presets. FWIW, the next generation (`v1.0.0`) of this tool will not use presets but it is still being written.

fejta: Awesome! What is the eta for v1 that doesn't use presets?

clarketm: I am working on the job `creation` now, then will tackle the `transformation` with the former being a replacement for `generate.go` and latter being the replacement for `genjobs.go` effectively unifying the tools. I estimate days to weeks depending on the time I dedicate to it; the [design](https://docs.google.com/document/d/1bUNo7mpERzwMerDkqwB2dA2PMThwURgLyoIr1Awse5w/edit?usp=sharing) is still incomplete and not yet open for review (tbh, I am intentionally iterating slowly and trying to incorporate best practices). I will backport the working POC code to the design doc once it satisfies the requirements of both `creation` and `transformation` and is generic enough for both Istio and non-Istio use cases.

Instead of presets, it will accept  _requirements_ either directly in the job config entry or via a list of keys where each key maps to a _requirement_ in a config file (similar to the concept of defining presets). The difference between the latter and presets is that these are resolved at the creation time of _PJ config_ rather than the creation time of _PJ object_.

This is subject to change, but a _requrement_ is defined similarly to the following:

```golang
type JobRequirement struct {
	metav1.ObjectMeta
	corev1.Container
	corev1.PodSpec
}
```

For the first iteration, the fields will be pruned to the common fields found in a PJ spec (e.g. `VolumeMounts`, `Resources`, `Volumes`, `SecurityContext`, etc) so not _everything_ in a `corev1.PodSpec`, for example, will be valid.

fejta (approved):

PR #2481, opened 2020-03-07
Automator: improve and simplify error handling

clarketm:
- Add `print_error()` utility function to print errors while retaining return code.
- Additional logging when are error does occur.
- PRs for a particular _repo_ will **not** be created or update if an error occurs.
- _Any_ error in `work()` returns early and execution proceeds to the next `$repo`
- General simplification of `bash` logic using `set -e` and less conditionals inside `work()`.

... otherwise the behavior is _exactly_ the same as before.

### DIFF
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -200,33 +200,35 @@ add_labels() {
 }
 
 commit() {
-  git add --all &&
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

howardjohn:
We can archive/delete/remove all config for the istio-release org if it helps. We do not use it now

PR #2492, opened 2020-03-10
bp: allow force push istio-private

clarketm:
**feature:** 
https://github.com/kubernetes/test-infra/pull/16669

**rationale:** 
https://github.com/kubernetes/test-infra/issues/16077

More specifically, the motivation is to simplify the private repository syncing process: 

> [Synching Repositories](https://github.com/istio-private/istio/wiki#synching-repositories)

> [Managing Private/Security Build Branches](https://docs.google.com/document/d/19qf0-0GIopEZZWuG-lj8_VpyQF4N9nIKXiZgcJHwBRM/edit?usp=sharing)

---

The preferred approach to keep the private repositories updated is to clone a bare copy of the public repo and mirror it to the private repo. There is a script to do this.

> Note: this will overwrite a private repo to the current state of the public repo for every ref.

> Note: prow turns on branch protection automagically which prevents this script from working. Before one needed to first visit the branch protection pages of all the istio-private repos (e.g., https://github.com/istio-private/release-builder/settings/branches) and turn off branch protection on master and on the release-1.* branches.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -281,6 +281,7 @@ branch-protection:
             master:
               protect: true
     istio-private:
+      allow_force_pushes: true
       required_pull_request_reviews:
         <<: *protection_required_pull_request_reviews
         require_code_owner_reviews: false


howardjohn (approved):

fejta:
Can you update the PR description to say why you're allowing this?

AKA this says what the PR does but not why

clarketm:
> Can you update the PR description to say why you're allowing this?
> 
> AKA this says what the PR does but not why

My fault for not linking the info. Updated the description with links and rationale for the change.

fejta:
a) The rationale seems pretty circular ("we need to allow force pushing so people can force push if they want")
b) Can you put the rationale on this PR?

clarketm:
> a) The rationale seems pretty circular ("we need to allow force pushing so people can force push if they want")
> b) Can you put the rationale on this PR?

The context from the links and blockquote should make it clear? 

`istio-private` developers (ones with write access) need to be able to force push to protected branches for repository synchronization from `istio/$repo` -> `istio-private/$repo`.

fejta:
I'm asking for a sentence in the PR description. Perhaps there is one in https://github.com/istio-private/istio/wiki#synching-repositories? But try opening that in an incognito window.

Here's some guides for how things ideally look:
* https://dev.to/jacobherrington/how-to-write-useful-commit-messages-my-commit-message-template-20n9
* https://chris.beams.io/posts/git-commit/

I don't think you need to apply all the rules in these documents, and I'm not particular about whether the message is in the git commit vs the PR, but -- especially since force pushes are dangerous and the private repo is generally for security purposes -- I would like to see an at least minimal rationale in the PR or git commit.

It should not require multiple days and rounds of reviews to get a sentence in the PR description or git commit explaining the motivation here.

> need to be able to force push to protected branches for repository synchronization from istio/$repo -> istio-private/$repo.

Force push enables/allows erasure history. This is often necessary but also dangerous: I could use it to revise a 2018 to now contain a vulnerability that I can then exploit.

Normal activity simply appends to the git history and does not require erasure.

So what is the motivation here for revising history? I assume they keys are something along the lines of "commits still under development" and "rebasing". I would like this to be well understood and called out in the PR.

"I want to enable force pushes so developers can force push" is insufficient justification.

clarketm:
Spoke to @fejta offline. The permittance of force push is likely not necessary with altered branch protection configuration. Will reevaluate the approach then reopen.

Issue #2282, opened 2020-01-17
Migrate to workload-identity


fejta:
Jobs running on GKE should use https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity to authenticate with GCP rather than by loading a json file from a secret

praveenbingo:
@howardjohn Does istio work with gke workload identity. I am running into issues trying to setup istio with gke workload identity (it is unable to use the credentials from the identity). 

I googled around a ton, but unable to find any documentation which clarifies the support either positively or negatively :)

howardjohn:
This is mostly done, through a variety of PRs  over the years. https://github.com/istio/test-infra/pull/4790 should be the last one

PR #2503, opened 2020-03-11
use correct requirements for mason image push

clarketm:
Pushing the `mason` image requires _docker_ and _GCR access_. 🤦‍♂

### DIFF
--- a/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
+++ b/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
@@ -196,7 +196,7 @@ postsubmits:
     cluster: test-infra-trusted
     decorate: true
     labels:
-      preset-prow-deployer-service-account: "true"
+      preset-service-account: "true"
     max_concurrency: 1
     name: push-mason_test-infra_postsubmit
     path_alias: istio.io/test-infra
@@ -219,8 +219,14 @@ postsubmits:
             memory: 3Gi
         securityContext:
           privileged: true
+        volumeMounts:
+        - mountPath: /var/lib/docker
+          name: docker-root
       nodeSelector:
         prod: prow
+      volumes:
+      - emptyDir: {}
+        name: docker-root
   - annotations:
       testgrid-alert-email: istio-oncall@googlegroups.com
       testgrid-dashboards: istio_test-infra_postsubmit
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
```diff
@@ -196,7 +196,7 @@ postsubmits:
     cluster: test-infra-trusted
     decorate: true
     labels:
-      preset-prow-deployer-service-account: "true"
+      preset-service-account: "true"
```
cjwagner, line 199: This doesn't seem quite right. The `service-account` secret's SA shouldn't have permission to upload images? I guess the trusted cluster could have a different identity in the `service-account` secret, but this smells like either confusing naming or incorrect credentials.
/hold

cjwagner (approved):

prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
clarketm: The SA is `istio-prow-test-job@istio-testing.iam.gserviceaccount.com` and it is K8s admin, Storage owner (among other things). I will follow-up and create a properly _scoped_ `preset-prow-pusher-service-account` SA for image push jobs. It will be a challenge for another day to update the SA in the build cluster to strip away unnecessary perms.

fejta: Will you a) please use workload identity instead of a service account? b) not use a preset?

clarketm: > a: use workload identity

Workload identity (WI) is not something I have prioritized (or learned) yet. It is being tracked in: https://github.com/istio/test-infra/issues/2282 and recently labeled as _help wanted_ to solicit assistance. It should probably not be a prerequisite for fixing this one misconfigured job - since it will impact other jobs and clusters as well; unless, of course, there is a way to progressively rollout WI for this one job without affecting the others?

> b: do not use preset

I can "resolve" the value of this preset (and directly add the env,  mounts, etc.) instead of actually using the preset. And I will put it behind a flag for portability.

fejta (approved):
/hold

prow/cluster/jobs/istio/test-infra/istio.test-infra.master.gen.yaml
clarketm: /unhold
Support for requirement resolution and using "pusher" SA in: https://github.com/istio/test-infra/pull/2508

PR #2506, opened 2020-03-12
genjobs: replace refs when extra refs are used

howardjohn:
Otherwise, there is no way to override a ref

### DIFF
--- a/prow/genjobs/Makefile
+++ b/prow/genjobs/Makefile
@@ -14,7 +14,7 @@
 
 PROJECT = istio-testing
 HUB = gcr.io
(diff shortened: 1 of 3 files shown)


clarketm (approved):
lgtm and thanks! Can you increment version in [`Makefile`](https://github.com/istio/test-infra/blob/811663a21f6e27dd7265ccbce4085d017b2c5410/prow/genjobs/Makefile#L17)? And optionally add a note to [changelog](https://github.com/istio/test-infra/blob/master/prow/genjobs/README.md#changelog) that `extra-refs` is now replacive rather than additive (or something similar).

/hold

PR #2509, opened 2020-03-13
Authentikos: force refresh of cached token

clarketm:
#### History
The default token source for google oauth2 http client uses [`ReuseTokenSource`](https://github.com/roabay/util/blob/70cefad84e9cd69a508e81ff60dfa905aa2f231b/golang.org/x/oauth2/oauth2.go#L313-L340) which is employed to reuse tokens from a cache (such as a file on disk) between runs of a program, rather than obtaining new tokens unnecessarily. 

#### Change
Add a `-r`, `--reuse` option for reusing a cached token before it is expired. If this option is omitted or false, credentials will be forcibly refreshed even if not yet expired.

### DIFF
--- a/authentikos/README.md
+++ b/authentikos/README.md
@@ -39,6 +39,7 @@ The following is a list of supported options for `authentikos`:
   -c, --creds string           Path to a JSON credentials file.
   -k, --key string             Name of secret data key. (default "token")
   -n, --namespace strings      Namespace(s) to create the secret in. (default [default])
+  -r, --reuse                  Reuse a cached token until it is expired.
   -s, --scopes strings         Oauth scope(s) to request for token.
   -o, --secret string          Name of secret to create. (default "authentikos-token")
   -t, --template string        Template string for the token.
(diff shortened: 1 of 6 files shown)


authentikos/README.md
```diff
@@ -53,3 +54,4 @@ The following is a list of supported options for `authentikos`:
 - 0.0.3: add new `TimeToUnix`, `UnixToTime`, and `Parse` template variable and change method signature for math template variables from `(a, b time.Duration) time.Duration` to `(a, b int64) int64`.
 - 0.0.4: add `--key` option for specifying the name of the data key in the created Kubernetes secret.
 - 0.0.5: use [Sprig](http://masterminds.github.io/sprig/) as the library for template functions.
+- 0.0.6: add `--force-refresh` option for forcing a token refresh even if not yet expired. If this option is omitted or false, credentials will only refresh after expiration.
```
cjwagner: Should this be the default behavior?

authentikos/authentikos.go
```diff
@@ -218,23 +224,32 @@ func loadClusterConfig() (*rest.Config, error) {
...
 	clientOpts := []option.ClientOption{option.WithScopes(o.scopes...)}
 
 	if len(o.creds) > 0 {
 		clientOpts = append(clientOpts, option.WithCredentialsFile(o.creds))
 	}
 
-	client, err := transport.Creds(context.Background(), clientOpts...)
-	if err != nil {
-		return nil, err
+	if o.forceRefresh {
```
cjwagner: nit: A comment explaining how this forces a refresh would be helpful.

cjwagner (commented):
Please add some unit tests.

authentikos/README.md
clarketm: I prefer to be backwards compatible. Also, this is consistent with the default behavior of `gcloud config config-helper` and the oauth2 Go client library.

clarketm: Only because I like default booleans to be `false`, I renamed this to `--reuse` and made the default `false`, so force refreshing is now the default behavior.

clarketm:
Going to combine this with #2510 and have the refresh by _default_ be dynamic (i.e. lazy) based on comparing tick interval and token expiry. I will still add a flag to `--force-refresh` but the former will maximize performance as it leverages the cache as much as possible.

PR #2511, opened 2020-03-13
Authentikos: intelligent token refresh

clarketm:
#### Background
The default token source for google oauth2 http client uses [`ReuseTokenSource`](https://github.com/golang/oauth2/blob/aaccbc9213b0974828f81aaac109d194880e3014/oauth2.go#L354-L381) which is employed to reuse tokens from a cache (such as a file on disk) between runs of a program, rather than obtaining new tokens unnecessarily. 

Unfortunately, in **Authentikos**, this _will_ result in a token expiring before it has been refreshed in the k8s secrets due to the tick-based reconciliation loop. Currently, since the refresh interval is `30m` and the default oauth2 token expiration is `60m` (w/ a `10s` delta), in most cases the token will only be expired for ~seconds before being refreshed, minimizing the impact. Regardless, this can easily be fixed by introspecting the token expiry and force refreshing the token as needed (see below). 

#### Changes

1. Add a `-r`, `--force-refresh` option for forcing a token refresh. (_default_: credentials will only refresh when necessary (i.e. cached tokens will be returned when possible)).

> **NOTE:** the default here is also _new_, improved behavior. The token will refresh as needed by inspecting the expiry of the token and comparing it with the time of next token refresh (i.e. reconciliation interval).

2. Add. a `-i`, `--interval` option for customizing the token refresh interval (_default_: 30).

3. Set reasonable default scopes (i.e. the same ones as `gcloud auth application-default print-access-token`).

4. Tests are in a companion PR: https://github.com/istio/test-infra/pull/2512.

**Dependent on: https://github.com/istio/test-infra/pull/2512**

### DIFF
--- a/authentikos/README.md
+++ b/authentikos/README.md
@@ -37,9 +37,11 @@ The following is a list of supported options for `authentikos`:
 
 ```console
   -c, --creds string           Path to a JSON credentials file.
+  -r, --force-refresh          Force a token refresh. Otherwise, the token will only refresh when necessary.
+  -i, --interval duration      Token refresh interval [1m0s - 50m0s). (default 30m0s)
   -k, --key string             Name of secret data key. (default "token")
   -n, --namespace strings      Namespace(s) to create the secret in. (default [default])
-  -s, --scopes strings         Oauth scope(s) to request for token.
+  -s, --scopes strings         Oauth scope(s) to request for token (see: https://developers.google.com/identity/protocols/oauth2/scopes).
   -o, --secret string          Name of secret to create. (default "authentikos-token")
   -t, --template string        Template string for the token.
   -f, --template-file string   Path to a template string for the token.
@@ -48,8 +50,9 @@ The following is a list of supported options for `authentikos`:
 
 ## Changelog
 
-- 0.0.1: initial release
-- 0.0.2: remove `--format` option and add `--template` and `--template-file` options.
-- 0.0.3: add new `TimeToUnix`, `UnixToTime`, and `Parse` template variable and change method signature for math template variables from `(a, b time.Duration) time.Duration` to `(a, b int64) int64`.
-- 0.0.4: add `--key` option for specifying the name of the data key in the created Kubernetes secret.
-- 0.0.5: use [Sprig](http://masterminds.github.io/sprig/) as the library for template functions.
+- 0.0.1: Initial release
+- 0.0.2: Remove `--format` option and add `--template` and `--template-file` options.
+- 0.0.3: Add new `TimeToUnix`, `UnixToTime`, and `Parse` template variable and change method signature for math template variables from `(a, b time.Duration) time.Duration` to `(a, b int64) int64`.
+- 0.0.4: Add `--key` option for specifying the name of the data key in the created Kubernetes secret.
+- 0.0.5: Use [Sprig](http://masterminds.github.io/sprig/) as the library for template functions.
+- 0.0.6: Add `--force-refresh` option for forcing a token refresh. If this option is omitted or false, the token will only refresh when necessary. Add `--interval` option for customizing the token refresh interval. If unspecified, default scopes to _userinfo.email_, _cloud-platform_, and _openid_.
--- a/authentikos/authentikos.go
+++ b/authentikos/authentikos.go
@@ -23,6 +23,8 @@ import (
 	"fmt"
 	"html/template"
 	"io/ioutil"
+	"log"
+	"math"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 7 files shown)


authentikos/README.md
```diff
@@ -37,9 +37,11 @@ The following is a list of supported options for `authentikos`:
 
 ```console
   -c, --creds string           Path to a JSON credentials file.
+  -r, --force-refresh          Force a token refresh. Otherwise, the token will only refresh when necessary.
+  -i, --interval int           Token refresh interval in minutes [1 - 60). (default 30)
```
cjwagner: nit: Consider using a duration string instead. e.g. `2m`.

authentikos/authentikos.go
```diff
@@ -218,24 +247,45 @@ func loadClusterConfig() (*rest.Config, error) {
...
 	if len(o.creds) > 0 {
 		clientOpts = append(clientOpts, option.WithCredentialsFile(o.creds))
 	}
 
+	// Reusing the client leverages the token source cache.
 	client, err := transport.Creds(context.Background(), clientOpts...)
-	if err != nil {
-		return nil, err
+
+	clientCreator := func(reuse bool) (*google.Credentials, error) {
```
cjwagner: `reuse` is unused.

authentikos/authentikos.go
```diff
@@ -218,24 +247,45 @@ func loadClusterConfig() (*rest.Config, error) {
...
+		client, err := clientCreator(forceRefresh)
+		if err != nil {
+			return nil, err
+		}
 		token, err := client.TokenSource.Token()
 		if err != nil {
 			return nil, err
 		}
+		nextRec := time.Now().Add(time.Duration(o.interval) * time.Minute)
+		if token.Expiry.Round(0).Add(-expiryDelta).Before(nextRec) {
```
cjwagner: Why `Round(0)`?

authentikos/authentikos.go
```diff
@@ -218,24 +247,45 @@ func loadClusterConfig() (*rest.Config, error) {
...
+			// Force recreate the token if it will expire before the next reconciliation.
+			return tokenCreator(true)
```
cjwagner: This would recurse forever if we repeatedly generate tokens that expire before the next reconciliation for some reason. Is that something that could happen?

authentikos/authentikos.go
clarketm: Thanks, this should be `forceRefresh`.

authentikos/authentikos.go
clarketm: This is to use "wall clock" instead of "monotonic clock": https://github.com/golang/oauth2/commit/00dc70155e4c968f5afa403a799926838a4aa28f but since I do not really see this used anywhere else in production environments or in the google go client itself, I will remove it.

xref: https://github.com/golang/go/blob/cbcb031fe8b26be9d40c5fbb0c1934b0afdcf422/src/time/time.go#L46
xref: https://github.com/golang/oauth2/blob/bf48bf16ab8d622ce64ec6ce98d2c98f916b6303/token.go#L130

authentikos/authentikos.go
clarketm: It will not happen as long as the assumption that the expiry of new tokens is an hour since we limit the interval to ~60 minutes which is below the token expiration (including the delta for variance). Google does not really document this outside of the code itself. To be safe, I added retry logic with a max retry limit.

authentikos/README.md
clarketm: That would be a nice improvement. I will update!

cjwagner (approved):
LGTM. Holding in case you want to change to use duration strings in this PR.
/hold

authentikos/README.md
clarketm: Updated. Added unit tests to support this change as well.

authentikos/authentikos.go
```diff
@@ -218,24 +264,54 @@ func loadClusterConfig() (*rest.Config, error) {
...
+			return withBackoff(1, maxTries-tries, create).(tokenCreator)(forceRefresh, tries-1)
+		}
+
 		token, err := client.TokenSource.Token()
 		if err != nil {
-			return nil, err
+			return withBackoff(1, maxTries-tries, create).(tokenCreator)(forceRefresh, tries-1)
+		}
+
+		nextRec := time.Now().Add(o.interval * time.Minute)
```
cjwagner: This needs to be updated now that the interval has changed types (multiplying by `time.Minute` is now wrong). Please make sure this is covered by unit or integration checks.

clarketm: Good catch. I fixed and added unit tests for the expiration checking logic. Will contrive a integration test(s) in another PR to handle the more complex logic.

PR #2515, opened 2020-03-13
[WIP] add presubmit job for perf pipeline changes

carolynhu:
put on hold, until I've done the tools part of those script and config files.

### DIFF
--- a/prow/config/jobs/tools.yaml
+++ b/prow/config/jobs/tools.yaml
@@ -30,6 +30,12 @@ jobs:
     regex: 'docker/.+|cmd/.+'
     requirements: [gcp, docker]
 
+  - name: perf-test-validation
+      type: presubmit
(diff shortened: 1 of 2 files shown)


prow/config/jobs/tools.yaml
```diff
@@ -30,6 +30,12 @@ jobs:
     regex: 'docker/.+|cmd/.+'
     requirements: [gcp, docker]
 
+  - name: perf-test-validation
```
howardjohn, line 33: we should only run this on relevant changes. otherwise it will run for every change. You can do this with `regex` like above

carolynhu: yeah, thanks for pointing this out, let me figure out how can we identify something in the istio/tools PR as a relevant change.

PR #2519, opened 2020-03-16
Ensure "get-api-resources" is rerun per distinct cluster

clarketm:
Ensure the `get-api-resources` is rerun (i.e. remade) and separate per distinct cluster. Also, add _velodrome_ to auto-deploy job. 

fix: https://prow.istio.io/view/gcs/istio-prow/logs/deploy-prow_test-infra_postsubmit/40


chases2 (approved):
/hold
for further comment

Issue #2442, opened 2020-02-27
Automatically redirect gcsweb and velodrome from http to https


clarketm:
HTTP to HTTPS redirection using GCP load balancer is currently in alpha ([ref](https://code.google.com/p/google-compute-engine/issues/detail?id=255&thanks=255&ts=1446612833)). The options are whitelist the project, wait for GA, or handle the redirect with [NGINX](https://serverfault.com/a/866912/375458).

PR #2521, opened 2020-03-17
Redirect http -> https for velodrome.istio.io

clarketm:
for: #2442

### DIFF
--- a/prow/cluster/velodrome/velodrome_configmap-nginx.yaml
+++ b/prow/cluster/velodrome/velodrome_configmap-nginx.yaml
@@ -1,0 +1,22 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nginx
+  namespace: velodrome
+  labels:
+    app.kubernetes.io/part-of: velodrome
+data:
+  nginx.conf: |
+    server {
+      listen 80 default_server;
+
+      server_name _;
+
+      location / {
+        if ($http_x_forwarded_proto = 'http') {
+            return 301 https://$host$request_uri;
+        }
+        proxy_pass http://velodrome-external:80/;
+      }
+
+    }
--- a/prow/cluster/velodrome/velodrome_service.yaml
+++ b/prow/cluster/velodrome/velodrome_service.yaml
@@ -7,13 +7,29 @@ metadata:
   namespace: velodrome
 spec:
   type: NodePort
+  selector:
+    app: velodrome
+  ports:
+  - port: 80
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/part-of: velodrome
+  name: velodrome-external
+  namespace: velodrome
+spec:
+  type: NodePort
   ports:
   - port: 80
 ---
 kind: Endpoints
 apiVersion: v1
 metadata:
-  name: velodrome
+  labels:
+    app.kubernetes.io/part-of: velodrome
+  name: velodrome-external
   namespace: velodrome
 subsets:
 - addresses:
--- a/prow/cluster/velodrome/velodrome_deployment.yaml
+++ b/prow/cluster/velodrome/velodrome_deployment.yaml
@@ -1,0 +1,30 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: velodrome
+  namespace: velodrome
+  labels:
+    app.kubernetes.io/part-of: velodrome
+    app: velodrome
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      app: velodrome
+  template:
+    metadata:
+      labels:
+        app: velodrome
+    spec:
+      containers:
+      - name: nginx-proxy
+        image: nginx:1.17
+        ports:
+        - containerPort: 80
+        volumeMounts:
+        - name: nginx
+          mountPath: /etc/nginx/conf.d
+      volumes:
+      - name: nginx
+        configMap:
+          name: nginx
--- a/prow/cluster/velodrome/velodrome_ing.yaml
+++ b/prow/cluster/velodrome/velodrome_ing.yaml
@@ -10,11 +10,6 @@ metadata:
     networking.gke.io/managed-certificates: velodrome-istio-io
     kubernetes.io/ingress.class: "gce"
 spec:
-  rules:
-  - host: velodrome.istio.io
-    http:
-      paths:
-      - path: /*
-        backend:
-          serviceName: velodrome
-          servicePort: 80
+  backend:
+    serviceName: velodrome
+    servicePort: 80


prow/cluster/velodrome/velodrome_configmap-nginx.yaml
```diff
@@ -0,0 +1,22 @@
...
+    server {
+      listen 80 default_server;
+
+      server_name _;
+
+      location / {
+        if ($http_x_forwarded_proto = 'http') {
+            return 301 https://$host$request_uri;
+        }
+        proxy_pass http://130.211.224.127:80/;
```
cjwagner: Shouldn't this be `http://velodrome/` to use the k8s service instead of the k8s endpoint that was deleted?

clarketm: This is the public static IP of the velodrome deployment running in the other cluster: https://github.com/kubernetes/test-infra/blob/f421e19a48dc0578eaa937c7ce34927737caa40e/velodrome/config.yaml#L172. I can leveragea service for this. Let me undo the deletion of the `Endpoint` and see if that will work.

clarketm: Does this now look correct? The ingress forwards all traffic to the _velodrome_ service which is associated with a lightweight nginx proxy deployment to redirect http -> https. Otherwise (if already https), proxy traffic to the  _velodrome-external_ service which has an endpoint pointed to the [static ip](https://github.com/kubernetes/test-infra/blob/f421e19a48dc0578eaa937c7ce34927737caa40e/velodrome/config.yaml#L172) of the actual velodrome instance for Istio running in another cluster.

prow/cluster/velodrome/velodrome_service.yaml
```diff
@@ -5,6 +5,20 @@ metadata:
     app.kubernetes.io/part-of: velodrome
   name: velodrome
   namespace: velodrome
+spec:
+  type: NodePort
```
fejta, line 9: Why are both velodrome and velodrome-external nodeports on port 80?

clarketm: Because local proxy is running on localhost:80 and external veldorome on 130.211.224.127:80

fejta: The proxy is running in a container right? Why do we need to expose the host's port 80?

clarketm: For ingress. What do you propose?

fejta: I thought the whole point here was to ensure all traffic from the internet to velodrome is https? That doesn't work when you're exposing it.

I'd expect ClusterIP

prow/cluster/velodrome/velodrome_service.yaml
```diff
@@ -13,10 +27,12 @@ spec:
 kind: Endpoints
 apiVersion: v1
 metadata:
-  name: velodrome
+  labels:
+    app.kubernetes.io/part-of: velodrome
+  name: velodrome-external
```
fejta, line 32: Wouldn't velodrome-internal make more sense here?

fejta (commented):
What is the expected packet flow here?

I'm expecting something like:
* User -> Internet -> Nginx -> Velodrome

Thus I'd expect the nginx service to be nodeport (people will talk to nginx) and velodrome to be clusterip (internal connection between nginx and velodrome)

I'd expect that we're explicitly try to close the following path:
* User -> Internet -> Velodrome

This currently remains open (by using NodePort). I also don't think it will work because multiple services are trying to claim the same port 80.

prow/cluster/velodrome/velodrome_service.yaml
clarketm: ClusterIP does not work for GKE load balancer.

fejta: You're using nginx, not the GKE load balancer?

fejta: The whole point here is to route all traffic through nginx so we ensure it is https.

Therefore we explicitly want all traffic to the velodrome deployment to be internal, aka ClusterIP

clarketm: We are using a GKE load balancer like we use everywhere else for Ingress not the nginx load balancer. Nginx here is just a reverse proxy in a container.

prow/cluster/velodrome/velodrome_service.yaml
clarketm: No it is an external IP.

fejta: Then what is the point of this PR?

prow/cluster/velodrome/velodrome_service.yaml
fejta: What is the desired packet flow? What is the purpose of nginx and how does it work?

clarketm: Exactly what you said:
User -> Internet -> Nginx -> Velodrome

Velodrome is deployed in an external cluster, so the internal service (nginx) does the http->https redirect then proxies traffic to the external one. Also, AFAIK (and have reconed) GCE firewalls blocks NodePort port access unless explicitly whitelisted on the node.

prow/cluster/velodrome/velodrome_service.yaml
clarketm: I think that is clear from the title? to http redirect.

prow/cluster/velodrome/velodrome_service.yaml
fejta: * You cannot have two service using the same NodePort
* Since we're configuring velodrome to only accept internal requests from nginx, this should be ClusterIP

prow/cluster/velodrome/velodrome_service.yaml
fejta: What is the difference between velodrome and velodrome-external then? Why do both of these exist?

prow/cluster/velodrome/velodrome_service.yaml
clarketm: > You cannot have two service using the same NodePort

The second service maps to an external source defined in the endpoint; there is not a conflict. Also, `port` and `nodePort` fields are different values. see: https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors

> Since we're configuring velodrome to only accept internal requests from nginx, this should be ClusterIP

[GKE load balancers with ingress](https://cloud.google.com/kubernetes-engine/docs/tutorials/http-balancer) **require** `NodePort` services. Same goes for every other service we expose via an ingress  currently deployed (e.g. Deck, Hook, etc.).

prow/cluster/velodrome/velodrome_service.yaml
clarketm: The internal service (i.e. `veldorome`) is an intermediate hop for control over load balancing, cert provisioning, and redirecting before proxying to external service (i.e. `velodrome-external`).

User -> Internet -> Velodrome(internal service) -> Nginx proxy -> Velodrome(external service).

fejta (approved):

PR #2528, opened 2020-03-20
Mount grafana token for releae

howardjohn:


clarketm (approved):

prow/cluster/jobs/all-presets.yaml
```diff
@@ -74,6 +79,9 @@ presets:
   - name: rel-pipeline-github
     secret:
       secretName: rel-pipeline-github
+  - name: rel-pipeline-github
```
brian-avery: should this be rel-pipeline-grafana?

brian-avery (approved):

PR #2530, opened 2020-03-22
Add private 1.5 jobs

howardjohn:


clarketm:
I went ahead and created the configmaps needed for: 
- `preset-override-deps: release-1.5-release`
- `preset-override-deps: release-1.5-istio`

The updates to the script will make this easier (i.e. one command) going forward: https://github.com/istio/test-infra/pull/2531

You need to add these presets to [all-presets.yaml](https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/all-presets.yaml) as well.

howardjohn:
@clarketm thanks I updated and added operator jobs

clarketm (approved):

PR #2531, opened 2020-03-22
Refine script to create-deps-cm for istio-private dependency overrides

clarketm:
To create a _configmap_ in order to override dependencies for `istio@release-1.5` and `release-builder@release-1.5` in `istio-private`  run:
```
$ make -C prow create-deps-configmaps branch=release-1.5
```

Then, add the yaml snippets that the command outputs to stdout to the jobs (or encapsulate in a [preset](https://github.com/istio/test-infra/blob/e05abff488b2245187dc878cb90ab093e7c9b617/prow/cluster/jobs/all-presets.yaml#L161-L176))

```yaml
---
  env:
  - name: DEPENDENCIES
    valueFrom:
      configMapKeyRef:
        name: release-1.5-istio-deps
        key: dependencies
---
  env:
  - name: DEPENDENCIES
    valueFrom:
      configMapKeyRef:
        name: release-1.5-release-deps
        key: dependencies

```

### DIFF
--- a/prow/create-deps-cm.sh
+++ b/prow/create-deps-cm.sh
@@ -15,12 +15,12 @@
 # limitations under the License.
 
 # This script will:
-#   Create a configmap in a K8s cluster for specifying Istio deps
+#   Create a configmap in a K8s cluster for specifying Istio or Release Builder deps
 
 # e.g.
 #  create-deps-cm.sh --context=gke_prow_us-west1-a_prow --branch=master --namespace=test-pods --key=dependencies --interactive master-deps
 
-# TODO: support both older releases and custom dependencies. As written, this only works for release-builder `release-1.5` or newer.
+# NOTE: this only works for istio and release-builder `release-1.5` or newer.
 
 set -o errexit
 set -o nounset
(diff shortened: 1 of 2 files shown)


prow/create-deps-cm.sh
```diff
@@ -77,7 +79,7 @@ check_getopt() {
 get_options() {
   check_getopt
 
-  if opt=$(getopt -o b:c:n:k:i -l branch:,context:,namespace:,key:,interactive -n "$script_name" -- "$@"); then
+  if opt=$(getopt -o b:c:n:k:l:i -l branch:,context:,namespace:,key:,local,interactive,dry-run -n "$script_name" -- "$@"); then
```
fejta, line 82: This would be a lot easier to read (and therefore maintain) if it is written in python or golang

prow/create-deps-cm.sh
```diff
@@ -105,10 +107,18 @@ get_options() {
...
       ;;
+    -l | --local)
+      local=1
+      shift
+      ;;
     -i | --interactive)
       interactive=1
       shift
       ;;
+    --dry-run)
```
fejta, line 118: `--dry-run=true` is a better/safer default than the current `--dry-run=false`.

If so then `--confirm` might be a better name choice since it naturally defaults to the safe mode.

fejta (approved):

prow/create-deps-cm.sh
clarketm: I will rewrite in python.

PR #2532, opened 2020-03-22
Require multicluster tests in presubmit

nmittler:


howardjohn (approved):
I think we need this in postsubmit as well?

PR #2534, opened 2020-03-23
reduce perf benchmark job timeout

carolynhu:
since we are reducing the number of perf config runs, we only run both mode, and there are 6 configs, each config we will run cpu/mem (24min) and latency (24min) which is totally at most 6 hours. here, i reduced it to be 12 h in case we have other configs need to run, like in perf_record and linkerd.

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -35,9 +35,9 @@ periodics:
   name: daily-performance-benchmark
   branches: master
   decorate: true
-  timeout: 24h
+  timeout: 12h
   decoration_config:
-    timeout: 24h0m0s
+    timeout: 12h0m0s
   extra_refs:
     - org: istio
       repo: tools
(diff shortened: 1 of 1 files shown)


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -35,9 +35,9 @@ periodics:
   name: daily-performance-benchmark
   branches: master
   decorate: true
-  timeout: 24h
+  timeout: 12h
```
clarketm, line 38: I think this is a noop. The `timeout` is read from `decoration_config.timeout`.

clarketm (approved):

clarketm:
lgtm. Remove the hold at your convenience.

prow/cluster/jobs/all-periodics.yaml
carolynhu: you mean this is actually never been used? where is the `decoration_config.timeout`?

clarketm: > you mean this is actually never been used? where is the `decoration_config.timeout`?

Yes, you should only need `decoration_config.timeout`, which is the change on [line 40](https://github.com/istio/test-infra/pull/2534/files#diff-fa4a92fd9ad0bf1866a82628a7317117R40).

clarketm: This is more of an FYI; it is not an error to have it defined top-level.

PR #2537, opened 2020-03-24
Use go caching in unit tests

howardjohn:
This has been running for almost a month now with no issues. I am moving
unit tests to use the cache, and switch the experimental job to test an
integration test to see how that performs

@clarketm we probably need some way to exclude this volume mount in private?

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -69,8 +69,19 @@ postsubmits:
             memory: 3Gi
         securityContext:
           privileged: true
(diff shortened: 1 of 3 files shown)


clarketm:
> @clarketm we probably need some way to exclude this volume mount in private?

I will figure out an extensible way to exclude it (e.g. volume name blacklisting); alternatively, we could have a separate, _private_ cache available for the _private_ jobs?

clarketm (approved):
/hold

howardjohn:
/hold cancel

Getting a private cache, especially when on GOB, seems like a bit more trouble than its worth unless we start getting huge gains from the cache (bazel builds from hours to 10s of minutes or something). Right now its just like a 10% improvement

PR #2539, opened 2020-03-24
Update api prow job images

jasonwzm:


howardjohn (changes requested):
We should update all jobs in sync. If you update istio/common-files it will automatically generate this PR

jasonwzm:
> We should update all jobs in sync. If you update istio/common-files it will automatically generate this PR

The automation seems to only update master jobs. For 1.5, i am only updating the job that is affected by tools change.

howardjohn (approved):

clarketm:
> > We should update all jobs in sync. If you update istio/common-files it will automatically generate this PR
> 
> The automation seems to only update master jobs. For 1.5, i am only updating the job that is affected by tools change.

There was a job misconfiguration. It _should_ have been automatic. The combination of the follwing fixes will resolve this for **all** repos ;`release-1.*` branches:
https://github.com/istio/test-infra/pull/2542
https://github.com/istio/test-infra/pull/2543

PR #2545, opened 2020-03-24
Avoid potential conflicts in fork branch name

clarketm:
The current heuristic of using `branch`, and `modifier` per `org/repo` fork as the **branchname** is **not** sufficient to _avoid naming conflict_.  This has be a reoccurring problem with automator.

To eliminate this issue for good, I am including all _identifying_ fields in the **branchname**. Note that `title` is additionally being hashed to comply with git branchname restrictions: https://git-scm.com/docs/git-check-ref-format.

### DIFF
--- a/tools/automator/utils.sh
+++ b/tools/automator/utils.sh
@@ -42,3 +42,8 @@ split_on_commas() {
 evaluate_tmpl() {
   bash -c "eval echo \"$1\""
 }
+
+hash() {
+  local val="$1"
+  echo -n "$val" | md5sum | cut -c1-8
+}
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -198,7 +198,7 @@ create_pr() {
     --title="$title" \
     --match-title="\"$match_title\"" \
     --body="$body" \
-    --source="$user:$src_branch-$modifier" \
+    --source="$user:$fork_name" \
     --confirm
 }
 
@@ -211,7 +211,7 @@ add_labels() {
 commit() {
   git -c "user.name=$user" -c "user.email=$email" commit --message "$title" --author="$user <$email>"
   git show --shortstat
-  git push --force "https://$user:$token@github.com/$user/$repo.git" "HEAD:$branch-$modifier"
+  git push --force "https://$user:$token@github.com/$user/$repo.git" "HEAD:$fork_name"
   pull_request="$(create_pr)"
   add_labels "$pull_request"
 }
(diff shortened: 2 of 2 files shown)


tools/automator/utils.sh
```diff
@@ -42,3 +42,8 @@ split_on_commas() {
 evaluate_tmpl() {
   bash -c "eval echo \"$1\""
 }
+
+hash() {
+  local val="$1" chars="${2:-8}"
```
fejta: $chars is always 8, please leave this as a constant until it is used.

tools/automator/utils.sh
```diff
@@ -42,3 +42,8 @@ split_on_commas() {
...
+  echo -n "$val" | md5sum | cut -c1-"$chars"
```
fejta: no need for echo

```bash
md5sum <<< "$1"
```

tools/automator/utils.sh
clarketm: Yea, this is acting as a function _default_ value. But I get your point that until there is a usage of this function both with and without a need for the default there is not explicit benefit.

tools/automator/utils.sh
clarketm: `<<<` is not POSIX-compliant and also a bashism so I prefer to avoid it where possible.

tools/automator/utils.sh
clarketm: Converted to a constant.

tools/automator/utils.sh
fejta: Check the `#!` this is a `bash` program, not `sh`

fejta (approved):

tools/automator/utils.sh
fejta: No need for this variable, it is only used once. `-c1-8` is easier to read than `-c1-"$chars"`

tools/automator/utils.sh
clarketm: Thanks. I know the shebang is `bash` in this case. It is just a scripting habit for maximal portability. As `<<<` is only available in newer bash version (3+ I believe).

tools/automator/utils.sh
clarketm: This is a utility script (`utils.sh`), so as more jobs use automator, we will get more benefit from a generic, flexible function definition with defaults. But I will hardcode until that need arises. Also, by constant I though you meant literally declaring `chars=8`.

clarketm: Updated

fejta (approved):

PR #2546, opened 2020-03-25
Respect errexit and propogate exit code

clarketm:
In modern shells like `bash` the `errexit` (ie. `set -e`) option does not extend to an arbitrary level of nested functions for command(s) to the left of the`||` or `&&` ([source](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) ; [so](https://stackoverflow.com/questions/19789102/why-is-bash-errexit-not-behaving-as-expected-in-function-calls)).

The downside here is that if the [`commit`](https://github.com/istio/test-infra/blob/7fb215c03d0c33bba022182c2f916f9b656cb721/tools/automator/automator.sh#L211) function fails then the job will _succeed_ when it should _fail_ ([example](https://prow.istio.io/view/gcs/istio-prow/logs/update-build-tools-image_common-files_release-1.5_postsubmit/5)) because the exit code from `create_pr`, for example, is not returned to the caller (`commit`) and hence the exit code is `0` when call stack pops to `work` as if no error ever occurred.

Since these _customized_ error messages I am removing are not informative anyway, I am changing this to a _generic_ error message. It can act as a sentinel in logs to quickly see the job failure point. 

In the future, if we do want custom error messages, there is a simple workaround and it is to define the function body in a subshell and run `set -e` in that subshell (similar to the `work` function); but atm this is needless complexity.

### DIFF
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -235,7 +235,7 @@ work() { (
 
   if ! git diff --cached --quiet --exit-code; then
     fork_name="$src_branch-$branch-$modifier-$(hash "$title")"
-    commit
+    trace commit
   fi
 
   popd
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -213,7 +213,7 @@ commit() {
   git show --shortstat
   git push --force "https://$user:$token@github.com/$user/$repo.git" "HEAD:$fork_name"
   pull_request="$(create_pr)"
-  add_labels "$pull_request"
+  add_labels
 }
 
 work() { (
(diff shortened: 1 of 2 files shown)


tools/automator/automator.sh
```diff
@@ -229,13 +229,13 @@ work() { (
 
   AUTOMATOR_REPO_DIR="$(pwd)"
 
-  bash "${shell_args[@]}" "$script_path" "${script_args[@]}" || print_error "unable to execute command for: $repo"
```
fejta, line 232: Do we not think this is useful debugging information? Updating this function to do:

```bash
print_error() {
  ret=$?
  echo "$@" >&2
  return $ret
}

bash $shell_args $script_path $script_args || print_error "unable to execute command for $repo"
```

Will give you the desired behavior and preserve the debugging.


On the other hand "unable to execute command" is a poorly formed message (tell me the command, please)

I'd favor doing something more like this:
```bash
log-command() {
  set -o xtrace 
  $@
}

log-command bash $shell $script $args
```

clarketm: Thanks, I think the trace would be beneficial here. I will update.

clarketm: I already added support for xtrace in the bash command/script via the `--verbose` flag (#2488).  This is not enabled by default in case the job script logs sensitive info (e.g. API token, secret, password, etc.). I still think this tracing utility function can be useful for other commands.

clarketm: Added a `trace` function and use if for recording the the `commit` logic.

fejta (approved):

PR #2525, opened 2020-03-19
Cleanup gcsweb and velodrome ing object configs: part 2

clarketm:
pre: https://github.com/istio/test-infra/pull/2524

PR #2551, opened 2020-03-25
Support kind blacklist in pruned deployments

clarketm:
`core/v1/Binding` is not a prunable resource, thus adding a kind blacklist to exclude it.

fix: [deploy-prow_test-infra_postsubmit](https://storage.googleapis.com/istio-prow/logs/deploy-prow_test-infra_postsubmit/47/build-log.txt)
blocking: #2525

### DIFF
--- a/prow/Makefile.gcloud.mk
+++ b/prow/Makefile.gcloud.mk
@@ -41,11 +41,8 @@ configure-docker: activate-serviceaccount
 
 get%api-resources:
 	$(eval PRUNE_WL=$(shell KUBECONFIG=$(KUBECONFIG) \
-	python3 "$(current_dir)/api-resources.py" \
-	  --delimiter=" " \
-	  --format="--prune-whitelist='%s'" \
-	  --group-blacklist "authentication.k8s.io" "authorization.k8s.io"\
-	  ))
+	python3 "$(current_dir)/api-resources.py" --delimiter=" " --format="--prune-whitelist='%s'" \
+	))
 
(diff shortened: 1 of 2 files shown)


prow/Makefile.gcloud.mk
```diff
@@ -41,11 +41,8 @@ configure-docker: activate-serviceaccount
 
 get%api-resources:
 	$(eval PRUNE_WL=$(shell KUBECONFIG=$(KUBECONFIG) \
```
fejta, line 43: There is WAY too much eval everywhere, which violates our style guidelines for bash.

http://google.github.io/styleguide/shellguide.html#s6.6-eval

We should also try to putting logic in makefiles as these are not portable across repos

Why is this necessary here?

clarketm: This is to _set_ to global `PRUNE_WL` `make` variable used in the other `make` deploy rules in `prow/Makefile`, so a single target can handle all of these. I understand this is difficult to follow. I can work on a revision but I would like to get this specific change to `api-resources.py` in to fix the deploy job.

clarketm: I agree. I think a better approach is to pull this out of the Makefile like you suggest, into a script or program.

clarketm: I created this: https://github.com/istio/test-infra/issues/2552.  Will handle as a follow-up.

fejta (approved):

PR #2563, opened 2020-04-02
Metrics

### LINKED ISSUE #1774: Istio's Prow should use the new upstream monitoring and alerting stack.
https://github.com/kubernetes/test-infra/tree/master/prow/cluster/monitoring#monitoring
We'll get lots of nice preconfigured graphs and alerts for free. e.g. https://monitoring.prow.k8s.io
I think we can pretty easily move over anything from https://velodrome.istio.io that is still used too so that we can turn down velodrome and just have a single monitoring dashboard.

cc @clarketm @fejta @geeknoid @howardjohn

scottilee:
Add files required for a custom monitoring and alerting stack planned for monitoring.prow.istio.io.

- Sends alerts to the Istio Slack channel #test-alerts

Fixes #1774.


prow/cluster/monitoring/BUILD.bazel
```diff
@@ -0,0 +1,100 @@
+load("//prow:def.bzl", "BUILD_CLUSTER", "MULTI_KIND", "component", "release")
```
clarketm: Istio is not setup for `bazel` deployments. We can safely remove this (and the other) `BUILD.bazel` files from this PR. As an alternative, we can use `kubectl` (w/ `helm` for magic strings) but that can be in a follow-up.

prow/cluster/monitoring/OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
```
clarketm: You can omit this `OWNERS` file as ownership is already handled by [`CODEOWNERS`](https://github.com/istio/test-infra/blob/master/CODEOWNERS).

clarketm (changes requested):

prow/cluster/monitoring/OWNERS
scottilee: Done.

prow/cluster/monitoring/BUILD.bazel
scottilee: Done.

clarketm:
@scottilee – regarding the lint errors, can you omit the `vendor` directory and update your PR? 

Managing these vendor dependencies and auto-deploying can be a follow-up item.

clarketm (approved):
This is awesome, thanks for the help! I will create follow-up tasks for updating the README, installing vendor deps, and setting up auto-deployment.

PR #2568, opened 2020-04-04
Reserve rerun perms for periodics

clarketm:
Periodics do not have an org, so using `"*"` allows this rerun config to apply to both **istio** and all the periodics.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -75,7 +75,7 @@ deck:
 
   tide_update_period: 1s
   rerun_auth_configs:
-    istio:
+    '*':
       github_users:
       - cjwagner
       - fejta


prow/config.yaml
```diff
@@ -75,7 +75,7 @@ deck:
 
   tide_update_period: 1s
   rerun_auth_configs:
-    istio:
+    '*':
```
cjwagner, line 78: Does this allow the istio/maintainers team to rerun istio-private jobs or is only the most specific rerun_auth_config respected?

cjwagner (approved):
I have one concern, but if it isn't substantive then this LGTM.
/hold

prow/config.yaml
clarketm: It is only the *the most specific rerun_auth_config respected*.

PR #2572, opened 2020-04-07
add presubmit for operator controller test

richardwxn:
pending on https://github.com/istio/istio/pull/22641, specified as optional now

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -108,6 +108,19 @@ jobs:
       - name: TEST_SELECT
         value: "-postsubmit,-flaky,-multicluster"
 
+  - name: integ-operator-controller-tests
+    type: presubmit
+    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.operator.kube.presubmit]
+    requirements: [kind]
+    modifiers: [optional]
+    env:
+      - name: HUB
+        value: "gcr.io/istio-testing"
+      - name: TAG
+        value: "1.6-dev"
+      - name: TEST_SELECT
+        value: "-postsubmit,-flaky,-multicluster"
+
   - name: integ-local-tests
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -108,6 +108,19 @@ jobs:
       - name: TEST_SELECT
         value: "-postsubmit,-flaky,-multicluster"
 
+  - name: integ-operator-controller-tests
+    type: presubmit
+    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.operator.kube.presubmit]
+    requirements: [kind]
+    modifiers: [optional]
```
howardjohn: please make it `[optional,hidden]` until the test has stabilized

prow/config/jobs/istio.yaml
```diff
@@ -108,6 +108,19 @@ jobs:
...
+    env:
+      - name: HUB
+        value: "gcr.io/istio-testing"
+      - name: TAG
+        value: "1.6-dev"
```
howardjohn: don't set this, otherwise we are not testing the PR and instead are testing the already merged code

howardjohn (changes requested):

prow/config/jobs/istio.yaml
richardwxn: oh yes, but how can I specify hub and tag for this case? should I just leave it empty the test fw would infer?

howardjohn: the test will set hub and tag to `istio-testing` (not a real hub) and build images locally. Then the jobs will use those local images

richardwxn: got it

prow/config/jobs/istio.yaml
richardwxn: done

dgn (approved):

howardjohn (approved):

PR #2585, opened 2020-04-14
Split current bump PR into 2: one for jobs and one for prow object files.

### LINKED ISSUE #2550: Split current bump PR into 2: one for jobs and one for prow object files.
Separation of concerns so Prow oncall can bump Prow-specific components without the approval of Istio T&R and vice versa wrt jobs.

PR in question: https://github.com/istio/test-infra/pull/2538

If all the fields are customizable in `autobump` (e.g. title, dirs, etc.), then we can probably just use the script twice with different parameters, otherwise, Automator can be used for the job-specific and miscellaneous other image bumps.

clarketm:
fix https://github.com/istio/test-infra/issues/2550


prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
```diff
@@ -95,7 +95,7 @@ periodics:
       - name: PLANK_DEPLOYMENT_FILE
         value: prow/cluster/plank_deployment.yaml
       - name: COMPONENT_FILE_DIR
-        value: prow/cluster,prow/config/jobs,.
+        value: prow/cluster/build,prow/cluster/private,prow/cluster/gcsweb,prow/cluster/velodrome,prow/cluster/monitoring
```
clarketm, line 98: This needs rework or more support in autobump.

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
```diff
@@ -95,11 +95,11 @@ periodics:
       - name: PLANK_DEPLOYMENT_FILE
         value: prow/cluster/plank_deployment.yaml
       - name: COMPONENT_FILE_DIR
-        value: prow/cluster,prow/config/jobs,.
+        value: prow/cluster,prow/cluster/jobs/istio/test-infra
```
cjwagner: We need to bump this image with the rest of the prow components:
https://github.com/istio/test-infra/blob/89fa0e1fae5964c33780c59583b785a05c2e0d91/.prow.yaml#L8

prow/config/jobs/test-infra.yaml
```diff
@@ -136,3 +136,24 @@ jobs:
...
+    - ./tools/automator/automator.sh
+    - --org=istio
+    - --repo=test-infra
+    - "--title=Automator: bump k8s-prow images"
+    - --modifier=bump-k8s-prow-images
+    - --token-path=/etc/github-token/oauth
+    - --script-path=../test-infra/tools/automator/scripts/update-images.sh
+    - --verbose
+    - --
+    - --paths=prow/cluster/jobs/**/*.yaml,prow/config/jobs/**/*.yaml,.prow.yaml
```
cjwagner: Doesn't this need to exclude `prow/cluster/jobs/istio/test-infra`?

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
clarketm: I need to either change ownership of the file or move this job back to `.../jobs/istio/test-infra/` directory. I will do the former.

prow/config/jobs/test-infra.yaml
clarketm: Technically no, since the image version of plank (i.e. `--source`) should always be the same as the image versions of jobs in `test-infra/` (since they are bumped together by autobump).

It doesn't hurt to exclude it, but I will need to enable `extglob` first to ensure the syntax remains simple.

prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml
clarketm: Updated code and modified CODEOWNERS.

prow/config/jobs/test-infra.yaml
cjwagner: Ah so that should always be a noop. 👍 
Excluding might be preferable, but it doesn't seem important and shouldn't hold this up.

cjwagner (approved):

prow/config/jobs/test-infra.yaml
clarketm: I went ahead and added an exclusion on `test-infra` to tidy this up (and ensure this doesn't regress  in case the logic changes in the future)

cjwagner (approved):
LGTM, but I don't think I can approve fully.

howardjohn (approved):

PR #2591, opened 2020-04-16
Boost resources for envoy jobs

clarketm:
While the need is storage, I am also bumping cpu and memory proportionately since nodes are now basically dedicated (and for buffer).

image


howardjohn (approved):
how big are the machines? 64 core?

clarketm:
> how big are the machines? 64 core?

64 vcpu, 1TB storage, and 256GB ram. Also, the actual allocatable values are slightly less and what is set in the resources limits (roughly).

PR #2596, opened 2020-04-16
Increase storage limit again (w/ 2TB SSD disks)

clarketm:
Node pool already upgrade to support this change.

### DIFF
--- a/prow/cluster/jobs/istio/envoy/istio.envoy.master.gen.yaml
+++ b/prow/cluster/jobs/istio/envoy/istio.envoy.master.gen.yaml
@@ -25,11 +25,11 @@ presubmits:
         resources:
           limits:
             cpu: "64"
-            ephemeral-storage: 860G
+            ephemeral-storage: 1790G
             memory: 240G
           requests:
             cpu: "48"
-            ephemeral-storage: 650G
+            ephemeral-storage: 1500G
             memory: 180G
         securityContext:
           privileged: true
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

fejta:
What cleans up these files? Why do we expect this to work any differently than the last couple times we tried this patch?

clarketm:
> What cleans up these files? 

This is ephemeral storage, so my assumption is this is cleanup up by GKE once pod becomes nonexistent.

> Why do we expect this to work any differently than the last couple times we tried this patch?

The other patches were increasing to the storage capacity within the limits of the current node pool (~860GB allocatable). Since I have no way of knowing how much storage a particular job needs I was basically just moving the needle and monitoring the result. Unfortunately, 1TB disk was insufficient; after speaking with @PiotrSikora on [Slack](https://istio.slack.com/archives/C6RHFVB44/p1586993821005800) it was decided to play it safe and go with 2TB disks which brings us to this (hopefully final) change. 

Also, it is worth mentioning that if we overshoot (i.e. define more storage than needed) and the job succeeds, then we can see exactly what the storage upper bound is via metrics.

fejta:
It is not clear to me why 2TB is safe while 1TB is insufficient?

clarketm:
> It is not clear to me why 2TB is safe while 1TB is insufficient?

I do not know why this particular job(s) requires so much storage; I was informed that it was dictated by upstream envoy (i.e. where the tests and source-code originate). Please reach out the @PiotrSikora or another maintainer of envoy for a more detailed rationale (there is also this [thread](https://istio.slack.com/archives/C6RHFVB44/p1586993821005800)).

PR #2599, opened 2020-04-19
Some fixes and improvements to recreate_prow_configmaps.py

chizhg:
1. Also use `replace` to update the prow job config configmap, this can avoid downtime.
2. Add a new `--silent` flag to allow users skip the confirmation step in updating configmaps.

/cc @clarketm

### DIFF
--- a/prow/recreate_prow_configmaps.py
+++ b/prow/recreate_prow_configmaps.py
@@ -14,14 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script deletes and recreates the prow configmaps
-# USE AT YOUR OWN RISK! This is a break-glass tool.
-# See September 25th, 2018 in docs/post-mortems.md
-
 #
 # USAGE: have KUBECONFIG pointed at your prow cluster then from test-infra root:
 #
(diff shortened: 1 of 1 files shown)


prow/recreate_prow_configmaps.py
```diff
@@ -122,8 +117,8 @@ def main():
         "\n!!     WARNING THIS WILL RECREATE **ALL** PROW CONFIGMAPS.     !!"
         "\n!!    ARE YOU SURE YOU WANT TO DO THIS? IF SO, ENTER 'YES'.    !! "
     ) + '\n' + '!' * 65 + '\n\n: '
-    if args.wet:
-        if input(prompt) != "YES":
+    if args.wet and not args.silent:
+        if raw_input(prompt) != "YES":
```
clarketm: Python3 syntax is preferred.

```suggestion
        if input(prompt) != "YES":
```

clarketm (changes requested):

prow/recreate_prow_configmaps.py
chizhg: Done.

clarketm (approved):

PR #2603, opened 2020-04-21
Current image kubectl client does not support --dry-run=client

clarketm:
`--dry-run=client` not available in kubectl@v1.15

fix https://prow.istio.io/view/gcs/istio-prow/logs/deploy-prow_test-infra_postsubmit/53

### DIFF
--- a/prow/cluster/monitoring/mixins/Makefile
+++ b/prow/cluster/monitoring/mixins/Makefile
@@ -34,7 +34,7 @@ apply-configmaps: grafana-dashboards
 	@for input in $(shell ls dashboards_out); do \
 		dashboard_name="grafana-dashboard-$${input%.*}"; \
 		echo "Generating dashboard $${dashboard_name} from $${input} ..."; \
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
We can also update kubectl

clarketm:
> We can also update kubectl

I am hesitant to do this because v1.15 is the current [_regular_](https://cloud.google.com/kubernetes-engine/docs/release-notes-regular) channel release and what we use in CI. Although it should be backwards compatible it makes sense for client/server to match.

howardjohn:
Good point, I was thinking of our istio/istio CI which is on 1.18 but we don't use the kubectl in the image anyways

fejta (approved):
/hold

clarketm:
> Good point, I was thinking of our istio/istio CI which is on 1.18 but we don't use the kubectl in the image anyways

I will bump the `kubectl` version in the image in a follow-up. I still prefer `--dry-run` here over `--dry-run=client` since it works for both _new_ and _old_ versions.

Issue #1626, opened 2019-08-08
Create a periodic job to update istio.io reference docs


geeknoid:
We need a nightly job that updates istio.io's reference docs. To do this, the job needs to check out the istio/istio.io repo and run the scripts/grab_reference_docs.sh script. If after running the script any files are modified in the repo, create a PR for the changes.

This only needs to work on the master branch.

scottilee:
@geeknoid I wanted to take a shot at this. Does this look like a good start?

```
- interval: 24h
  name: istioio-update-periodic
  branches: master
  decorate: true
  annotations:
    testgrid-dashboards: istio_istio.io
    testgrid-alert-email: istio-oncall@googlegroups.com
    testgrid-num-failures-to-alert: '1'
  labels:
    preset-release-pipeline: "true"
  branches: master
  labels:
    preset-service-account: "true"
  spec:
    containers:
    - image: gcr.io/istio-testing/prowbazel:0.4.13
      args:
      - "--repo=github.com/istio/istio.io"
      - "--clean"
      - "--timeout=120"
      command:
        - ./scripts/grab_reference_docs.sh
      # Bazel needs privileged mode in order to sandbox builds.
      securityContext:
        privileged: true
    nodeSelector:
      testing: build-pool
```

Also, I wasn't sure about the whole looking for modified files and creating a PR part. Any examples of this somewhere?

geeknoid:
Since I filed the issue, things have evolved. Instead, you run "make update_ref_docs". And this needs to be run using the build-tools container (gcr.io/istio-testing/build-tools:2019-09-20T15-04-58).

As for the "looking for modified files" thing, the closest thing would be the linting job in the istio/api repo. It's running some code gen and making sure no files have changed.

clarketm:
@geeknoid 
I will take it off your hands.

clarketm:
> Also, I wasn't sure about the whole looking for modified files and creating a PR part. Any examples of this somewhere?

I recently added a lightweight script named [automator](https://github.com/istio/test-infra/blob/88763e6d7f88e8ff1f7fbef17ed430d2ccbb7808/tools/automator/automator.sh) to handle the diff checking and pr creation, so this should now be trivial to implement.

PR #2613, opened 2020-04-24
Set up prow for release 1.6

brian-avery:


prow/config/jobs/api-1.6.yaml
```diff
@@ -0,0 +1,52 @@
+branches:
+- release-1.6
+image: gcr.io/istio-testing/build-tools:master-latest
```
howardjohn: this will change immediately after right? since we should use a pinned version normally

howardjohn (approved):
lgtm!

clarketm:
> /test pull-test-infra-check-testgrid-config

@brian-avery to fix this, add the new dashboard names (e.g. `istio_release-1.6_istio.io_periodic`) to https://github.com/istio/test-infra/blob/master/testgrid/default.yaml. You can use the 1.5 ones as a reference (they are most likely the same if the jobs are the same).

prow/config/jobs/istio-1.6.yaml
```diff
@@ -0,0 +1,327 @@
...
+- command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=istio.io
+  - --token-path=/etc/github-token/oauth
+  - --cmd=make update_ref_docs
+  - --dry-run
+  modifiers:
+  - optional
+  name: update-ref-docs-dry-run
```
clarketm: We only build ref docs from `master`, so I think this job can be omitted (which will also resolve the test failure w/ testgrid).

clarketm (approved):

prow/config/jobs/istio.io-1.6.yaml
```diff
@@ -0,0 +1,63 @@
...
+- command:
+  - entrypoint
+  - prow/integ-suite-kind.sh
+  - test.kube.postsubmit
+  name: k8s-tests
+  requirements:
+  - kind
+  - gcp
+  type: postsubmit
+- command:
```
clarketm: Remove this job as well.  It is only needed on `master`. ref:  https://github.com/istio/test-infra/issues/1626

PR #2618, opened 2020-04-27
Update to pull references docs from release-1.6

ericvn:
For the time when we are using the release-1.6 branches for the upcoming release, we want to pull the reference doc updates from the new branch. We will switch back to master when the istio.io branch is made to `master`

@frankbu Does this make sense?


howardjohn (approved):

brian-avery (approved):

frankbu (approved):
Sounds good. Should also add a step to switch back in the release instructions: https://github.com/istio/istio.io#on-the-day-of-the-release

PR #2619, opened 2020-04-27
automator: assign AUTOMATOR_BRANCH=$branch

clarketm:
`AUTOMATOR_BRANCH` should equal `$branch` which is the _destination_ branch (i.e. what is cloned in `automator` and PRed against).

To be comprehensive, I am adding several _source_ env variables (which are set to existing [Prow variables](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables)):
* `AUTOMATOR_SRC_ORG` –  _source_ organization.
* `AUTOMATOR_SRC_REPO` –  _source_ repository.
* `AUTOMATOR_SRC_BRANCH` –  _source_ branch.

and removing futile options (`--sha` and `--src_branch`) which do not have a purpose.

### DIFF
--- a/prow/cluster/jobs/istio/api/istio.api.master.gen.yaml
+++ b/prow/cluster/jobs/istio/api/istio.api.master.gen.yaml
@@ -75,7 +75,8 @@ postsubmits:
         - ../test-infra/tools/automator/automator.sh
         - --org=istio
         - --repo=istio,client-go
-        - '--title=Automator: update istio/api@$AUTOMATOR_BRANCH dependency in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH'
+        - '--title=Automator: update istio/api@$AUTOMATOR_SRC_BRANCH dependency in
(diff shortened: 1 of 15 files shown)


ericvn:
Only issue with this is that _istio.io_ stays on _master_ and should maybe be pulling docs from the _release-1.6_ branches of the other repos (still verifying this) while we are in the feature freeze. Once we release 1.6, and the _istio.io release-1.6_ is created then the ref docs would revert to the _master_ branch inn the other repos. #2618 was a change for the later.

clarketm:
> Only issue with this is that _istio.io_ stays on _master_ and should maybe be pulling docs from the _release-1.6_ branches of the other repos (still verifying this) while we are in the feature freeze. Once we release 1.6, and the _istio.io release-1.6_ is created then the ref docs would revert to the _master_ branch inn the other repos. #2618 was a change for the later.

👍 I merged in your changes to the `istio.io` job, so that is now unchanged.

brian-avery:
This looks good to me. I'll let Eric provide the approval.

ericvn (approved):

PR #2617, opened 2020-04-25
100% automated release branch job generation

clarketm:
The goal of this PR is to make the `$ go run generate.go branch 1.x` command 100% comprehensive without requiring additional _manual_ modifications to the generated output. The two changes included in this PR are:

1. If `disable_release_branching: true` exists in the config for a job, then it will not be copied/cloned on release branching.

2. When _release_ jobs are created, the image tag should be updated (e.g. `master` -> `release-1.6`) and correctly tagged in **gcr**.

fix https://github.com/istio/test-infra/issues/2614

PR #2628, opened 2020-05-01
Update build-image and automator jobs for release-1.6

clarketm:
release-1.6 were incorrectly pointing at `master-` images. 

> Note: the release-branching feature should probably use another release branch as a base instead of `master` (it should work out of the box. I will configure this in a follow-up). There should also be logic to change `image`.

### DIFF
--- a/prow/cluster/jobs/istio/api/istio.api.release-1.6.gen.yaml
+++ b/prow/cluster/jobs/istio/api/istio.api.release-1.6.gen.yaml
@@ -15,7 +15,7 @@ postsubmits:
       - command:
         - make
         - presubmit
-        image: gcr.io/istio-testing/build-tools:master-2020-04-28T21-48-12
+        image: gcr.io/istio-testing/build-tools:release-1.6-2020-05-01T18-02-13
         name: ""
         resources:
           limits:
@@ -42,7 +42,7 @@ postsubmits:
       - command:
         - make
         - gen-check
-        image: gcr.io/istio-testing/build-tools:master-2020-04-28T21-48-12
+        image: gcr.io/istio-testing/build-tools:release-1.6-2020-05-01T18-02-13
         name: ""
         resources:
           limits:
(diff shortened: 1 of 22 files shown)


howardjohn:
> Note: the release-branching feature should probably use another release branch as a base instead of master (it should work out of the box. I will configure this in a follow-up). There should also be logic to change image.

Why? when we cut the branch 1.6 == master. If we use 1.5, all of the jobs would be for 1.5 which is no longer accurate (jobs added/removed/changed since then)

clarketm:
> > Note: the release-branching feature should probably use another release branch as a base instead of master (it should work out of the box. I will configure this in a follow-up). There should also be logic to change image.
> 
> Why? when we cut the branch 1.6 == master. If we use 1.5, all of the jobs would be for 1.5 which is no longer accurate (jobs added/removed/changed since then)

You are correct. I take that back. We simply need to:
1. explicitly set `istio/test-infra@master` in repos for `common-files.yaml` config
2. Have logic to change image from `master` -> `release-1.x`

and it will be error-free.

> I included **1.** here and will add **2.** in the follow-up.

howardjohn:
@clarketm another issue is when you create the 1.6 jobs, there is no 1.6 build-tools image because you need the 1.6 jobs set up to build it. Its a chicken and egg problem. I think we just needed (1) as part of the logic, and this PR should have been sent immediately instead of a while afterwards?

howardjohn (approved):

clarketm:
> I think we just needed (1) as part of the logic, and this PR should have been sent immediately instead of a while afterwards?

I think this problem can be avoid (and automated) by simply tagging an initial version of the image `gcr.io/istio-testing/build-tools:release-1.6-XXXXX` from `gcr.io/istio-testing/build-tools:master-latest` when the jobs are first created.

howardjohn:
That is a good idea

clarketm:
> That is a good idea

I will send something out for this in a few.

clarketm:
>> I think this problem can be avoid (and automated) by simply tagging an initial version of the image `gcr.io/istio-testing/build-tools:release-1.6-XXXXX` from `gcr.io/istio-testing/build-tools:master-latest` when the jobs are first created.

> That is a good idea

I added this to the release branching improvements PR I have open (and described it in the linked issue): #2617

PR #2644, opened 2020-05-06
Add extra perf cluster and nighthawk daily job

carolynhu:
https://github.com/istio/istio/issues/21161

### DIFF
--- a/boskos/resources.yaml
+++ b/boskos/resources.yaml
@@ -46,6 +46,7 @@ resources:
   - istio-boskos-perf-01
   - istio-boskos-perf-02
   - istio-boskos-perf-05
+  - istio-boskos-perf-06
 - type: gcp-project
   state: dirty
   names:
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -1,5 +1,3 @@
-
-
 istio_container_with_kind: &istio_container_with_kind
   image: gcr.io/istio-testing/build-tools:master-2020-05-04T23-19-43
   securityContext:
@@ -32,7 +30,7 @@ periodics:
       testing: test-pool
 
 - cron: "0 0 * * *" # starts every day at 00:00AM UTC
-  name: daily-performance-benchmark
+  name: daily-fortio-performance-benchmark
   branches: master
   decorate: true
   timeout: 24h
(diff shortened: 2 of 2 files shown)


boskos/resources.yaml
```diff
@@ -46,6 +46,7 @@ resources:
   - istio-boskos-perf-01
   - istio-boskos-perf-02
   - istio-boskos-perf-05
+  - istio-boskos-perf-06
```
clarketm, line 49: If we are adding two jobs, it may be better to add _two_ additional perf projects. There are times where we are at 100% utilization (ref: https://velodrome.istio.io/dashboard/db/boskos-dashboard?orgId=1&from=now-30d&to=now). Also, since the runtime is capped at 24hrs there is a high chance they will overlap:

```suggestion
  - istio-boskos-perf-06
  - istio-boskos-perf-07
```

clarketm (changes requested):

boskos/resources.yaml
carolynhu: what's the 01, 02, 05 used for? Since previously we only have two perf jobs. Why do we have 3 perf clusters?

carolynhu: I guess, we removed one perf job, but did not remove the extra perf cluster? That's why I'm only adding one extra cluster. Or actually one of 01, 02 or 05 is not available or used?

clarketm: Okay, I see. If there are only 4 jobs total using clusters, then 4 projects is sufficient.

clarketm (approved):

clarketm:
/hold
For creation and setup of the `istio-boskos-perf-06` boskos project.

boskos/resources.yaml
carolynhu: yes, i think so.

clarketm: Looking at the [code](https://github.com/istio/tools/blob/b7b1b8a3236ea95e733decb28df61cf4b1e95888/perf/benchmark/run_benchmark_job.sh#L37), anything running `perf/benchmark/run_benchmark_job.sh` and **not** overriding `RESOURCE_TYPE` is leasing a cluster from this pool. So unless I missed something, you are correct in that it is only **4**.

clarketm:
`istio-boskos-perf-06` has been created and configured. Unhold at your convenience.

PR #2650, opened 2020-05-08
Add prow jobs for upgrade-downgrade tests, upgrade tests, and downgrade tests

lei-tang:
- Add prow job for upgrade-downgrade test from 1.5 latest to 1.6 latest.
- Add prow job for upgrade test from 1.5 latest to 1.6 latest.
- Add prow job for downgrade test from 1.6 latest to 1.5 latest.
- Add prow job for upgrade-downgrade test from 1.6 latest to master.
- Add prow job for upgrade test from 1.6 latest to master.
- Add prow job for downgrade test from master to 1.6 latest.

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -157,9 +157,111 @@ periodics:
     nodeSelector:
       testing: test-pool
 
-# upgrade using istioctl from 1.5 latest to 1.6 latest
+# upgrade-downgrade test using istioctl from 1.5 latest to 1.6 latest
+- cron: "0 4 * * *"  # starts every day at 04:00AM UTC
+  name: istio-upgrade-downgrade-using-istioctl-1.5_latest-1.6_latest
+  branches: master
+  decorate: true
+  extra_refs:
+  - org: istio
+    repo: tools
+    base_ref: master
+    path_alias: istio.io/tools
+  annotations:
+    testgrid-dashboards: istio_release-pipeline
+    testgrid-alert-email: istio-oncall@googlegroups.com
+    testgrid-num-failures-to-alert: '1'
+  labels:
+    preset-service-account: "true"
+  spec:
+    containers:
+    - <<: *istio_container_with_kind
+      env:
+      - name: SOURCE_TAG
+        value: 1.5_latest
+      - name: TARGET_TAG
+        value: 1.6_latest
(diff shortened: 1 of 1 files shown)


JimmyCYJ (approved):

howardjohn (commented):
why do we need upgrade and downgrade test if we have upgrade+downgrade test? the single ones seem redundant?

lei-tang:
> why do we need upgrade and downgrade test if we have upgrade+downgrade test? the single ones seem redundant?

The upgrade-downgrade test first upgrade then downgrade, which is one of the scenarios. The other two scenarios are upgrade-only and downgrade-only. We want to cover all three scenarios. For example, a user may want to obtain test results (e.g., packet delivery ratio) when only a upgrade or downgrade is conducted.

howardjohn:
Why do we want to test all 3? we should either have 2 scenarios (upgrade, downgrade) or a single combined scenario (upgrade the downgrade).What value do we get from having both? To me it seems redundant

lei-tang:
> Why do we want to test all 3? we should either have 2 scenarios (upgrade, downgrade) or a single combined scenario (upgrade the downgrade).What value do we get from having both? To me it seems redundant

One of the values of testing all three scenarios (i.e., upgrade-then-downgrade, upgrade-only, downgrade-only) is that a user may try any one of these three scenarios and want to know the results through the corresponding test.

howardjohn:
Who is the user? An Istio developer or a normal istio user? I have never seen a normal user looking at these tests

howardjohn:
I have also not seen any case where we duplicate our tests

lei-tang:
Per review comments, the tests for the upgrade-downgrade scenario are removed.

clarketm (approved):
I think we will also need to increase the resource count in boskos config: 
https://github.com/istio/test-infra/blob/9e54f15c64a9a4e8a9a96d162547107677b50bf6/boskos/resources.yaml#L5-L6 

but I will do that in a follow-up.

clarketm:
> Per review comments, the tests for the upgrade-downgrade scenario are removed.

@lei-tang it looks like you forgot to remove the `upgrade-downgrade` tests. If you still plan to remove these as redundant, please do so in a follow-up PR.

lei-tang:
@clarketm Thanks! https://github.com/istio/test-infra/pull/2656 is created to remove the tests for the upgrade-downgrade scenario.

PR #2651, opened 2020-05-08
Increase boskos resources

clarketm:
Increase `gke-e2e-test` to match jobs using this pool and increase `gke-perf-preset` to number of projects.


carolynhu (approved):
LGTM

PR #2669, opened 2020-05-19
Explicitly specify --github-token-path.

cjwagner:
Explicitly specify the token path as the default value since the default will be removed in the near future.
```
{
  component: "hook"   
  file: "prow/flagutil/github.go:71"   
  func: "k8s.io/test-infra/prow/flagutil.(*GitHubOptions).Validate"   
  level: "warning"   
  msg: "missing required flag: please set to --github-token-path=/etc/github/oauth before June 2020"   
}
```
/assign @clarketm


fejta (approved):
🎉

PR #2680, opened 2020-05-28
recreated_prow_configmaps: Providing the options of only updating some of the prow configs

chaodaiG:
This change provides options to `recreated_prow_configmaps` for it to allow only update part of configmaps instead of everything. The default behavior doesn't change

/assign @clarketm 
/cc @chizhg

### DIFF
--- a/prow/recreate_prow_configmaps.py
+++ b/prow/recreate_prow_configmaps.py
@@ -90,18 +90,21 @@ def main():
     parser.add_argument(
         "--job-config-dir", default="config/jobs",
         help="root dir of prow jobs configmap")
+    parser.add_argument("--skip-job-configmap", action="store_true")
     # prow config
     parser.add_argument("--prow-configmap", default="config",
                         help="name of prow primary configmap")
     parser.add_argument(
         "--prow-config-path", default="config.yaml",
         help="path to the primary prow config")
+    parser.add_argument("--skip-prow-configmap", action="store_true")
     # plugins config
     parser.add_argument("--plugins-configmap", default="plugins",
                         help="name of prow plugins configmap")
     parser.add_argument(
         "--plugins-config-path", default="plugins.yaml",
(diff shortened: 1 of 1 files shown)


fejta:
We ought to be using https://github.com/kubernetes/test-infra/blob/master/def/configmap.bzl#L67 for this

fejta:
https://github.com/kubernetes/test-infra/blob/100609e548a3cca9f007557727ed83dee0992b14/config/prow/cluster/monitoring/mixins/dashboards_out/BUILD.bazel#L8-L16

chaodaiG:
> We ought to be using https://github.com/kubernetes/test-infra/blob/master/def/configmap.bzl#L67 for this

Sorry for not making the purpose of this PR clear, this is done because knative is using this script: https://github.com/knative/test-infra/blob/5ba38edc41ae0dd0c9a3eb7bd6b8577ed8b60fb0/config/prow_common.mk#L74

I can see that istio is using bazel already, so using `configmap.bzl` could be a good fit for istio. Since knative isn't using bazel at the moment, I might just steal this script to knative if istio decided to switch gear.

fejta:
knative should also use bazel for its deployments when/if it is part of our oncall rotation

chaodaiG:
For the part that will be migrated to oss-test-infra, i.e. core prow config and plugins, I agree, and they are not the target of this change.

The target of this change is jobs configs, since the jobs config will still live inside knative and managed by knative I'm not sure they'll onboard with bazel

fejta:
They do not need to be on board with bazel as how we deploy our configmaps is not their concern. 

The point is we already have a way to consistently convert a file or set of files into a configmap and deploy it -- the `k8s_configmap` rule.

There is no need for this python script and unknown stability.

chaodaiG:
Agreed, will think about a better approach, closing this PR for now

/close

PR #2703, opened 2020-06-09
add integ-multicluster-pilot-k8s-tests

stevenctl:
To incrementally move Pilot tests such as https://github.com/istio/istio/pull/24549 as part of https://github.com/istio/istio/issues/23618, we need a separate prow job that uses the multicluster topology. 

Suites that haven't been converted (had `RequireSingleCluster` removed) will be skipped. If this job is added we will be able to properly test https://github.com/istio/istio/pull/24549 under both single and multicluster scenarios.

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -42,6 +42,19 @@ jobs:
       - name: TEST_SELECT
         value: "-postsubmit,-flaky,-multicluster"
 
+  - name: integ-multicluster-pilot-k8s-tests
+      type: presubmit
+      command:
+        - entrypoint
+        - prow/integ-suite-kind.sh
+        - --topology
+        - MULTICLUSTER
+        - test.integration.pilot.kube.presubmit
+      requirements: [kind]
+      env:
+        - name: TEST_SELECT
+          value: "-postsubmit,-flaky,-multicluster"
+
   - name: integ-security-k8s-tests
     type: presubmit
     command: [entrypoint, prow/integ-suite-kind.sh, test.integration.security.kube.presubmit]
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -42,19 +42,6 @@ jobs:
       - name: TEST_SELECT
         value: "-postsubmit,-flaky,-multicluster"
 
-  - name: integ-multicluster-pilot-k8s-tests
-    type: presubmit
-    command:
-      - entrypoint
-      - prow/integ-suite-kind.sh
-      - --topology
-      - MULTICLUSTER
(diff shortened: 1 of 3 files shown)


howardjohn:
In the end will we have 2x the number of tests currently? As in single cluster tests and multicluster tests for pilot/security/telemetry/foo?

stevenctl:
> In the end will we have 2x the number of tests currently?

Whether we need to run in both single and multicluster mode long term still needs discussion. 

For the time being this job would let me incrementally convert tests without modifying the original integ-pilot-k8s-tests job to enable multicluster there. The alternative is something like https://github.com/istio/istio/pull/24162 where I'd make every test runnable multicluster enabled but not necessarily use all clusters, then adding the `--topology MULTICLUSTER` to the original pilot prow job.

howardjohn:
That's fine, I am just wondering the long term plan?

stevenctl:
Long term we'd keep this new job until all tests under pilot/ are converted, then remove it and use `--topology MULTICLUSTER` to the original pilot prow job, unless we see value in also running in single cluster mode.

stevenctl:
>- skipped # if set, the test will run only in postsubmit or by explicitly calling /test on it

@howardjohn could I add this job with skipped or hidden for the time-being so I can iterate?

howardjohn (commented):
yeah just add it "skipped, hidden, optional" and you can do whatever. Then it should not bug anyone even if it fails 100% of the time

stevenctl:
@howardjohn should be good now

prow/config/jobs/istio.yaml
```diff
@@ -124,6 +124,20 @@ jobs:
...
+      - entrypoint
+      - prow/integ-suite-kind.sh
+      - --topology
+      - MULTICLUSTER
+      - test.integration.pilot.kube
+    requirements: [kind]
+    modifiers: [optional, hidden, skipped]
+    env:
+      - name: TEST_SELECT
+        value: "-multicluster"
```
howardjohn, line 139: is this right? I guess it is since these tests don't have the MC label?

howardjohn (approved):

prow/config/jobs/istio.yaml
stevenctl: Right

stevenctl: Is `type: postsubmit` going to prevent me from using `/test` to trigger? Not sure if I'm just impatient or if that's why `/test integ-pilot-multicluster-tests_istio_postsubmit` isn't working.

howardjohn: ah yeah, sorry. It actually makes sense as well; we don't want someone to run `/test release-istio_postsubmit` on an unapproved pr for example

Issue #2704, opened 2020-06-09
Autobump job is performing no operations


chases2:
The autobump job `ci-prow-autobump` was creating PRs to update Prow, but is no longer doing so.
Job on Prow: https://prow.istio.io/?job=ci-prow-autobump
Example log: https://prow.istio.io/log?job=ci-prow-autobump&id=1584

The script seems to claim that there is no change, but the upstream prow images are updating and have newer images that exist.

chases2:
A recent change to bump.sh was made that did contain a bug:
https://github.com/kubernetes/test-infra/pull/17649

The bug fixed the order in which images are checked and the bump change is made. A fix was merged for this: https://github.com/kubernetes/test-infra/pull/17897

However, the fix doesn't explain why plank_deployment.yaml left the list of files-to-bump. Without this file, `autobump.sh` won't detect that a change actually occurred and won't do anything.
Previous Bump Example: https://prow.istio.io/view/gs/istio-prow/logs/ci-prow-autobump/1525
Broken Bump Example: https://prow.istio.io/view/gcs/istio-prow/logs/ci-prow-autobump/1583

cjwagner:
It has been working recently.
/close

PR #2705, opened 2020-06-09
Revert autobump to a previous version

chases2:
Some change to bump.sh broke bumping. Since autobump is bumped to a broken version, programmatic tools to revert won't work until autobump is reverted.

Previous Bump Example: https://prow.istio.io/view/gs/istio-prow/logs/ci-prow-autobump/1525
Broken Bump Example: https://prow.istio.io/view/gcs/istio-prow/logs/ci-prow-autobump/1583

May address Issue #2704


clarketm (approved):
/hold

michelle192837 (approved):

PR #2707, opened 2020-06-10
Enable ipv6 job

howardjohn:


howardjohn:
had to revert the ubuntu change, so this won't work

jwendell (approved):

PR #2727, opened 2020-06-23
Fix the test select flag for istio

ZhengzheYang:
This PR aims to fix the bug described in https://github.com/istio/istio/issues/24909 and https://github.com/istio/istio/pull/24876. Should be reviewed together.

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -487,8 +487,6 @@ postsubmits:
         env:
         - name: INTEGRATION_TEST_FLAGS
           value: ' --istio.test.retries=1 '
-        - name: TEST_SELECT
-          value: -multicluster
         image: gcr.io/istio-testing/build-tools:master-2020-05-31T23-50-32
         name: ""
         resources:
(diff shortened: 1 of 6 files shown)


howardjohn (commented):
Why do we need this for istio-1.6.yaml? The new VM stuff is only in master right?

Also we need this for integ-k8s-117, integ-k8s-119 as well right?

ZhengzheYang:
> Why do we need this for istio-1.6.yaml? The new VM stuff is only in master right?
> 
> Also we need this for integ-k8s-117, integ-k8s-119 as well right?

Oh I looked at the log of https://prow.istio.io/view/gcs/istio-prow/logs/k8s-tests_istio.io_release-1.6_postsubmit/57 and I thought it fails because of the VM test. Sorry! I will fix it. The latter 2 doesn't seem to set the test select flag so I didn't touch. https://github.com/istio/test-infra/blob/1ec8984742eb9e80aabb747d1891a87c4245d43a/prow/config/jobs/istio.yaml#L199-L201

ZhengzheYang:
I think integ-k8s-117, integ-k8s-119 will stop failing if https://github.com/istio/istio/pull/24876 is merged since if no test_select flag is given, it will default to run pre-submit jobs, which is the correct behavior.

howardjohn (approved):
oh didn't realize it was just this one. thanks

PR #2736, opened 2020-06-29
update golangcli-lint to 1.27

stevenctl:
While working on https://github.com/istio/istio/pull/25046 I'm running into https://github.com/golangci/golangci-lint/issues/826, which should be fixed in 1.22+


howardjohn (commented):
The idea lgtm but this isn't the right place to do it. The config file is mirrored from istio/common-files and I am not sure what that docker file is - I think it may be dead code. https://github.com/istio/tools/commit/4e55f0c57b3de0a94860ac5bbfd236254d21c9ed does the exact same thing with what we use in ci, but it's possible the change didn't work its way through everywhere. The flow is a pr to Istio/tools then the image is built and pushed in postsubmit. Then we manually take the built image (from looking at the logs) and update common-files repo (a file called setup_env I think). Once that merges it will automatically propogate to all the repos which will impact what devs run locally. At the same time it will update test-infra to change what runs in CI.

So I thought I did all of those, but maybe not - I would check what build-tools version is used in the failing job and run `golangci-lint version` with that locally and see. I didn't update the golangci config yaml so not sure if that somehow impacts it

stevenctl:
Closing in favor of https://github.com/istio/common-files/pull/265

PR #2737, opened 2020-06-29
Update autobump to include splitting fix.

cjwagner:
Update to include https://github.com/kubernetes/test-infra/pull/18060


fejta (approved):

howardjohn:
This might have broken something?
https://prow.istio.io/view/gcs/istio-prow/logs/ci-prow-autobump/1711

cjwagner:
> This might have broken something? https://prow.istio.io/view/gcs/istio-prow/logs/ci-prow-autobump/1711

It did. It is no more functional than it was before though so I opted to wait for the forward fix which recently merged and published. Here is the fix: https://github.com/istio/test-infra/pull/2739

PR #2763, opened 2020-07-10
Add job for benchmark test in presubmit and postsubmit

howardjohn:
Hidden for now, to ensure I don't break something

For https://github.com/istio/istio/issues/18135

Depends on https://github.com/istio/istio/pull/25257

The intent is we have two jobs:

Postsubmit job runs benchmark, reports reports to bucket/SHA.txt
Presubmit job, optional and run on demand, runs benchmark and fetches
the latest results from postsubmit and compares the two.


howardjohn:
Also blocked on https://github.com/istio/tools/pull/1069

stevenctl (approved):

PR #2779, opened 2020-07-14
Use allowlist and denylist in genjobs and its config

tbarrella:
As part of https://github.com/istio/istio/issues/25381

I didn't find `-b` or `-w` actually being used, and `-a` was already taken, so I just removed them


prow/genjobs/README.md
```diff
@@ -161,4 +161,4 @@ genjobs --mapping istio=istio-private --clean
 - 0.0.4: add `defaults` key for specifying _file-level_ defaults, support a `.defaults.yaml` file for _local_ defaults, and add `--global` option for _global_ defaults.
 - 0.0.5: rename `--extra-refs` option to `--refs` and designate `extra-refs` key for specifying a list of extra refs to append to job.
 - 0.0.6: `--extra-refs` will now replace existing refs, rather than adding to them.
-- 0.0.7: add `--env-blacklist` and `volume-blacklist` options for pruning env and volume/volumeMount objects, respectively, from generated jobs.
\ No newline at end of file
+- 0.0.7: add `--env-blacklist` and `volume-blacklist` options for pruning env and volume/volumeMount objects, respectively, from generated jobs.
```
clarketm, line 164: To be backward compatible here, I would favor listing this as new patch version (e.g. `0.0.8`) or minor version but since it is in pre-release patch should be fine. Also, if you additionally bump the `VERSION` in the [`Makefile`](https://github.com/istio/test-infra/blob/d812274c96329e1ad33dcc7c2a409f82a36cfaab/prow/genjobs/Makefile#L17) this will trigger an _automatic_ deployment of the docker image in GCR by way of [push-genjobs](https://github.com/istio/test-infra/blob/d812274c96329e1ad33dcc7c2a409f82a36cfaab/prow/config/jobs/test-infra.yaml#L126-L139) jobs.

clarketm (changes requested):
lgtm with one nit. See my comment regarding bumping the patch version from ~`0.0.7`~ -> `0.0.8` rather than overwriting the existing release.

clarketm (approved):

ericvn (approved):

PR #2784, opened 2020-07-15
Use "denylist" in api-resources.py

tbarrella:
As part of https://github.com/istio/istio/issues/25381

It looks like values are never passed to override the defaults, making this seem safe

### DIFF
--- a/prow/api-resources.py
+++ b/prow/api-resources.py
@@ -79,15 +79,15 @@ def fetch_versions():
         return {}
 
 
-def main(delimiter, format, group_blacklist, kind_blacklist):
+def main(delimiter, format, group_denylist, kind_denylist):
     resources = fetch_resources()
     versions = fetch_versions()
-    group_blacklist = set(group_blacklist)
-    kind_blacklist = set(kind_blacklist)
+    group_denylist = set(group_denylist)
+    kind_denylist = set(kind_denylist)
     r = []
 
     for name, _, group, _, kind in resources:
-        if kind in kind_blacklist or group in group_blacklist:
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

clarketm (approved):

prow/api-resources.py
```diff
@@ -109,18 +109,18 @@ def main(delimiter, format, group_blacklist, kind_blacklist):
         "--format", type=str, default="%s", help="format string for each api resource."
     )
     parser.add_argument(
-        "--group-blacklist",
+        "--group-denylist",
```
cjwagner, line 112: Aren't these breaking changes? I would have expected both flags to be supported temporarily or for the uses of the script to be updated in this PR as well.

tbarrella: I couldn't find these flags being used in any Istio repo; it seems like only the default value is ever used

cjwagner (approved):
/hold
Ah I see in the PR body that you say the flags are never used.

prow/api-resources.py
tbarrella: See for example #2551

tbarrella:
/hold cancel
Merging since @clarketm approved and has been the only previous contributor to the file

prow/api-resources.py
clarketm: Yes, I added these flags to make these attributes configurable (for any extended usage) but made is so we are using the _default_ values in this repo. And AFAIK this is not used anywhere else. So this should not be breaking for Istio.

PR #2790, opened 2020-07-16
Make proxy update optional and use build-tools image

ericvn:


howardjohn (approved):
I don't think optional does anything for postsubmit

ericvn:
@howardjohn Thanks. I was wondering about that as I'm not sure what happens on a failure.

brian-avery (approved):

PR #2801, opened 2020-07-16
Update Automator to do make clean gen

jasonwzm:
Fix issue in client-go where `make gen` does not produce the complete output.

It might take longer for istio/istio to finish the job, but it is postsubmit job.


howardjohn (approved):
Isn't it checking out an empty repo? what is it cleaning?

jasonwzm:
cleaning the generated files. For `make gen`, if the file exists, it won't regenerate again.

ericvn:
I would have thought the job would have just cloned the repo to a new directory and that one would have been starting with a clean repo to begin with.

PR #2808, opened 2020-07-17
Make AUTOMATOR job to update proxy dependency auto-merge

ericvn:


jwendell (commented):
Tests are going to fail because the new proxy binaries are not yet available at the time the PR is created, no?

ericvn:
@jwendell I was actually worried about that. Would require a `/retest` to get them to pass once the new binaries are available. @clarketm Are there any ways to cause a post-submit to wait for another post-submit. We could maybe cause a wait until the binaries are available before running the automator task? Have to set a timeout long enough for that to happen possibly.

howardjohn:
> Are there any ways to cause a post-submit to wait for another
post-submit. We could maybe cause a wait until the binaries are available
before running the automator task

There is no built in way in prow to do this as far as I know, you can do
some polling logic in the job potentially or something?

ericvn:
For a note, I see:
```
Downloading envoy: curl -fLSs https://storage.googleapis.com/istio-build/proxy/envoy-alpha-51d27798450ec4be430111eceb1ffd08041a3a50.tar.gz to /home/prow/go/src/istio.io/istio/out/linux_amd64/release/envoy-51d27798450ec4be430111eceb1ffd08041a3a50
curl: (22) The requested URL returned error: 404 
```
So possibly something I can poll on. It looks like the job takes 30-60 minutes to generate.

clarketm:
> @clarketm Are there any ways to cause a post-submit to wait for another post-submit. We could maybe cause a wait until the binaries are available before running the automator task? Have to set a timeout long enough for that to happen possibly.

As @howardjohn pointed out the simplest approach is to poll or wait for the condition to be met using code logic inside the job itself. An alternative option is to combine the jobs which have a dependency on one another and execute the logic (again in code) serially. Lastly, a more experimental approach (but definitely the most _progressive_ option) is to use a Tekton pipeline (https://github.com/kubernetes/test-infra/issues/13874). The latter will require some additional setup as well as some digging into the documentation since we do not use Tekton in Istio ATM.

ericvn:
Polling added in https://github.com/istio/istio/pull/25651 to delay updating dependency until the envoy tar is available.

Added a timeout to the dependency job to make that of the build job.

jwendell (approved):

PR #2809, opened 2020-07-17
Add Istio CNI install test job

bcheung:
This adds the integration test job mentioned in [this comment](https://github.com/istio/istio/pull/25332#pullrequestreview-450606543).

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -208,6 +208,11 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: " --istio.test.retries=1 "
 
+  - name: install-cni-test
+    modifiers: [hidden, optional]
+    command: [make, cni.install-test]
+    requirements: [docker]
+
   - name: analyze-tests
     modifiers: [optional, skipped]
     type: presubmit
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

prow/config/jobs/istio.yaml
```diff
@@ -208,6 +208,11 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: " --istio.test.retries=1 "
 
+  - name: install-cni-test
+    modifiers: [hidden, optional]
```
jwendell, line 212: What does `hidden` mean? I'd like this job results to appear on github UI so that we are aware of its results.
+1 for keeping it optional for a while, once it proves stabilized we can make it required.

jwendell:
This is waiting on https://github.com/istio/istio/pull/25332

howardjohn:
The intent of hidden is to make sure it actually passes. If we are
confident it will we can skip that step, or just revert if it fails. But
having a broken test apply to peoples PR, even if its optional, is not a
good experience

jwendell:
Ok, makes sense. Now that the main PR is merged, this is ready to go

jwendell (approved):

jwendell:
@howardjohn how can we see the health of this job? It seems they don't show up in testgrid.

howardjohn:
It will show up in testgrid but its dependant on approval in https://github.com/kubernetes/test-infra/ (automatic PR, not automatic merge). Until then: https://prow.istio.io/?job=install-cni-test_istio

howardjohn:
Actually i am not sure if its in testgrid if its hidden.

Seems its not 100% stable? https://prow.istio.io/view/gs/istio-prow/logs/install-cni-test_istio_postsubmit/13

Issue #2731, opened 2020-06-24
Cannot use NFS mounts in prow, errors during pod startup


howardjohn:
https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_istio/24928/lint_istio/14574

Get errors like

```
(combined from similar events): MountVolume.SetUp failed for volume "fileserver" : mount failed: exit status 1 Mounting command: systemd-run Mounting arguments: --description=Kubernetes transient mount for /var/lib/kubelet/pods/59b58feb-8b9f-4592-aa44-7057eb4b5439/volumes/kubernetes.io~nfs/fileserver --scope -- /home/kubernetes/containerized_mounter/mounter mount -t nfs 10.3.165.146:/istio_prow_cache /var/lib/kubelet/pods/59b58feb-8b9f-4592-aa44-7057eb4b5439/volumes/kubernetes.io~nfs/fileserver Output: Running scope as unit: run-r9d99c0f687024dfcadad64cbcb4ee325.scope Mount failed: Mount issued for NFS V3 but unable to run rpcbind: Output: rpcbind: another rpcbind is already running. Aborting Error: exit status 1
```

howardjohn:
Not used anymore

PR #2813, opened 2020-07-20
Another attempt at go module caching

howardjohn:
previously we tried NFS volumes:
https://github.com/istio/test-infra/issues/2731

This has two issues, one that NFS is slow and 2 that there was
apparently a race at pod startup.

This PR once again tries to add caching, but this time using HostPaths.
I add 2 hidden/skipped jobs to experiment with this. In addition, I
removed the 1.6 jobs using the NFS mount. This will ensure they are not
impacted by this change, and will allow me to delete the NFS drive (its
kind of expensive)


howardjohn:
I tested this locally by the way, seems to work at least at a basic level

stevenctl (approved):
Nice! Out of curiosity, how many nodes do we run in the prow cluster?

howardjohn:
> Nice! Out of curiosity, how many nodes do we run in the prow cluster?

Autoscaling between 3 and 60. Usually ~20 or so on a normal busy day.

PR #2817, opened 2020-07-20
Add cache to all jobs

howardjohn:
I am somewhat concerned about the blast radius of this change, but at the same time it seems to lead to a decent improvement to test times without any introduced flakes, so I figured we can always revert it if it ends up causing some issues

### DIFF
--- a/prow/cluster/jobs/istio-private/envoy/istio-private.envoy.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/envoy/istio-private.envoy.master.gen.yaml
@@ -35,8 +35,17 @@ presubmits:
             memory: 180G
         securityContext:
           privileged: true
+        volumeMounts:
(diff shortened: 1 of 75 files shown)


sdake:
John, I would approve, but there are multiple reviewers.

Looks fine to me. Two comments:

- /tmp is a ram filesystem on most/all Linux distributions. This will consume some extra ram. I think this is a good improvement. I had this in the original volume mounts, but for some reason Martin removed it. We never cached this information in CI though.
- I am hopeful this does not generate race conditions. I am not sure golang uses fcntl() to lock a file, but it may not matter. It really depends on how golang implements the file read/write logic. I am pretty sure this type of cache would not work on bezel.

This can be reverted if this triggers problems.

The problems should be obvious: test failures when building the software, but not in other conditions.

ericvn (approved):

PR #2818, opened 2020-07-21
Increase scheduling timeout

howardjohn:
Unlike K8s prow (I think), we have autoscaling on our build cluster. As
a resultm during spikes in tests we see pod scheduling timeouts, as the
cluster autoscaling can take a few minutes for large increases. As a
result the tests fail. I don't see a reason to fail fast when waiting is
very likely to be succesful, so I propose we wait longer.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -5,7 +5,7 @@ plank:
     '*': https://prow.istio.io/view/gcs/
     istio-private: https://prow-private.istio.io/view/gcs/
   pod_pending_timeout: 15m
-  pod_unscheduled_timeout: 5m
+  pod_unscheduled_timeout: 30m
   default_decoration_configs:
     '*':
       timeout: 2h


prow/config.yaml
```diff
@@ -5,7 +5,7 @@ plank:
     '*': https://prow.istio.io/view/gcs/
     istio-private: https://prow-private.istio.io/view/gcs/
   pod_pending_timeout: 15m
-  pod_unscheduled_timeout: 5m
+  pod_unscheduled_timeout: 30m
```
cjwagner, line 8: This seems a little excessive to me since it will make unschedulable jobs sit for 30mins before failing and autoscaling probably shouldn't take that long. This is easy to tweak though.

cjwagner (approved):

fejta (approved):

PR #2831, opened 2020-07-22
Remove unnecessary perf deploy postsubmit job

carolynhu:
These jobs might be override the master deploy job pushes, so I remove them here.


howardjohn (commented):
Need to run `make generate-config`

ericvn (approved):

PR #2832, opened 2020-07-22
Experimental: cache go build results as well as modules

howardjohn:
This adds support for caching of build artifacts. Currently we cache go
modules only. I have added an experimental job with this enable to test
stability/perf.

One caveat here is we *probably* want to disable test caching, so that
we aren't skipping tests. In theory it might be safe to keep it on, but
I think it makes things confusing and hides test flakes, etc


howardjohn:
@ericvn let me know what you think. I expect this to save us ~3 minutes on each test

ericvn (approved):

PR #2833, opened 2020-07-22
Add prow.k8s.io/gerrit-report-label=Verified label for Prow jobs running with Gerrit

chizhg:
For Prow jobs running with Gerrit, we need `prow.k8s.io/gerrit-report-label=Verified` label to block on presubmit.

/cc @howardjohn

### DIFF
--- a/prow/genjobs/cmd/genjobs/main.go
+++ b/prow/genjobs/cmd/genjobs/main.go
@@ -100,6 +100,7 @@ type transform struct {
 	Resolve          bool              `json:"resolve,omitempty"`
 	SSHClone         bool              `json:"ssh-clone,omitempty"`
 	OverrideSelector bool              `json:"override-selector,omitempty"`
+	Internal         bool              `json:"internal,omitempty"`
 	Verbose          bool              `json:"verbose,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


duderino (approved):

prow/genjobs/cmd/genjobs/main.go
```diff
@@ -152,6 +153,7 @@ func (o *options) parseOpts() {
 	flag.BoolVar(&o.Resolve, "resolve", false, "Resolve and expand values for presets in generated job(s).")
 	flag.BoolVar(&o.SSHClone, "ssh-clone", false, "Enable a clone of the git repository over ssh.")
 	flag.BoolVar(&o.OverrideSelector, "override-selector", false, "The existing node selector will be overridden rather than added to.")
+	flag.BoolVar(&o.Internal, "internal", false, "Generate Prow jobs for the internal repos.")
```
howardjohn, line 156: I am not sure what `internal` means. Maybe this would make sense as something like "support gerrit reporting" (bool) or "report label" (string)?

chizhg: You are right, `internal` is a pretty confusing word. I have created https://github.com/istio/test-infra/pull/2839 to change to `support-gerrit-reporting` as you suggested.

PR #2838, opened 2020-07-23
Apply the release-notes-none label to cherrypicks

jwendell:


ericvn (dismissed):

prow/cluster/cherrypicker_deployment.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         - --dry-run=false
         - --use-prow-assignments=false
         - "--labels=cla: yes"
+        - --labels=release-notes-none
```
ericvn, line 27: Is the double --labels going to cause an issue?

ericvn: It looks like GitHub-endpoint has multiple.

jwendell: on the contrary. It only works this way. Comma separated will be interpreted as only one string.

jwendell: Played locally with https://github.com/kubernetes/test-infra/blob/master/prow/external-plugins/cherrypicker/main.go#L65 and https://github.com/kubernetes/test-infra/blob/master/prow/flagutil/strings.go

jwendell: Result of `fmt.Printf("%d - %#v\n", len(o.labels.Strings()), o.labels.Strings())`:
```
 ~/src/k8s/test-infra/prow/external-plugins/cherrypicker [master|✚ 1] 
12:23 $ go run main.go --labels=x1,x2,x3
1 - []string{"x1,x2,x3"}

✔ ~/src/k8s/test-infra/prow/external-plugins/cherrypicker [master|✚ 1] 
12:23 $ go run main.go --labels=x1 --labels=x2 --labels=x3
3 - []string{"x1", "x2", "x3"}
```

ericvn (approved):

prow/cluster/cherrypicker_deployment.yaml
ericvn: Thanks

jwendell: > It looks like GitHub-endpoint has multiple.

yeah, they use the same backend to parse array of strings

brian-avery: Thanks.

brian-avery (approved):

howardjohn (approved):

prow/cluster/cherrypicker_deployment.yaml
brian-avery, line 27: Any idea why the prior one is in quotes and this one is not?

brian-avery: I guess the space would change it.

jwendell: space. it could be the binary would interpret `yes` as another argument.

fejta (approved):

PR #2846, opened 2020-07-28
Update script to check for release notes files rather than querying GitHub

brian-avery:
With the current release notes script, we are performing two GitHub API operations for every pull request. One of these is necessary to get the labels. The other is unnecessary as we already have a git clone. This pull request updates the verification to use the cloned repo for information on added files instead of using the API. 

Also, the GitHub API limits the number of files returned from a pull request to 30 before it starts paginating. Some pull requests in Istio are quite a bit larger. This fixes that such that it will return the entire list of added/renamed/modified files instead of just looking at the first 30 files.

### DIFF
--- a/tools/check_release_notes.sh
+++ b/tools/check_release_notes.sh
@@ -111,6 +111,21 @@ validate_opts() {
         print_error_and_exit "PULL_NUMBER is a required option. It must match the GitHub pull request number."
         exit 1
     fi
+
+    if [ -z "${REPO_PATH:-}" ]; then
+        echo "REPO_PATH not specified. Using current working directory."
+        REPO_PATH=$(pwd)
+    fi
+
+    if [ -z "${PULL_PULL_SHA:-}" ]; then
+        echo "PULL_PULL_SHA not specified. This must match the HEAD SHA for the pull request."
+        exit 1
+    fi
+
+    if [ -z "${PULL_BASE_REF:-}" ]; then
+        echo "PULL_BASE_REF not specified. This must match the target branch for the pull request."
+        exit 1
+    fi
 }
 
 # Curl the GitHub API to get a list of files for the specified PR. If files are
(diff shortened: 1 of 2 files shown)


tools/check_release_notes.sh
```diff
@@ -99,27 +111,38 @@ validate_opts() {
...
-        ghFiles=$(curl -L -s --show-error --fail \
-            --header "Accept: application/vnd.github.v3+json" \
-            --header "Authorization: Bearer ${token}" \
-            "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/pulls/${PULL_NUMBER}/files")
-    fi
+    echo "Checking files from pull request ${REPO_OWNER}/${REPO_NAME}#${PULL_NUMBER} head SHA: ${PULL_PULL_SHA} destination branch: ${PULL_BASE_REF}"
+
+    pushd "${REPO_PATH}"
+
+    addedFiles=$(git diff "${PULL_BASE_REF}...${PULL_PULL_SHA}" --name-only --diff-filter=AMR)
```
jwendell, line 141: Does it work with three dots? In my system it's only two.

tools/check_release_notes.sh
```diff
@@ -99,27 +111,38 @@ validate_opts() {
...
+    echo "Added files: ${addedFiles}"
+    echo ""
```
jwendell: nit: just `echo` works fine

tools/check_release_notes.sh
```diff
@@ -99,27 +111,38 @@ validate_opts() {
         print_error_and_exit "PULL_NUMBER is a required option. It must match the GitHub pull request number."
         exit 1
     fi
+
+    if [ -z "${REPO_PATH:-}" ]; then
+        echo "REPO_PATH not specified. Using current working directory."
```
jwendell: nit: I'd move this down in this section, leaving the fatal errors first, and warnings like this last. You could also print the current directory, it might be useful for debugging. I know that this is used only outside prow though.

tools/check_release_notes.sh
jwendell: sorry for the noise, you are right. I was testing on my system using wrong order of branches.

jwendell (commented):
LGTM

jwendell (approved):

PR #2867, opened 2020-08-07
Update common-files and fix lint issue

howardjohn:


scripts/update-e2e-cluster.sh
```diff
@@ -1,131 +0,0 @@
-#!/bin/bash
```
jwendell, line 1: This file is not used anymore? Do you want to update README then?

howardjohn: Yes, probably could clean up more as well. I just updated the readme

jwendell (approved):

PR #2877, opened 2020-08-19
use standard build tools for community repo

howardjohn:
Needed so we can execute some go code


howardjohn:
Needs https://github.com/istio/community/pull/396

linsun (approved):

PR #2883, opened 2020-08-24
Add upgrade-downgrade periodic jobs

carolynhu:
TODO: need to update cron time


carolynhu:
we might not need this test

PR #2886, opened 2020-08-27
Add validation to release notes

brian-avery:
This PR updates the release-notes tooling to execute validation of release notes based on the gen-release-notes utility.

### DIFF
--- a/tools/check_release_notes.sh
+++ b/tools/check_release_notes.sh
@@ -151,11 +151,15 @@ checkForFiles() {
     # gen-release-notes returns EX_NOINPUT (return code 66) for files not found. Any other error codes or return codes are valid reasons to exit.
     EX_NOINPUT=66
     if [ "${returnCode}" -eq ${EX_NOINPUT} ]; then
+        echo "No release notes files found in '/releasenotes/notes/'."
         echo ""
+        popd
         return
+    else
+        echo "found release notes entries"
+        #gen-release-notes will either return 0 (no error) or an error.
+        exit "${returnCode}"
     fi
-    exit "${returnCode}"
-    popd
 
 }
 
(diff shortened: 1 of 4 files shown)


brian-avery:
This depends on https://github.com/istio/tools/pull/1186 being merged first.

tools/check_release_notes.sh
```diff
@@ -138,22 +138,29 @@ checkForFiles() {
...
+    if [ "${returnCode}" -eq ${EX_NOINPUT} ]; then
         echo "No release notes files found in '/releasenotes/notes/'."
-        echo
+        echo ""
+        popd
+        return
     else
-        echo "Found release notes entries"
-        exit 0
+        echo "found release notes entries"
```
jwendell: According to the comment below, this might not be true. I mean, an error can occur, right?

tools/check_release_notes.sh
```diff
@@ -138,22 +138,29 @@ checkForFiles() {
...
-    if [ -z "${releaseNotesFiles}" ]; then
+
+    # gen-release-notes returns EX_NOINPUT (return code 66) for files not found. Any other error codes or return codes are valid reasons to exit.
+    EX_NOINPUT=66
+    if [ "${returnCode}" -eq ${EX_NOINPUT} ]; then
         echo "No release notes files found in '/releasenotes/notes/'."
-        echo
+        echo ""
+        popd
+        return
```
jwendell: You can remove this `return`.

tools/check_release_notes.sh
brian-avery: That's true. And this would be after it saying "I found release notes entries. Processing foo.yaml, bar.yaml, etc...." "Oh by the way, I found release notes entries!"

jwendell (approved):

PR #2769, opened 2020-07-10
Remove clarketm from CODEOWNERS

brian-avery:

PR #2891, opened 2020-08-31
Add @chaodaiG and @e-blackwelder to CODEOWNERS

e-blackwelder:
New oncall rotation members.

(Noticed @clarketm was still listed under `/authentikos/`, but that's expected per #2769)


howardjohn:
You'll need to join https://github.com/istio/community/blob/master/org/developers.yaml otherwise it will break CODEOWNERS

e-blackwelder:
Gotcha. It will probably be best for Chao and I to join that individually/separately. I'll close this PR for now (and ask my team to assist with oncall responsibilities in the short-term). Thanks!

PR #2909, opened 2020-09-09
Update boskos and fix the deployments

### LINKED ISSUE #2806: Boskos CPU/memory leak
If you look at the boskos nodes in our prow cluster, the boskos nodes are constantly leaking memory and using max CPU. I suspect something is going wrong here since we only use boskos for ~4 jobs per day

chizhg:
1. Update boskos component images to the latest
2. Some CRD definitions need to be added after updating the image version, as what has been done in https://github.com/knative/test-infra/pull/1720

Fixes https://github.com/istio/test-infra/issues/2806

### DIFF
--- a/boskos/cluster/boskos-deployment.yaml
+++ b/boskos/cluster/boskos-deployment.yaml
@@ -68,7 +68,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
       - name: boskos
-        image: gcr.io/k8s-prow/boskos/boskos:v20190730-dbac84bd7
+        image: gcr.io/k8s-staging-boskos/boskos:v20200819-984516e
         args:
         - --config=/etc/config/resources.yaml
         - --namespace=boskos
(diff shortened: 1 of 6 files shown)


boskos/cluster/boskos-deployment.yaml
```diff
@@ -68,7 +68,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
       - name: boskos
-        image: gcr.io/k8s-prow/boskos/boskos:v20190730-dbac84bd7
+        image: gcr.io/k8s-staging-boskos/boskos:v20200819-984516e
```
carolynhu, line 71: @chizhg I saw the knative PR you referenced is using /k8s-prow/ image, why do we switch to use k8s-staging-boskos?

chizhg: Boskos has been moved to a separate repo in https://github.com/kubernetes-sigs/boskos, and k8s-staging-boskos is the new registry it's using, plus Knative has also switched to it - https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/knative/cluster/build/400-boskos-deployment.yaml#L36

carolynhu: I see, thanks for your explanation!

carolynhu (approved):
@chizhg Thanks for this fix!

PR #2931, opened 2020-09-17
Make boskos is able to provision clusters from rapid release channel

carolynhu:
https://github.com/istio/test-infra/issues/2910

### DIFF
--- a/boskos/resources.yaml
+++ b/boskos/resources.yaml
@@ -35,7 +35,9 @@ resources:
         - clusters:
           - machinetype: n1-standard-32
             numnodes: 3
-            version: 1.16
+            version: 1.17
+            releasechannel:
+              channel: RAPID
             zone: us-central1-f
             scopes:
             - https://www.googleapis.com/auth/cloud-platform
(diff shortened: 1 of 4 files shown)


boskos/resources.yaml
```diff
@@ -35,7 +35,9 @@ resources:
         - clusters:
           - machinetype: n1-standard-32
             numnodes: 3
-            version: 1.16
+            version: 1.17
+            releasechannel:
+              channel: RAPID
```
howardjohn, line 40: isn't it channel or version? how does both work?

howardjohn (approved):

boskos/resources.yaml
carolynhu: They are in different config setting struct.
- `version` is in `ServerConfig`, where `ServerConfig` also has a field called `channel`.
- `ReleaseChannel` field is in &container.Cluster, which is used in the gke.go create cluster func.
Not quite clear about their differences. Seems a kind of overlapping.

Issue #2937, opened 2020-09-21
Add automator for updating istio/envoy SHA in istio/proxy periodically


carolynhu:
As a prepare for easing the difficulties of regression detection. We need to frequently update istio/envoy SHA in istio/proxy so that once a regression is detected, we do not have to tracking back a few months of code changes.

ericvn:
Note https://github.com/istio/istio/issues/24586 is for an update in the release streams to update istio/proxy when istio/envoy changes. This doesn't work for `master` since we don't have a way to automate on the upstream repo.

PR #2938, opened 2020-09-22
Automator update envoy SHA to istio/proxy + Add AUTOMATOR_SHA_COMMIT_DATE

carolynhu:
https://github.com/istio/test-infra/issues/2937


prow/config/jobs/envoy.yaml
```diff
@@ -33,6 +33,22 @@ jobs:
...
+  type: postsubmit
+  command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=proxy
+  - "--title=Automator: update envoy@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+  - --labels=auto-merge,release-notes-none
+  - --modifier=update_envoy_dep
+  - --token-path=/etc/github-token/oauth
+  - --cmd=bin/update_envoy.sh $AUTOMATOR_SHA
```
howardjohn: Will you be adding this?

carolynhu: yes! working on that

prow/config/jobs/envoy.yaml
```diff
@@ -33,6 +33,23 @@ jobs:
...
 
+periodics:
+- interval: 168h
+  name: update-proxy
+  command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=proxy
+  - "--title=Automator: update envoy@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+  - --labels=auto-merge,release-notes-none
```
bianpengyuan: nit: proxy repo does not need release notes label.

prow/config/jobs/envoy.yaml
```diff
@@ -33,6 +33,23 @@ jobs:
...
+  - --modifier=update_envoy_dep
```
bianpengyuan, line 45: Not familiar with the configuration here, but is `update_envoy_dep` a keyword that points to something? or just a random name?

bianpengyuan (commented):
How do we make sure this job only runs on master?

prow/config/jobs/envoy.yaml
carolynhu: https://github.com/istio/test-infra/tree/master/tools/automator `A segment of text to append to the branch name of the fork for additional context. This option is optional and will default to automator.sh.`

prow/config/jobs/envoy.yaml
```diff
@@ -33,6 +33,23 @@ jobs:
     value: "/home/prow/go/src/istio.io/envoy"
   command: [./ci/do_ci.sh, bazel.release]
 
+periodics:
+- interval: 168h
```
bianpengyuan: nit: do we support something like `1w` or `7d`?

bianpengyuan (approved):

carolynhu:
put on hold: need to verify this periodic job config

prow/config/jobs/envoy.yaml
carolynhu: sorry, it's golang parseduration impl, in which `Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`

howardjohn (approved):

PR #2967, opened 2020-10-01
dedicate node for pilot-mc job

stevenctl:
We can probably tune this so it takes up "half" a node. Just want to see if this makes the tests stable. 

---

I think we're hitting fd limits on our prow nodes.. I spot checked one of the ubuntu test pool nodes and they have ubuntu's default of just over 1 million as the hard limit for a single user. 

On my personal machine (8 CPU, 32GB ram), running Ubuntu with nothing else, I ran this job like we would in prow, using 5 kind clusters and consistently saw over 400k file descriptors being used via `lsof`. 5+ runs with no failures. 

fds

I manually set the limit to 300,000 `ulimit -Hn 300000` and was able to replicate the sort of flakes seen here [integ-pilot-multicluster-tests_*](https://prow.istio.io/?job=integ-pilot-multicluster-tests_*)

```
=== RUN   TestTraffic/sniffing/grpc_naked->a_from_cluster-3
2020-10-01T23:05:18.819352Z	info	tf	=== BEGIN: Test: 'pilot[TestTraffic/sniffing/grpc_naked->a_from_cluster-3]' ===
    traffic.go:75: call for grpc naked->a from cluster-3/0 failed, retrying: failed calling naked->'grpc://a.echo-1-85969.svc.cluster.local:7070/': rpc error: code = Unknown desc = 5/25 requests had errors; first error: rpc error: code = DeadlineExceeded desc = context deadline exceeded
    traffic.go:75: call for grpc naked->a from cluster-3/0 failed, retrying: failed calling naked->'grpc://a.echo-1-85969.svc.cluster.local:7070/': rpc error: code = Unknown desc = 5/25 requests had errors; first error: rpc error: code = Unavailable desc = upstream connect error or disconnect/reset before headers. reset reason: local reset
    traffic.go:72: retry.UntilSuccessOrFail: timeout while waiting (last error: failed calling naked->'grpc://a.echo-1-85969.svc.cluster.local:7070/': rpc error: code = Unknown desc = 5/25 requests had errors; first error: rpc error: code = Unavailable desc = upstream connect error or disconnect/reset before headers. reset reason: local reset)
2020-10-01T23:05:28.942708Z	info	tf	=== DONE (failed):  Test: 'pilot[TestTraffic/sniffing/grpc_naked->a_from_cluster-3] (10.123348298s)' ===
```

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -377,8 +377,8 @@ postsubmits:
           limits:
             memory: 24Gi
           requests:
-            cpu: "5"
-            memory: 3Gi
+            cpu: "15"
+            memory: 8Gi
         securityContext:
           privileged: true
         volumeMounts:
@@ -1610,8 +1610,8 @@ presubmits:
           limits:
             memory: 24Gi
           requests:
-            cpu: "5"
-            memory: 3Gi
+            cpu: "15"
+            memory: 8Gi
         securityContext:
           privileged: true
         volumeMounts:
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
@@ -385,8 +385,8 @@ postsubmits:
           limits:
             memory: 24Gi
           requests:
-            cpu: "5"
-            memory: 3Gi
+            cpu: "15"
+            memory: 8Gi
         securityContext:
           privileged: true
         volumeMounts:
@@ -1515,8 +1515,8 @@ presubmits:
           limits:
             memory: 24Gi
           requests:
-            cpu: "5"
-            memory: 3Gi
+            cpu: "15"
+            memory: 8Gi
         securityContext:
           privileged: true
         volumeMounts:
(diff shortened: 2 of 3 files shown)


howardjohn:
we get these errors for single cluster as well. maybe we can tune the limits? we already have a daemonset tuning some other setting (open file watches I think).

1 million fd seems like a lot - is that an unexpected amount?

howardjohn (approved):

howardjohn:
seems like on my machine (glinux laptop) the limit is `32768`, and I can run the tests without issues.. why is the multicluster test using 10x+ that amount? I would expect at most 5x since there are 5 clusters

stevenctl:
My guess is that things aren't closing properly. The number of TIME_WAIT conns doesn't correlate though. 

Are you checking with ulimit -Hn? I think the soft limit is fine. I'm not super confident this is the issue either.. because you would expect more things to fail.

howardjohn:
Yeah -
```
$ ulimit -Hn
32768
```

Note this is not ubuntu though. Also `lsof | wc -l` gives me 250k lines

stevenctl:
I'll have to rerun this with cat /proc/sys/fs/file-nr

howardjohn:
if it helps:
```
$ cat /proc/sys/fs/file-nr
30699   0       9223372036854775807
```

suspiciously close to my limit...

stevenctl:
With file-nr I only saw a 17k difference from baseline. We have enough FDs in the ulimit for 50+ multicluster jobs on a single node, the resource requests we have set should keep us well below that. 

Screen Shot 2020-10-02 at 12 00 55 PM

PR #2977, opened 2020-10-12
Prow monitoring make rules: make sure using correct cluster

chaodaiG:


prow/cluster/monitoring/mixins/Makefile
```diff
@@ -17,6 +17,8 @@
```
cjwagner, line 17: This file also needs to include these parameters:
https://github.com/istio/test-infra/blob/7164f2aa151b18a39fe7e20850e4916664d78cf3/prow/Makefile#L8-L15
so that targets works as intended when used directly (rather than from the `prow/` Makefile).

cjwagner (changes requested):

prow/cluster/monitoring/mixins/Makefile
chaodaiG: That's right, fixed

cjwagner (approved):

PR #2992, opened 2020-10-20
Generate 1.8 branch jobs

bianpengyuan:


prow/config/jobs/tools-1.8.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - docker
+  resources: build
+  type: presubmit
+- command:
+  - entrypoint
+  - make
+  - -C
+  - perf_dashboard
+  - deploy
+  name: deploy-perf-dashboard
```
howardjohn, line 49: this seems suspicious @carolynhu do we only want this job on master to avoid reverting changes to master?

howardjohn (approved):

ericvn (approved):

prow/config/jobs/tools-1.8.yaml
ericvn: Would just need to add  disable_release_branching: true on the job in master so this doesn't happen in the future and remove here.

PR #3017, opened 2020-10-29
Delete plank resources (no longer used)

fejta:


fejta:
Will you leave an approving review please? That is what istio requires (doesn't recognize the /lgtm)

chaodaiG (approved):

PR #3021, opened 2020-10-30
Add Helm3 integration tests to CI

nrjpoddar:
Create new presubmit Prow check for running Helm3 integration tests

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -170,6 +170,11 @@ jobs:
       - name: TEST_SELECT
         value: "-multicluster"
 
+  - name: integ-helm-tests
+    type: presubmit
+    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.helm.kube.presubmit]
+    requirements: [kind]
+
     # The node image must be kept in sync with the kind version we use.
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -170,6 +170,11 @@ jobs:
       - name: TEST_SELECT
         value: "-multicluster"
 
+  - name: integ-helm-tests
+    type: presubmit
```
howardjohn: We should add this as both presubmit and postsubmit to track flakes, etc like all of our other tests. and I recommend adding `modifiers: [hidden, optional]` for a while so we can ensure it stabilizes before we block every PR on it as we have not proven it is a stable test yet

nrjpoddar: Makes sense, done

nrjpoddar:
@howardjohn let me know if anything else needs to be done for this PR.

howardjohn (approved):

prow/config/jobs/istio.yaml
nrjpoddar: I don't know how to add it for both pre & post submit. I deleted the type which based on README makes it run for both. What target should i use?

PR #3033, opened 2020-11-02
Add metrics ports for prow/crier and prow/horologium

chaodaiG:
So that prometheus can monitor them


howardjohn (approved):
You can also use PodMonitor to do this without a Service FWIW

e-blackwelder (approved):

Issue #2659, opened 2020-05-11
Idea: Extending Authentikos to other credential types


wlynch:
Hi!

Over at the [Tekton](https://tekton.dev) project, we've been thinking about how to handle GitHub App credentials. We ended up leaning towards a [refresher design](https://docs.google.com/document/d/1dauJtzMWwxutwV38FwtSpauBhA9nKLNjZTIrYEjSaao/edit) just like Authentikos, since GitHub App short-lived installation tokens act very similar to Google OAuth access tokens. It would be great if we could extend Authentikos to support this. :D

I wanted to reach out to find out if this would this be of interest to the Authentikos project (i.e. should we upstream changes here or fork our own version)?

Thanks!

clarketm:
Hi @wlynch,

I read the proposal. It would be ideal to coalesce the [refresher design](https://docs.google.com/document/d/1dauJtzMWwxutwV38FwtSpauBhA9nKLNjZTIrYEjSaao/edit) proposal (and support for new credential types) with upstream [Authentikos](https://github.com/istio/test-infra/tree/master/authentikos) (given the similarities). Currently, Authentikos is quite configurable but we can extend it as needed to support Tekton's needs. I would be more than happy to assist in any way I can! Alternatively, if you prefer to fork that is fine as well.

wlynch:
Great!
We have interns ramping up on the team right now, and I think this might make a good project. We'll circle back in a few weeks with a more concrete design w.r.t. integrating with Authentikos for review!

PR #3036, opened 2020-11-02
Authentikos: Refactor Token/Secret creators to interfaces.

wlynch:
This allows for other token types to be created while reusing the same
common creation / reconcilation components.

As an example for how this can be extended, adds a GitHub App token
creator as an alternative to the existing Google OAuth implementation.

Also changes secret updates to patches of just the data to fix an issue
dicovered during testing where the reconciler would overwrite any
user-added labels/annotations.

Implements https://github.com/istio/test-infra/issues/2659


repos.bzl
```diff
@@ -2,944 +2,58 @@ load("@bazel_gazelle//:deps.bzl", "go_repository")
 
 def go_repositories():
```
wlynch, line 3: Bit of a large diff here, though looks like a lot of reorderings - this was generated via `gazelle update-repos -from_file=go.mod -to_macro=repos.bzl%go_repositories` from the authentikos directory. Let me know if there's a better way to update this file!

PR #3041, opened 2020-11-03
Cleanup unused folder

carolynhu:


howardjohn (commented):
This is used by docker/prowbazel which is in turn used by the 1.4 jobs. probably should remove all of these

howardjohn (approved):
LGTM,  i asked envoy team to double check its not used so put hold until then

carolynhu:
> LGTM, i asked envoy team to double check its not used so put hold until then

yes, sure

PR #3044, opened 2020-11-05
Install Metallb for single cluster too

gargnupur:


common/scripts/kind_provisioner.sh
```diff
@@ -155,6 +155,9 @@ EOF
   if [[ -n ${METRICS_SERVER_CONFIG_DIR} ]]; then
     kubectl apply -f "${METRICS_SERVER_CONFIG_DIR}"
   fi
+
+  # Install Metallb 
+  install_metallb ""
```
stevenctl, line 160: slight nit: Do we need to pass `""`? Maybe we can make sure it's set in the `install_metallb` func with `if -z`.

stevenctl (approved):
This was tested as it's written now - if you want to merge that's fine just remove the hold.

gargnupur:
> This was tested as it's written now - if you want to merge that's fine just remove the hold.

Yeah I would leave it as is now.. and we can improve it after

stevenctl (changes requested):
This PR needs to be in istio/common-files

gargnupur:
> This PR needs to be in istio/common-files

Thanks for letting me know.. Added https://github.com/istio/common-files/pull/343

PR #3059, opened 2020-11-16
Fix mason for using release channel to create GKE clusters

chizhg:
Fix mason for using release channel to create GKE clusters: If release channel is specified, find the current valid versions in the release channel to determine the final version to use.

/cc @carolynhu

### DIFF
--- a/boskos/gcp/gke.go
+++ b/boskos/gcp/gke.go
@@ -130,11 +130,35 @@ func (cc *containerEngine) create(ctx context.Context, project string, config cl
 	if err != nil {
 		return nil, err
 	}
-	if config.Version == "" {
-		version = serverConfig.DefaultClusterVersion
+
+	// If release channel is specified, find the corresponding valid version in the release channel.
+	if config.ReleaseChannel != nil && config.ReleaseChannel.Channel != "" {
+		var chConfig *container.ReleaseChannelConfig
+		for _, ch := range serverConfig.Channels {
+			if ch.Channel == config.ReleaseChannel.Channel {
+				chConfig = ch
+			}
+		}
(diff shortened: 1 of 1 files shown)


boskos/gcp/gke.go
```diff
@@ -130,11 +130,35 @@ func (cc *containerEngine) create(ctx context.Context, project string, config cl
...
+	// If release channel is specified, find the corresponding valid version in the release channel.
+	if config.ReleaseChannel != nil && config.ReleaseChannel.Channel != "" {
+		var chConfig *container.ReleaseChannelConfig
+		for _, ch := range serverConfig.Channels {
+			if ch.Channel == config.ReleaseChannel.Channel {
+				chConfig = ch
+			}
+		}
+
+		if chConfig != nil {
```
carolynhu: does this logic cover all cases? what if chConfig is nil, we did nothing, `version` will be nil?
I think your line 154 `else`, should be not within an `else `block. Just put it at the function level

chizhg: If `chConfig` is nil, that means the channel name is not configured correctly, and we will get errors any way. If the `else` block is removed, `version` will always be overwritten to `serverConfig.DefaultClusterVersion` even when release channel is not empty, so it's logically incorrect.

Also `version` is a `string`, and it can never be `nil`.

carolynhu: then, could you log out the error or info message out?

chizhg: Good point. I'm now returning an error there directly if the release channel cannot be found.

carolynhu (approved):

PR #3070, opened 2020-11-19
Remove repos with only `main` from release branch common-file updates

ericvn:
Looking at all the jobs in: 

https://prow.istio.io/job-history/gs/istio-prow/logs/update-common_common-files_release-1.x_postsubmit,

they ALL show failed. The jobs in 

https://prow.istio.io/job-history/gs/istio-prow/logs/update-common_common-files_postsubmit

for the most part are successful.

I believe the problem is the fatal error in the logs like:
```
Cloning into 'src/istio.io/test-infra'...
warning: Could not find remote branch release-1.6 to clone.
fatal: Remote branch release-1.6 not found in upstream origin
```
caused by trying to checkout the non-existent branches in those repos that only have a `master/main` (test-infra, bots, cri)

This PR removes those repos from the automated job.

### DIFF
--- a/prow/cluster/jobs/istio/common-files/istio.common-files.master.gen.yaml
+++ b/prow/cluster/jobs/istio/common-files/istio.common-files.master.gen.yaml
@@ -49,6 +49,60 @@ postsubmits:
       org: istio
       path_alias: istio.io/test-infra
       repo: test-infra
+    name: update-common-mainonly_common-files_postsubmit
+    path_alias: istio.io/common-files
+    spec:
+      containers:
+      - command:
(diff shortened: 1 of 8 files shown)


ericvn:
I also split the main job into 3 jobs for the future:

- one that contains the repos not in master/main to copy at branch cut (these should always pass)
- one that only contains repos with master/main and disabled copying that at branch cut (prevent this problem going forward)
- one for istio.io  (limit failure to a single repo expected to fail from main branch cut until istio.io is cut at the release)

howardjohn (approved):

PR #3074, opened 2020-11-20
Fix proxy automator job and configgen improvement

### LINKED ISSUE #28876: update-proxy_envoy broken
https://prow.istio.io/view/gs/istio-prow/logs/update-proxy_envoy_periodic/1326675059149901824

It points to a test-infra SHA, not envoy

chizhg:
1. Fixes https://github.com/istio/istio/issues/28876
2. `path_alias` for periodic jobs does not have any meaning, since periodic Prow jobs do not necessarily need a repo, and the `path_alias` is already specified in `extra_refs`, so removing it from the job generator tool to avoid confusion.

/cc @howardjohn @carolynhu

### DIFF
--- a/prow/config/jobs/envoy.yaml
+++ b/prow/config/jobs/envoy.yaml
@@ -47,7 +47,7 @@ jobs:
   - --token-path=/etc/github-token/oauth
   - --cmd=scripts/update_envoy.sh $AUTOMATOR_SHA $AUTOMATOR_SHA_COMMIT_DATE
   requirements: [github]
-  repos: [istio/test-infra@master]
+  repos: [istio/envoy@master, istio/test-infra@master]
   image: gcr.io/istio-testing/build-tools:master-2020-11-12T22-29-05
   timeout: 4h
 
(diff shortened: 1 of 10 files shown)


prow/config/jobs/envoy.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
   - --token-path=/etc/github-token/oauth
   - --cmd=scripts/update_envoy.sh $AUTOMATOR_SHA $AUTOMATOR_SHA_COMMIT_DATE
   requirements: [github]
-  repos: [istio/test-infra@master]
+  repos: [istio/envoy@master, istio/test-infra@master]
```
carolynhu, line 50: i wonder what's the `repos` field here means? I thought it is to find the repo which has the automator.sh? If we put two repos here, what it will do? Could you explain.

chizhg: They will be generated as the `extra_refs` field in the actual Prow job config, see https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md#what-the-test-container-can-expect for more information. Specifically
```
`periodic` jobs will have the working directory set to the root of the repo specified 
by the first ref in `extra_refs` (if specified). 
See the `extra_refs` field if you need to clone more than one repo.
```

carolynhu: got it! that's might be why it checked out to test-infra SHA not envoy

carolynhu (approved):

PR #3092, opened 2020-12-09
Prow alert when images are stale for 14 days

chaodaiG:


prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
```diff
@@ -0,0 +1,24 @@
...
+            alert: 'Prow images are stale',
+            expr: |||
+              time()-max(prow_version) > %d * 24 * 3600
+            ||| % $._config.prowImageStaleByDays,
+            'for': '5m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              message: '@test-infra-oncall The prow images are older than %d days.' % $._config.prowImageStaleByDays,
```
cjwagner: Please remove the `@test-infra-oncall`. That is specific to k8s-prow not istio, is not specific to the alert, and only makes sense for a slack alerting sinks. We should include this kind of thing in the alert routing config (but probably not at all here since we don't have an oncall slack team for istio AFAIK).

cjwagner (commented):
This is missing the update to https://github.com/istio/test-infra/blob/master/prow/cluster/monitoring/mixins/prometheus/prometheus.libsonnet

prow/cluster/monitoring/mixins/prometheus/stale_alerts.libsonnet
chaodaiG: done

cjwagner:
This still needs to be addressed: https://github.com/istio/test-infra/pull/3092#pullrequestreview-548585391

chaodaiG:
Thanks, done

cjwagner (approved):

PR #3128, opened 2021-01-15
Refactor the configgen tool

chizhg:
This PR is a big refactor of the current config generator tool:
1. Make all the hardcoded fields configurable.
2. Support adding a `.global.yaml` config file in the folder level whose configuration fields are shared across all the meta config files under the same folder.
3. Add a matrix feature (enlightened by [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix)) to make it easier to configure Prow jobs that cover multiple dimensions.

Please note I didn't add unit tests for these changes as one of my teammates will be taking over the following work to upstream this config generator tool to k8s/test-infra, and he will be adding more test coverage once he gets there.


prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
```diff
@@ -34,9 +34,6 @@ postsubmits:
         - mountPath: /home/prow/go/pkg
           name: build-cache
           subPath: gomod
-        - mountPath: /gocache
```
howardjohn: is this expected to be removed?

chizhg: I think it's expected, there are two volumes mounts for the `build-cache` which I think are duplicated:
```
- mountPath: /home/prow/go/pkg
  name: build-cache
  subPath: gomod
- mountPath: /gocache
  name: build-cache
  subPath: gocache
```
, and only a few jobs are having both.

howardjohn: The two caches are different. /home/prow/go/pkg is go module pkg cache which is globally enabled since its strictly better (no need to `go mod download` every time). The /gocache is selectively enabled since it only works properly if the repo is setup to disable test caching where desired. For example, in istio/istio we want to the build cache (since compiling binaries is WAY faster), but we still want to run every test so we explicitly pass `-count 1`. Adding /gocache speed up some jobs in istio/istio 2x so I would like to keep it in some form or another.

Its confusing since its the same volume, but we are accessing differernt folders in that volume

chizhg: Thanks John, that is good to learn, I just saw we have `GOCACHE=/gocache` configured in the Docker image.
I have made the change accordingly, please take a look when you get time.

howardjohn (approved):

PR #3130, opened 2021-01-19
Add a flag to genjobs to allow disabling the limit set on the job name length

chizhg:
It doesn't really matter if the job name has more than 63 characters AFAICT, and truncating the job name sometimes makes it harder to parse, so I think it's better to remove this logic from the `genjobs` tool.

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -16,3 +16,4 @@ presubmits:
       - --strict
       - --exclude-warning=mismatched-tide
       - --exclude-warning=non-decorated-jobs
+      - --exclude-warning=long-job-names
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.5.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.5.gen.yaml
@@ -450,7 +450,7 @@ postsubmits:
       preset-enable-ssh: "true"
       preset-override-deps: release-1.5-istio
       preset-override-envoy: "true"
-    name: e2e-bookInfoTests-envoyv2-v1alpha3_istio_release-1.5_posts_priv
+    name: e2e-bookInfoTests-envoyv2-v1alpha3_istio_release-1.5_postsubmit_priv
     path_alias: istio.io/istio
     spec:
       containers:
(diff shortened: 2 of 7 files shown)


chaodaiG (approved):

howardjohn:
I think we added this due to failures from long job names. Have the limits been lifted?

cjwagner:
> Have the limits been lifted?

I don't think so. The limit exists to ensure the job name will fit in a kubernetes resource label. ([PR](https://github.com/kubernetes/test-infra/pull/9386))  I don't know if this label is used by Prow itself or if it is just to help with manually searching through ProwJobs and associated Pods.
If we confirm it is the latter we could make `genjobs` length limit optional, but I don't think we should omit the length limit by default.

chizhg:
> > Have the limits been lifted?
> 
> I don't think so. The limit exists to ensure the job name will fit in a kubernetes resource label. ([PR](https://github.com/kubernetes/test-infra/pull/9386)) I don't know if this label is used by Prow itself or if it is just to help with manually searching through ProwJobs and associated Pods.
> If we confirm it is the latter we could make `genjobs` length limit optional, but I don't think we should omit the length limit by default.

Thanks Cole. 
I didn't dig it deeper but Knative has had long-name Prow jobs for a while and we haven't seen any issues so far (e.g. https://github.com/knative/test-infra/blob/master/config/prod/prow/jobs/config.yaml#L10968 has 65 characters), so I assume it's the latter. 
And when you say making `genjobs` length limit optional, do you mean adding a flag to the tool to determine whether to enforce the limit or not?

cjwagner:
@chizhg I did some digging through the Prow code base and came to the same conclusion.

> And when you say making `genjobs` length limit optional, do you mean adding a flag to the tool to determine whether to enforce the limit or not?

Yes, exactly.

chizhg:
> @chizhg I did some digging through the Prow code base and came to the same conclusion.
> 
> > And when you say making `genjobs` length limit optional, do you mean adding a flag to the tool to determine whether to enforce the limit or not?
> 
> Yes, exactly.

Thanks for the clarification. Done.

cjwagner (approved):

howardjohn (approved):

Issue #3135, opened 2021-01-19
automator broken by upstream Kubernetes change


howardjohn:
https://github.com/kubernetes/test-infra/commit/37a45bba8ca1acbef74f635da3f0c5d4d34a477f#diff-e2bfedd79231a66d062709747fb5c7b22ce062aa4a139cf26f876676e7ee972b broke this: https://prow.istio.io/view/gs/istio-prow/logs/update-build-tools-image_common-files_postsubmit/1351654441589149696

We should probably pin the repo, but also update

PR #3148, opened 2021-01-22
Revert "Automator: update build-tools:release-1.9"

ericvn:
Reverts istio/test-infra#3126

The 01-14 build image has some newer Kubernetes tools which don't have some of the flags we use. Reverting back to an older build image. A new build-image will be made pinning the tools version and when that is merged, the files will have the latest (and working) build-tools version again.


ericvn:
See https://github.com/istio/test-infra/issues/3135 for info on failure.

howardjohn (approved):

PR #3169, opened 2021-02-01
Allow istio/proxy (releases) tests to use RBE cache and execution.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>

PR #3170, opened 2021-02-02
Do not use RBE for istio-private proxy builds

jacob-delgado:
Build jobs are failing as istio-private isn't setup to use RBE for proxy builds. Remove this capability, even though it'll be slower.


howardjohn (approved):
I am skeptical only .gen.yaml changed - seems like the source needs to change?

jacob-delgado:
Going to put a hold on this due to https://github.com/istio/test-infra/pull/3169

jacob-delgado:
> I am skeptical only .gen.yaml changed - seems like the source needs to change?

Modified the source, re-generated and cleaned up with make gen. Looks a lot better now. I'm waiting to see if we can get proxy builds going before we merge.

PR #3172, opened 2021-02-03
Istio prow monitoring: update based on all recent improments from k8s prow monitoring stack

chaodaiG:
Also:
- Added a stale image alert


chaodaiG:
bad base branch
/close

PR #3184, opened 2021-02-11
Add periodic job to update the istio reference used by istio.io tests

ericvn:


jacob-delgado (approved):

prow/config/jobs/istio.io.yaml
```diff
@@ -64,6 +64,21 @@ jobs:
     requirements: [github]
     repos: [istio/test-infra@master]
 
+  - name: update-istio-ref
+    disable_release_branching: true
+    types: [periodic]
+    cron: "0 2 * * 1"  # every Monday at 02:00AM UTC
```
jacob-delgado: only idea would be to do it on Saturday instead of Monday as fewer changes are going in over the weekend. Saturday is ideal as its Sunday for people in Asia-Pacific too.

ericvn: Changed to Sunday since it's 2 am UTC. This should be late Saturday to Sunday. I agree a time is better if few changes are going as the images to match the commit might not be built yet. Worst case is a retest on tests in the automated PR after the images are built.

johnma14 (approved):

PR #3195, opened 2021-02-23
Revert "Manual update build-tools:master"

ericvn:
Reverts istio/test-infra#3193

Is this causing some failures in the istio.io tests?


ericvn:
Let's cancel this PR and think about just reverting the image for istio.io if needed.

PR #3209, opened 2021-03-05
Add istio/client-go, istio/pkg and istio/gogo-genproto to list of automator jobs

jacob-delgado:
A concern I do have with doing this is the cycle that can happen when istio/common-files propagates causing a lot of Automator PRs.

### DIFF
--- a/prow/cluster/jobs/istio/client-go/istio.client-go.master.gen.yaml
+++ b/prow/cluster/jobs/istio/client-go/istio.client-go.master.gen.yaml
@@ -109,6 +109,61 @@ postsubmits:
           path: /tmp/prow/cache
           type: DirectoryOrCreate
         name: build-cache
+  - annotations:
+      testgrid-alert-email: istio-oncall@googlegroups.com
+      testgrid-dashboards: istio_client-go_postsubmit
+      testgrid-num-failures-to-alert: "1"
+    branches:
+    - ^master$
+    decorate: true
+    extra_refs:
+    - base_ref: master
+      org: istio
+      path_alias: istio.io/test-infra
+      repo: test-infra
+    name: update_client-go_dep_client-go_postsubmit
+    path_alias: istio.io/client-go
+    spec:
+      containers:
+      - command:
+        - ../test-infra/tools/automator/automator.sh
+        - --org=istio
+        - --repo=istio
+        - '--title=Automator: update istio/client-go@$AUTOMATOR_SRC_BRANCH dependency
+          in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH'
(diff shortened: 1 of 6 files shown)


jacob-delgado:
We should start out on `master` branch and maybe backport it to supported releases (1.8 and 1.9). Let's see how this evolves.

ericvn:
I agree that we might see a far number of extra PRs as things work across repos with a common-files update. Not sure if we can detangle enough to figure out a good way to handle that one. Have to draw out the changes to see if we could simplify those. Just adding a delay may not work as the intermediate jobs may have had a test failure and not merge until later.

howardjohn:
Can we somehow make automator skip common-file updates? We actually already have this problem, just to a smaller extent

howardjohn:
I mean it can update common-files, but if it updates common files then don't trigger the next repos in the chain

howardjohn:
Something like 
```
if pr name ~= "update common-files" {
  return
}
else {
  send update PR
}
```

ericvn:
I like @howardjohn's idea, and suggest a functional enhancement to automator.sh to allow a set of files/paths to skip (and maybe default to common/). Then check the git diff results to see if there are files outside the skipped changes. 

Actually, looks like git may allow a string to be specified that could be simply added in the script.

I tried  ` git diff -- . ":(exclude)common"` with a change in common and it seemed to work:
```

> git diff --exit-code  (returned 1)
diff --git a/common/.commonfiles.sha b/common/.commonfiles.sha
index 1ec2dc1071..fb16a2eb6e 100644
--- a/common/.commonfiles.sha
+++ b/common/.commonfiles.sha
@@ -1 +1 @@
-6fce4da9178ec1ade6fb361def33a0c4de8abb6b
+6fce4da9178ec1ade6fb361def33cca0c4de8abb6b

> git diff -- . ":(exclude)common" --exit-code     (returned 0) 
```

ericvn:
Only drawback to skipping the update on changes in the repos to `common` would be that someone running the `update_deps.sh` would see additional changes as it would grab the most recent commit which includes the `common` commits. Not sure if that is an issue or not.

If that is an issue, I say we can merge this PR as it stands. If it's not an issue, the we just update this PR to ignore changes in the `common` directory.

jacob-delgado:
I updated the PR to include skipping common. I think it's more important to make sure that the code that lives in `istio/pkg`, `istio/gogo-genproto` and `istio/client-go` get updated in `go.mod` in `istio/istio`.

From what I can tell though, not every change pushed to those repos is also met with a subsequent PR to istio/istio; that's why I made the PR in the first place. Release managers shouldn't have to update repos before branching, if they even do at all.

ericvn (approved):

PR #3213, opened 2021-03-11
Add labels to specify merge type for a repo

howardjohn:
Context: https://github.com/istio/envoy/pull/304#issuecomment-797069745
cc @PiotrSikora 

I *think* this means we stick that label on it and it will merge instead of squash

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -399,6 +399,9 @@ tide:
     istio-private: squash
   target_url: https://prow.istio.io/tide
   blocker_label: tide/merge-blocker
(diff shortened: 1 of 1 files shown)


PiotrSikora (approved):
Perfect if it works, thanks @howardjohn!

cjwagner (approved):
> I _think_ this means we stick that label on it and it will merge instead of squash

That's exactly right.
Please keep in mind that repos need to allow the alternate merge methods in their branch protection settings in order for this to work.

PR #3219, opened 2021-03-15
Added a prow job for automatic merging of upstream changes into 1.9 envoy release branch

dmitri-d:


dmitri-d:
The job itself is expected to fail, as there's a conflict in VERSION between 1.9 istio release branch and upstream 1.17

dmitri-d:
ping @PiotrSikora: is this what you had in mind?

prow/config/jobs/envoy-1.9.yaml
```diff
@@ -54,6 +54,23 @@ jobs:
...
 
+- name: update-envoy
+  types: [periodic]
+  interval: 24h
+  command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=envoy
+  - "--title=Automator: merge upstream changes to $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+  - --labels=auto-merge
```
PiotrSikora: Please also add `tide/merge-method-merge` label (see: https://github.com/istio/test-infra/pull/3213) and make sure the PR is merged, not squashed.

prow/config/jobs/envoy-1.9.yaml
```diff
@@ -54,6 +54,23 @@ jobs:
...
+  - --modifier=merge_upstream_changes
+  - --token-path=/etc/github-token/oauth
+  - --cmd=VERSION="v$(cut --fields=1,2 -d. VERSION)"; git remote add upstream https://github.com/envoyproxy/envoy.git && git fetch upstream release/"${VERSION}" && git merge upstream/release/"${VERSION}"
```
PiotrSikora, line 68: I have no idea what will happen afterwards. Is this merge commit going to be pushed as-is and used in a PR, or will the Automator try to create another commit or bail? I'm looking at [`automator.sh`](https://github.com/istio/test-infra/blob/master/tools/automator/automator.sh), and I don't see any support for this kind of workflow.

ericvn:
Is there a particular need to do a periodic update of Envoy in the actual release streams? In the past, this wasn't done, and I believe that the RMs would want to know about incoming Envoy changes. This might be a good topic for the T&R workgroup call.

PiotrSikora:
This pulls updates from Envoy's stable release branches (i.e. only security, reliability and build fixes). We already pull those as part of the sync during security release, but it's done manually and not in an automated fashion.

I think could remove `auto-merge` flag to let Release Managers review those changes before merging them, if you think that would be helpful.

prow/config/jobs/envoy-1.9.yaml
dmitri-d: This doesn't actually create a commit, just stages the changes for one; I'm not sure what will happen if there were conflicts during a merge; I'll take a closer look at automator.sh. Any chance we could test this on a dummy branch?

brian-avery:
I'm ok with this if we have the release manager/networking approval on PRs. I'm not sure what we would pull in with automated updates, but they can decide if it should be merged. I also think it's better to keep more in sync with Envoy which this would help with.

prow/config/jobs/envoy-1.9.yaml
dmitri-d: > make sure the PR is merged, not squashed.

Not sure how/where this is controlled, could you point me to it please?

dmitri-d:
* removed auto-merge label

PiotrSikora (approved):

brian-avery (approved):

ericvn:
This test continues to fail...

https://prow.istio.io/job-history/gs/istio-prow/logs/update-envoy_envoy_release-1.9_periodic

PR #3234, opened 2021-03-25
Fix testgrid presubmit failure

chaodaiG:


chaodaiG:
Unfortunately there is a chicken and egg problem, the change in this PR can only work together with job change, but job change apparently is also blocked by this test failure, so:

testgrid config checking job was triggered from job change in https://github.com/istio/test-infra/pull/3235, it has progressing much further now, I think it's fixed.

michelle192837 (approved):

ericvn (approved):

PR #3236, opened 2021-03-29
Switched to a plain string for the upstream branch name

dmitri-d:
This should resolve `fatal: Invalid refspec '+refs/heads/"release/$(cut --fields=1,2 -d. VERSION)":refs/remotes/upstream/"release/$(cut --fields=1,2 -d. VERSION)'` error (see here: https://storage.googleapis.com/istio-prow/logs/update-envoy_envoy_release-1.9_periodic/1376286882396114944/build-log.txt)

### DIFF
--- a/prow/cluster/jobs/istio/envoy/istio.envoy.release-1.9.gen.yaml
+++ b/prow/cluster/jobs/istio/envoy/istio.envoy.release-1.9.gen.yaml
@@ -29,7 +29,7 @@ periodics:
       - --modifier=merge_upstream_changes
       - --token-path=/etc/github-token/oauth
       - --merge-repository=https://github.com/envoyproxy/envoy.git
-      - --merge-branch="release/$(cut --fields=1,2 -d. VERSION)"
+      - --merge-branch="release/v1.17"
       image: gcr.io/istio-testing/build-tools:master-2021-03-01T22-30-49
       name: ""
       resources:
--- a/prow/config/jobs/envoy-1.9.yaml
+++ b/prow/config/jobs/envoy-1.9.yaml
@@ -66,7 +66,7 @@ jobs:
   - --modifier=merge_upstream_changes
   - --token-path=/etc/github-token/oauth
   - --merge-repository=https://github.com/envoyproxy/envoy.git
-  - --merge-branch="release/$(cut --fields=1,2 -d. VERSION)"
+  - --merge-branch="release/v1.17"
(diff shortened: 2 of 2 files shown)


dmitri-d:
ping @ericvn, @brian-avery

prow/cluster/jobs/istio/envoy/istio.envoy.release-1.9.gen.yaml
```diff
@@ -29,7 +29,7 @@ periodics:
       - --modifier=merge_upstream_changes
       - --token-path=/etc/github-token/oauth
       - --merge-repository=https://github.com/envoyproxy/envoy.git
-      - --merge-branch="release/$(cut --fields=1,2 -d. VERSION)"
+      - --merge-branch="release/v1.17"
```
brian-avery, line 32: This will introduce additional manual steps that will have to be done on branch cutting. The Envoy version doesn't match the Istio version, which we generally pull out and populate in https://github.com/istio/test-infra/blob/master/Makefile.core.mk for each branch. If we can't do that, I'd prefer having an automatic way to pull out the version like was present in the previous PR. 

I'm ok merging what is here right now as long as we can follow up and see if we can take care of this in some way that we don't expect someone to come in and make manual changes each release.

prow/config/jobs/envoy-1.9.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
   - --modifier=merge_upstream_changes
   - --token-path=/etc/github-token/oauth
   - --merge-repository=https://github.com/envoyproxy/envoy.git
-  - --merge-branch="release/$(cut --fields=1,2 -d. VERSION)"
+  - --merge-branch="release/v1.17"
```
brian-avery, line 69: This will introduce additional manual steps that will have to be done on branch cutting. The Envoy version doesn't match the Istio version, which we generally pull out and populate in https://github.com/istio/test-infra/blob/master/Makefile.core.mk for each branch. If we can't do that, I'd prefer having an automatic way to pull out the version like was present in the previous PR. 

I'm ok merging what is here right now as long as we can follow up and see if we can take care of this in some way that we don't expect someone to come in and make manual changes each release.

dmitri-d: sgtm

brian-avery (approved):

PR #3244, opened 2021-04-05
Istio prow: enable secret managed by kubernetes external secrets

chaodaiG:
This was enabled in upstream and working: https://github.com/kubernetes/test-infra/blob/master/prow/prow_secrets.md

### DIFF
--- a/prow/README.md
+++ b/prow/README.md
@@ -182,3 +182,10 @@ repos:
             contexts:
             - "merges-blocked-needs-admin"
 ```
+
+## Prow Secrets
+
+Some of the prow secrets are managed by kubernetes external secrets, which
+allows prow cluster creating secrets based on values from google secret manager
+(Not necessarily the same GCP project where prow is located). See more detailed
+instruction at [Prow Secret](https://github.com/kubernetes/test-infra/blob/master/prow/prow_secrets.md).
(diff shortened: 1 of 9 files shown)


howardjohn (commented):
It would be really nice to have gitops for Secrets as well, something like SealedSecrets. I don't suppose this allows us to do that?

chaodaiG:
> It would be really nice to have gitops for Secrets as well, something like SealedSecrets. I don't suppose this allows us to do that?

Afaict not yet, but I don't see what would be the blocker https://github.com/external-secrets/kubernetes-external-secrets

chaodaiG:
Realized that istio-prow cluster doesn't have workload identity enabled yet, would be good to enable before merging this PR.
/hold

@howardjohn , any objection enabling workload identity in prow cluster?

howardjohn:
Not particularly, unless there are some downsides I don't know of? I assume there should be no issues?

If it will evict pods from the node upgrade though, can we wait until night/weekend? We just cut release branch so we are at peak load on prow right now and don't want to cause issues

chaodaiG:
> Not particularly, unless there are some downsides I don't know of? I assume there should be no issues?
> 
> If it will evict pods from the node upgrade though, can we wait until night/weekend? We just cut release branch so we are at peak load on prow right now and don't want to cause issues

https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#limitations, no obvious downside that would affect prow afaict. And yes I believe it will evict pods, so I'll wait until tomorrow before enabling WI on prow

fejta (approved):

fejta:
We're using this in the other prow instances and it is pretty great.

Any idea when this will be ready for istio?

chaodaiG:
Workload identity was enabled. This PR is now ready to be merged @howardjohn

chaodaiG:
Added prow labels so these can be deployed. @fejta , may I have another stamp please?
/unhold

PR #3245, opened 2021-04-06
Add enhancements repo

brian-avery:


prow/config.yaml
```diff
@@ -218,6 +218,8 @@ branch-protection:
           branches: *blocked_branches
         client-go:
           branches: *blocked_branches
+        test-infra:
```
howardjohn: I am confused why this is not:
```
        enhancements:
          branches: *blocked_branches
```
?

howardjohn (approved):

prow/config.yaml
brian-avery: Fixed. Fat fingered.

chaodaiG (approved):

PR #3261, opened 2021-04-07
Update presets for 1.8 1.9 and 1.10

brian-avery:
This follows the pattern of https://github.com/istio/test-infra/pull/2939

PR #3262, opened 2021-04-07
Update presets for 1.8 1.9 and 1.10 

brian-avery:
Other half of #3261


brian-avery:
@Monkeyanator ... realized that I missed the other set of jobs here that need to be added for each release.

> Should this be documented as part of our branch cut process?

Yes, we should either document or automate adding this and #3261 as part of our branch cut process.

Monkeyanator (approved):

PR #3265, opened 2021-04-09
Add a postsubmit job for CNI

bianpengyuan:
https://github.com/istio/istio/issues/31765

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -265,7 +265,7 @@ jobs:
         # Skip revision test because CNI is not revisioned.
         # Skip traffic test because dns is not supported yet.
         # TODO: enable traffic test when https://github.com/istio/istio/issues/29511 is closed.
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -250,6 +250,23 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: " --istio.test.retries=1 "
 
+  - name: integ-cni-k8s-tests
```
brian-avery, line 253: @bianpengyuan can you make this optional and we can see if it passes consistently? https://github.com/bianpengyuan/test-infra/blob/cni-test/prow/config/jobs/istio.yaml#L21

brian-avery: Actually, nevermind. Talking to Eric, I think that the hidden argument makes it optional as well.

brian-avery (approved):

prow/config/jobs/istio.yaml
bianpengyuan: This is a post submit job. So I guess optional or not does not really matter?

PR #3269, opened 2021-04-12
Swap 1.21 with 1.20

howardjohn:
Goes with https://github.com/istio/istio/pull/32117


ericvn:
Just noticed this as well in my similar #3270. Not sure if we should also update the other images to later versions.

ericvn (approved):

PR #3278, opened 2021-04-13
Add prefixes private repos

brian-avery:
The [GOPRIVATE](https://medium.com/swlh/go-modules-with-private-git-repository-3940b6835727) flag allows us to override repos with private versions using a `GOPRIVATE` variable and a `gitproxy` file.

### DIFF
--- a/prow/cluster/jobs/all-presets.yaml
+++ b/prow/cluster/jobs/all-presets.yaml
@@ -133,6 +133,45 @@ presets:
       - key: secret
         path: id_rsa
         mode: 0600
+- labels: 
+    preset-use-api-private: "true"
+  volumeMounts:
+  - mountPath: $HOME/.gitconfig
+    name: gitconfig
+    readOnly: true
+  volumes:
+  - name: gitconfig
+    secret:
+      secretName: gitconfig
+  env:
+  - name: GOPRIVATE
+    value: istio.io/api
+- labels:
+    preset-use-go-control-plane-private: "true"
+  volumeMounts:
+  - mountPath: $HOME/.gitconfig
+    name: gitconfig
+    readOnly: true
+  volumes:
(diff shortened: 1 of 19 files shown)


prow/cluster/jobs/all-presets.yaml
```diff
@@ -133,6 +133,45 @@ presets:
...
+  - name: gitconfig
+    secret:
+      secretName: gitconfig
+  env:
+  - name: GOPRIVATE
+    value: istio.io/api
+- labels:
+    preset-use-go-control-plane-private: "true"
+  volumeMounts:
+  - mountPath: /etc/gitconfig
```
howardjohn: I thought it had to be in $HOME/.gitconfig?

prow/cluster/jobs/all-presets.yaml
```diff
@@ -133,6 +133,45 @@ presets:
...
+    name: gitconfig
+    readOnly: true
+  volumes:
+  - name: gitconfig
+    secret:
+      secretName: gitconfig
+  env:
+  - name: GOPRIVATE
+    value: github.com/envoyproxy/go-control-plane, istio.io/api
```
howardjohn: does space work?

```suggestion
    value: github.com/envoyproxy/go-control-plane,istio.io/api
```

prow/cluster/jobs/all-presets.yaml
brian-avery: $HOME/.gitconfig might be better. Pulled /etc/gitconfig from [docs](https://git-scm.com/docs/git-config#_configuration_file), but I don't actually see it in that location on my system. Didn't see $HOME used anywhere and wasn't sure if prow would evaluate it.

howardjohn (approved):

PR #3282, opened 2021-04-16
Add private jobs for 1.10

brian-avery:
This adds the private jobs for 1.10. In the future we should automate this.


brian-avery:
Should be automated in https://github.com/istio/istio/issues/32242

jacob-delgado (approved):

PR #3292, opened 2021-04-20
Add periodic job to for envoy-1.10 from envoy-1.9

ericvn:
This PR copies the envoy-1.9 periodic job to envoy-1.10. For now, the upstream `release-v1.18` branch doesn't exist so this shouldn't be merged yet.


brian-avery:
actually @ericvn I think this needs testgrid updates?

```
 2021/04/20 20:50:50 FAIL: could not apply prowjob annotations: couldn't find dashboard "istio_release-1.10_envoy_periodic" for job "update-envoy_envoy_release-1.10_periodic" 

```

brian-avery (approved):

ericvn:
I'm also not sure if we want to stay with this automation, or if we remove istio/envoy totally from the picture and point istio/proxy WORKSPACE to the latest commit in the relative envoyproxy/envoy branch. The release-builder logic would need to be updated for release-1.9 to pull the commit from upstream the same way that 1.10 currently does (at last look, I do know there was a 1.10 PR to move it back to istio/envoy commits).

The drawback to doing this would be Istio-private builds would still use a private repo forked from upstream (since private upstream PRs wouldn't be available).

ericvn:
We also lose the ability to add Istio patches outside of upstream and we could do now.

PR #3296, opened 2021-04-21
Disable RBE in 1.5/1.6 private

howardjohn:


jacob-delgado (approved):

prow/cluster/jobs/istio-private/proxy/istio-private.proxy.release-1.5.gen.yaml
```diff
@@ -22,7 +22,6 @@ postsubmits:
         - ./prow/proxy-postsubmit.sh
         env:
         - name: BAZEL_BUILD_RBE_INSTANCE
-          value: projects/istio-prow-build/instances/default_instance
```
jacob-delgado, line 25: you may have to set `value: null`. It's worth trying this first.

PR #3303, opened 2021-04-23
automator.sh now creates tickets in istio/istio on failed merges

dmitri-d:
* The tickets are assigned "area/networking/envoy" label
* istio/wg-networking-maintainers group is notified
* A new issue is not created if there's an open one already
* Failed automatic merges are not considered failures anymore.

ping @brian-avery, @ericvn.

### DIFF
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -258,8 +258,10 @@ merge() {
   local code=$?
   if [ "$code" -ne 0 ]; then
     echo "$token" | gh auth login --with-token
-    local issue_exists=$(gh issue list -S "Automatic merge of $merge_branch into $branch failed." -R "istio/isio" | wc -l)
-    if [ $issue_exists -eq 0 ]; then
+    export GITHUB_TOKEN="$token"
+    local issue_exists
+    issue_exists=$(gh issue list -S "Automatic merge of $merge_branch into $branch failed." -R "istio/isio" | wc -l)
+    if [ "$issue_exists" -eq 0 ]; then
       gh issue create -b "Automatic merge of $merge_branch into $branch failed. @istio/wg-networking-maintainers" -t "Automatic merge of upstream envoy release branch failed" -l "area/networking/envoy" -R "istio/istio"
       print_error "Conflicts detected, manual merge is required. An issue in istio/istio has been created." 0
     else
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -257,7 +257,14 @@ merge() {
   git -c "user.name=$user" -c "user.email=$email" merge --no-ff -m "$title" --log upstream/"$merge_branch"
   local code=$?
   if [ "$code" -ne 0 ]; then
-    print_error "$(git status)" 1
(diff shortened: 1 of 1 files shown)


tools/automator/automator.sh
```diff
@@ -257,7 +257,15 @@ merge() {
   git -c "user.name=$user" -c "user.email=$email" merge --no-ff -m "$title" --log upstream/"$merge_branch"
   local code=$?
   if [ "$code" -ne 0 ]; then
-    print_error "$(git status)" 1
+    export GITHUB_TOKEN="$token"
+    local issue_exists
+    issue_exists=$(gh issue list -S "Automatic merge of $merge_branch into $branch failed." -R "istio/isio" | wc -l)
```
brian-avery, line 262: will this list only open issues? I.e.:
*  if it fails once
* opens an issue
* someone fixes it and it continues
* another merge failure occurs -- what happens here?
If I understand correctly from the gh docs, then this will find the original issue and not open a new one?

tools/automator/automator.sh
```diff
@@ -257,7 +257,15 @@ merge() {
...
+    if [ "$issue_exists" -eq 0 ]; then
+      gh issue create -b "Automatic merge of $merge_branch into $branch failed. @istio/wg-networking-maintainers" -t "Automatic merge of upstream envoy release branch failed" -l "area/networking/envoy" -R "istio/istio"
+      print_error "Conflicts detected, manual merge is required. An issue in istio/istio has been created." 0
+    else
+      print_error "Conflicts detected, manual merge is required. An issue in istio/istio already exists." 0
```
brian-avery, line 267: is there any point to this else? I.e. is it cleaner to do: 

```python
if [ "$issue_exists" -eq 0 ]; then
      gh issue create -b "Automatic merge of $merge_branch into $branch failed. @istio/wg-networking-maintainers" -t "Automatic merge of upstream envoy release branch failed" -l "area/networking/envoy" -R "istio/istio"
fi
print_error "Conflicts detected, manual merge is required. An issue in istio/istio has been created." 0
```

Seems to be the same log message in both if and else cases

tools/automator/automator.sh
dmitri-d: `list` shows open issues only; my assumption is that these issues will be closed as soon as they have been resolved; the worst case is that we won't see new issues while there's one open.

tools/automator/automator.sh
dmitri-d: The other message is slightly different -- it mentions that an issue already exists. I don't feel strongly about it though, can remove it if you don't think it's useful.

brian-avery: Thanks :) Misread that earlier.

tools/automator/automator.sh
brian-avery: Thanks. Didn't try testing it, but the docs made it look like they would show anything.

brian-avery: I think it's safe to trust that the issues will be closed once fixed. Otherwise it's hard to differentiate.

brian-avery (approved):

PR #3305, opened 2021-04-27
Don't fail automator's merge workflow if there's a merge conflict

dmitri-d:


dmitri-d:
ping @brian-avery

brian-avery:
Code changes look good to me, and this flow is new specifically for Envoy, but line 226 hard codes networking labels and assignee. I think this is ok for now as it's only currently used for Envoy, but this is a general script used throughout Istio.  We may want to consider adding arguments for issue-labels and issue-comment that get passed in via prow's job config.

brian-avery (approved):

PR #3307, opened 2021-04-30
 Authentikos: support using real token expiration time instead of hard coded 3600s

chaodaiG:
Authentikos produces http cookiefile with template provided by user, which allows hard coded expiration time, which is not always the same as the token expiration time, this PR generates expiration time based on token expiration time.

### DIFF
--- a/authentikos/authentikos.go
+++ b/authentikos/authentikos.go
@@ -77,7 +77,7 @@ const (
 var timeNow = time.Now
 
 // tokenCreator is a function that creates an oauth token.
-type tokenCreator func(forceRefresh bool, tries int) ([]byte, error)
+type tokenCreator func(forceRefresh bool, tries int) ([]byte, *time.Time, error)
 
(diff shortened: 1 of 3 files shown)


chaodaiG:
I don't understand the unit test failure:
```
--- FAIL: TestValidateFlags (0.00s)
    --- FAIL: TestValidateFlags/error:_template-file_unreadable (0.00s)
        authentikos_test.go:147: expected error: true != actual error: false: <nil>
```

This PR doesn't touch the flags parsing logic at all, why does this test fail

/retest

chaodaiG:
ok, master branch presubmit is broken for authentikos https://github.com/istio/test-infra/pull/3308

chaodaiG:
ok, the tests are broken in master branch #3309

chaodaiG:
Close for now, will work on it after https://github.com/istio/test-infra/issues/3309 is resolved
/close

PR #3310, opened 2021-05-03
temporary jobs for istiodless remotes

stevenctl:
These jobs are likely broken (tested networking in https://github.com/istio/istio/pull/32601 and it fails for an unknown reason)

We should have a quick way to trigger these, but eventually we should change our main mc jobs to use this mode.

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -76,6 +76,24 @@ jobs:
       - name: TEST_SELECT
         value: "-multicluster"
 
+  - name: integ-telemetry-istiodless-mc-k8s-tests
+    modifiers:
+      - optional
+      - hidden
+      - skipped
+    command:
+      - entrypoint
+      - prow/integ-suite-kind.sh
+      - --topology
+      - MULTICLUSTER
+      - test.integration.telemetry.kube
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -76,6 +76,24 @@ jobs:
...
+    command:
+      - entrypoint
+      - prow/integ-suite-kind.sh
+      - --topology
+      - MULTICLUSTER
+      - test.integration.telemetry.kube
+    requirements: [kind]
+    env:
+      - name: TEST_SELECT
+        value: "-multicluster"
```
nmittler, line 93: I thought we were trying to get rid of the multicluster tests? Should we run networking tests instead?

stevenctl: This flag removes the multicluster label selector which targets the multicluster suite. The _last_ thing in that suite is the external istiod stuff, which we can hopefully put into our greater topology.

stevenctl: The added jobs are dupes of `pilot-` `telemetry-` and `security-`

nmittler (approved):

PR #3332, opened 2021-05-18
Fix arg

brian-avery:
The extension on the schema should be `.json` instead of `.yaml`.


johnma14 (approved):
Makes sense now :). Thanks @brian-avery

brian-avery:
You're welcome. Thank you.

PR #3335, opened 2021-05-19
Istio prow: switch grafana password to be from external secret

chaodaiG:
The password was generated from openssl and already in istio-testing project


prow/cluster/kubernetes_external_secrets.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: kubernetes-client.io/v1
+kind: ExternalSecret
+metadata:
+  name: grafana
```
cjwagner: This is not the correct secret name: https://github.com/istio/test-infra/blob/cc140e1076e8ee23f651a99d124e7b025e72fa05/prow/cluster/monitoring/grafana_deployment.yaml#L44

cjwagner (changes requested):

prow/cluster/kubernetes_external_secrets.yaml
chaodaiG: fixed

cjwagner (approved):

Issue #3309, opened 2021-04-30
Authentikos presubmit test is broken


chaodaiG:
No op change in authentikos is blocked by `unit-test-authentikos_test-infra` and `integ-test-authentikos_test-infra`, see https://github.com/istio/test-infra/pull/3308

clarketm:
Fixed by:
#3342 / integ
#3370 / unit

PR #3342, opened 2021-05-22
fix authentikos integ test

clarketm:
Fixes towards https://github.com/istio/test-infra/issues/3309. AFACT, this will not resolve the issue with the secret not being created: 

`"Error from server (NotFound): secrets "authentikos-token" not found"` ([e.g.](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_test-infra/3342/integ-test-authentikos_test-infra/1396039731891933184/build-log.txt)). 

I presume the cause is some state change in the cluster no longer working with the test setup/execution. Possibly an issue with `GOOGLE_APPLICATION_CREDENTIALS`?

https://github.com/istio/test-infra/blob/784193f780e93c8ac32d3b8e05cced0af0350073/authentikos/test/integ-simple.sh#L66

I added a log line after the test concludes for more visibility and debugability.

### DIFF
--- a/authentikos/test/integ-simple.sh
+++ b/authentikos/test/integ-simple.sh
@@ -38,26 +38,26 @@ run_test() {
   local tokeninfo="$1"
 
   echo "Test 'error' is null"
-  echo "$tokeninfo" | jq -r '.error' | xargs -r -I% test % = "null"
+  test "$(echo "$tokeninfo" | jq -r '.error')" = "null"
 
   echo "Test 'issued_to' exists"
-  echo "$tokeninfo" | jq -r '.issued_to' | xargs -r test -n
+  test -n "$(echo "$tokeninfo" | jq -r '.issued_to')"
 
   echo "Test 'audience' exists"
-  echo "$tokeninfo" | jq -r '.audience' | xargs -r test -n
+  test -n "$(echo "$tokeninfo" | jq -r '.audience')"
 
   echo "Test 'user_id' exists"
-  echo "$tokeninfo" | jq -r '.user_id' | xargs -r test -n
+  test -n "$(echo "$tokeninfo" | jq -r '.user_id')"
 
   echo "Test 'email' exists"
-  echo "$tokeninfo" | jq -r '.email' | xargs -r test -n
+  test -n "$(echo "$tokeninfo" | jq -r '.email')"
 
   echo "Test 'expires_in' greater than 0"
-  echo "$tokeninfo" | jq -r '.expires_in' | xargs -r -I% test % -lt 0
+  test "$(echo "$tokeninfo" | jq -r '.expires_in')" -gt 0
 
   echo "Test 'scope' includes 'userinfo.email', 'cloud-platform', and 'openid'"
   echo "$tokeninfo" | jq -r '.scope' |
-    grep -w "openi" |
+    grep -w "openid" |
     grep -w "https://www.googleapis.com/auth/cloud-platform" |
     grep -w "https://www.googleapis.com/auth/userinfo.email" >/dev/null
 }
@@ -68,6 +68,7 @@ main() {
   kubectl wait deployment authentikos --for="condition=available" --timeout="$timout"
 
   run_test "$(with_timeout get_tokeninfo "$timout")"
+  kubectl logs -l app=authentikos
 }
 
 main
--- a/authentikos/test/integ-simple.sh
+++ b/authentikos/test/integ-simple.sh
@@ -32,10 +32,13 @@ get_tokeninfo() {
   curl -sSfL "https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=$token"
 }
 
-run_test() {
-  set -x
+run_test() { (
+  set -ex
 
   local tokeninfo="$1"
+
+  echo "Test 'tokeninfo' response body exists"
+  test -n "$tokeninfo"
 
   echo "Test 'error' is null"
   test "$(echo "$tokeninfo" | jq -r '.error')" = "null"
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

authentikos/test/integ-simple.sh
```diff
@@ -68,6 +68,7 @@ main() {
   kubectl wait deployment authentikos --for="condition=available" --timeout="$timout"
 
   run_test "$(with_timeout get_tokeninfo "$timout")"
+  kubectl logs -l app=authentikos
```
cjwagner: Currently if `run_test` fails one of its checks the script exits immediately. I think we can do something like the following to instead capture the error code to return after the log dump.
```suggestion
  local code=0
  run_test "$(with_timeout get_tokeninfo "$timout")" || code="$?"
  kubectl logs -l app=authentikos
  exit "${code}"
```

cjwagner (commented):
Hi Travis, hope you are well! Thanks for the PR.

cjwagner:
It looks like the service account is provided by `preset-service-account` and is istio-prow-test-job@istio-testing.iam.gserviceaccount.com. I checked the details for the SA in the cloud console and confirmed that the SA keys are still active so AFAICT `GOOGLE_APPLICATION_CREDENTIALS` is set up correctly.

authentikos/test/integ-simple.sh
clarketm: I am good Cole! I hope you are doing well also! I like the suggestion. Let me make the update.

clarketm: Updated with minor tweaks to the syntax to get the desired behavior. PTAL

authentikos/test/integ-simple.sh
```diff
@@ -32,42 +32,49 @@ get_tokeninfo() {
...
-}
+); }
 
 main() {
   kubectl create secret generic service-account --from-file="service-account.json=$GOOGLE_APPLICATION_CREDENTIALS"
   kubectl apply --filename="$ROOT/testdata/authentikos-simple.yaml"
   kubectl wait deployment authentikos --for="condition=available" --timeout="$timout"
 
   run_test "$(with_timeout get_tokeninfo "$timout")"
+  exit_code="$?"
```
cjwagner: We're not reaching this line because the previous line fails and exit on error is enabled.  I spent some time messing around with the behavior of `set -e` and it is pretty confusing when trying to handle errors that occur in nested functions so I think my previous advice was bad.
Now I think `trap`ping the log dump is the easiest way to ensure we do so after tests complete or fail.

clarketm: I ended up wrapping the function body in a subshell. This will allow the function to exit on error but not terminate the parent shell [[ref](https://stackoverflow.com/a/58964551/4279440)]. Let me know what you think. Alternatively, I can add a trap like you suggest; I will just need to make sure the trap that logs runs before the other trap that deletes the kind cluster in `lib.sh`.

https://github.com/istio/test-infra/blob/a21aba25a5e916fc738673a60713c73362b5e962/authentikos/test/integ-simple.sh#L35

cjwagner: Hmm that does seem like it should work, but I still don't see any logs from Authentikos. Perhaps Authentikos has no logs? (I'm trying to debug locally, but I'm having problems getting a kind cluster up on my local machine.)

clarketm: The logging is not the best but it exists. Make sure the `--verbose` flag is set on the Authentikos deployment (which reminds me this should probably default to `true`). When I run locally using a personal SA I get the following logs. If an error occurs, it will log it after `reconcile complete` line.

https://github.com/istio/test-infra/blob/d8909f110a2da2d95ba1c9f495456b001ac4ae69/authentikos/test/testdata/authentikos-simple.yaml#L69

```
2021/05/26 22:01:31 expired: false; token expiry (minus delta): 2021-05-26 22:56:30.520797031 +0000 UTC m=+3309.168705502; next reconcile: 2021-05-26 22:31:31.520799684 +0000 UTC m=+1810.168708140
2021/05/26 22:01:31 creating secret: authentikos-token in namespace: default
2021/05/26 22:01:31 2021-05-26T22:01:31Z: reconcile complete; secrets: 1; errors: 0; next reconcile: 30m 
```

cjwagner: I started my own test PR to mess around with this in CI. I'm not seeing logs from Authentikos, but it also seems like the subshell is not preventing an early exit as expected: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_test-infra/3368/integ-test-authentikos_test-infra/1400242374365417472
(Note that there is no echoed "Dumping logs before exit".)

clarketm: Thanks for debugging. Let me verify this locally, but I think I just need to unset the `errexit` option (i.e. `set +e`)  in the parent shell to prevent the premature exit. I did something similar in automator to handle this same situation (see below).

https://github.com/istio/test-infra/blob/ce16c4297dc86fe77e85ac9b3ce7fa02b1b50018/tools/automator/automator.sh#L331-L337

clarketm: @cjwagner – I made the modifications mentioned above and verified locally that the Authentikos logs are emitted on failure.

cjwagner: Cool, now we're getting some logs:
> 2021/06/03 02:38:58 2021-06-03T02:38:58Z: reconcile complete; secrets: 0; errors: 1; next reconcile: 30m
2021/06/03 02:38:58 errors: default:maximum tries: 5 exceeded to force refresh token

Unfortunately it doesn't look like we're logging the relevant error messages. I made a PR to add some more logging: https://github.com/istio/test-infra/pull/3370

cjwagner:
Now that we should have more logs:
/retest

cjwagner:
Interesting, this looks like some sort of networking problem? 
> 2021/06/04 17:35:54 Failed to get oauth token from client: oauth2: cannot fetch token: Post https://accounts.google.com/o/oauth2/token: dial tcp: lookup accounts.google.com on 10.96.0.10:53: read udp 10.244.0.3:45866->10.96.0.10:53: read: connection refused.

I wonder if this has something to do with how kind is running in CI. I'm not sure how best to debug this. I'll ask for some help.

cjwagner (approved):
/retest

PR #3358, opened 2021-05-27
Revert "Disable go cache"

howardjohn:
Reverts istio/test-infra#3357

Things are still broken, they are just slow *and* broken now. Lets revert to make them fast at least


howardjohn:
will make a new one

PR #3370, opened 2021-06-03
Make Authentikos log token creation error messages.

cjwagner:
/cc @chaodaiG @clarketm

### DIFF
--- a/authentikos/authentikos.go
+++ b/authentikos/authentikos.go
@@ -306,16 +306,19 @@ func getOauthTokenCreator(o options) (tokenCreator, error) {
 
 		client, err := clientCreator(forceRefresh)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cjwagner:
Hmm this unit test error seems unrelated to my change: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_test-infra/3370/unit-test-authentikos_test-infra/1400504203780034560

cjwagner:
That unit test passes locally by the way. This is another example of Authentikos tests behaving differently in CI than they do locally.

cjwagner:
Oh, this unit test is failing in CI because it runs as root which would make the file readable even with 0000 permissions. It is unclear to me how this ever passed in CI, but I've confirmed that running as root causes the test case to fail.
I'm going to just drop this unit test, it doesn't seem particularly important.

clarketm (approved):

howardjohn (approved):

PR #3373, opened 2021-06-04
Postsubmit prow deploy job automatically sends slack notifications

chaodaiG:
This was pretty much a copy of https://github.com/kubernetes/test-infra/blob/aee86026c87da353a8c5b14f5b3fe80bcb820212/config/jobs/kubernetes/test-infra/test-infra-trusted.yaml#L102, which had been working pretty well for k8s/test-infra

based on https://testgrid.k8s.io/istio_test-infra_postsubmit#deploy-prow_test-infra_postsubmit, this shouldn't happen very frequently, for now it's only once per week


prow/config/jobs/test-infra.yaml
```diff
@@ -34,6 +34,15 @@ jobs:
   - name: deploy-prow
     types: [postsubmit]
     regex: '^prow/cluster/(?:gcsweb/|velodrome/|monitoring/|build/|private/|[^/]+\.yaml$)'
+    reporter_config:
+      slack:
+        channel: "oncall"
```
howardjohn, line 39: doesn't every pustsubmit already report to slack? is this just so it goes to oncall instead of test-failures?

howardjohn (approved):
approved with hold since I had a question, feel free to merge it though

prow/config/jobs/test-infra.yaml
chaodaiG: yeah they report on faillures and the reports are mixed with all other jobs https://github.com/istio/test-infra/blob/c0a883877ac28436cf3b2baaebd9b474d8123197/prow/config.yaml#L109. This is for replacing the manual post from oncall, so that we have one less thing to do on Mondays

PR #3386, opened 2021-06-16
Update to use first commit's parent sha to find file differences

### LINKED ISSUE #31851: Release notes tooling can fail if a PR needs to be rebased
**Bug description**
In the [release notes tooling](https://github.com/istio/tools/blob/master/cmd/gen-release-notes/main.go#L246), we check to see if a release note has been added/modified in this PR. This works fine most of the time and lets us track whether release notes have been documented. 

We do this through a git diff command

```
git diff-tree -r --diff-filter=AMR --name-only --relative=releasenotes/notes '4f001b70f61e41df9d3f068a590b1a9c9f9217af' 'd4f88fae62c5b2ae6c12f4e9949b68b46ba5f7e8'
```
The problem here happens if a release note is removed between when a branch is started and it's merged back in: 

1. PR branch is created from master
2. Release note gets removed in another PR that gets merged into master
3. PR gets opened from this PR branch
4. Git command above gets run. It looks for files added/modified/removed from this PR
5. When it compares the currently existing release note (removed in step 2) against the version in master, it finds that the file wasn't added/removed/modified, but also doesn't exist. It errors and the release notes tooling fails.  
6. A rebase of the PR branch should fix this, but it would be nice to not have this problem to begin with.

ericvn:
Fixes istio/istio#31851

Currently we have two problems with the release notes checker:
1. We validate files with aren't in the PR. They are in the commits between there the PR's branch was made and the current PR. This can cause failures in validating files not in the PR: see https://github.com/istio/istio/pull/33296 which was failing on `revision-tag-command.yaml` which was not originally the PR.
2. We fail to fail when a PR which might have no release notes and no label. See: https://github.com/istio/istio/pull/33426. Here we find additional files (just like in 1) so the checker thinks there is a release notes file, but in reality there is not. 

How to fix? Instead of looking for files from the current commit back to the commit where the branch was made, just look for file changes from the current commit back to the parent of the current PR's first commits.

PULL_BASE_SHA is no longer used, instead using the current PR's first commit's parent commit.

### DIFF
--- a/tools/check_release_notes.sh
+++ b/tools/check_release_notes.sh
@@ -72,7 +72,7 @@ get_opts() {
             shift 2
             ;;
         --base-sha)
-            PULL_BASE_SHA="$2"
+            base_sha="$2"
             shift 2
             ;;
         --pr-head-sha)
@@ -122,11 +122,6 @@ validate_opts() {
         exit 1
     fi
 
-    if [ -z "${PULL_BASE_SHA:-}" ]; then
-        echo "PULL_BASE_SHA not specified. This must match the base SHA for the pull request."
-        exit 1
-    fi
-
     if [ -z "${REPO_PATH:-}" ]; then
         echo "REPO_PATH not specified. Using current working directory."
         REPO_PATH=$(pwd)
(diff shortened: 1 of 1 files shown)


ericvn:
Marking DNM while @brian-avery has time to review.

ericvn:
Test output using the PR for failure #2 which now fails:
```
../test-infra/tools/check_release_notes.sh 
REPO_PATH not specified. Using current working directory.
base-sha not specified. Calculating from the PULL_NUMBER
base_sha: e0ef3dee766c011a4bb84e34ebcceb5407457eaa
~/work/go/src/istio.io/tools/cmd/schema-validator ~/work/go/src/istio.io/istio
~/work/go/src/istio.io/istio
Checking files from pull request istio/istio#33426 head SHA: 29aefcddb703b67755204089026e8c144e8b8f59 destination branch: , base SHA: e0ef3dee766c011a4bb84e34ebcceb5407457eaa
~/work/go/src/istio.io/istio ~/work/go/src/istio.io/istio
~/work/go/src/istio.io/tools/cmd/gen-release-notes ~/work/go/src/istio.io/istio ~/work/go/src/istio.io/istio
~/work/go/src/istio.io/istio ~/work/go/src/istio.io/istio
Looking for release notes in ./releasenotes/notes.
Executing: cd ./releasenotes/notes; git diff-tree -r --diff-filter=AMR --name-only --relative=releasenotes/notes 'e0ef3dee766c011a4bb84e34ebcceb5407457eaa' '29aefcddb703b67755204089026e8c144e8b8f59'
Found 0 files.

Parsing release notes
failed to find any release notes.
No release notes files found in '/releasenotes/notes/'.

~/work/go/src/istio.io/istio
Requesting labels from pull request istio/istio#33426
Missing "release-notes-none" label

Missing release notes and missing "release-notes-none" label.
If this pull request contains user facing changes, please create a release note based on the template: https://github.com/istio/istio/blob/master/releasenotes/template.yaml by going here: https://github.com/S-Chan/istio/new/air-set-workloadentry-locality-from-bootstrap-node/releasenotes/notes
Release notes documentation can be found here: https://github.com/istio/istio/tree/master/releasenotes
If this pull request has no user facing changes, please add the release-notes-none label to the pull request. Note that the test will have to be manually retriggered after adding the label.
```

brian-avery:
Fixes https://github.com/istio/istio/issues/31851

brian-avery (commented):
LGTM but lets wait until tomorrow to merge.

brian-avery (approved):

PR #3387, opened 2021-06-17
Update build-tools and build-tools-centos images.

PiotrSikora:

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -1,5 +1,5 @@
 istio_container_with_kind: &istio_container_with_kind
-  image: gcr.io/istio-testing/build-tools:master-2021-06-15T16-37-30
+  image: gcr.io/istio-testing/build-tools:master-2021-06-16T23-16-54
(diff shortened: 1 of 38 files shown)


howardjohn (commented):
sorry about the lack of docs on this, but I think we want https://github.com/istio/common-files/pull/434. Once that is merged there is some automation to generate an equivalent PR to this.

I would approve this as is since its already ready but I think the build-tools-centos is not updated here which is what we care about anyways?

PiotrSikora:
> sorry about the lack of docs on this, but I think we want [istio/common-files#434](https://github.com/istio/common-files/pull/434). Once that is merged there is some automation to generate an equivalent PR to this.

Yeah, I've noticed previous commit by the automator, but it wasn't triggered by a postsubmit in istio/tools, so I figured I'd speed it up... but we can wait for the automator to do its job if you prefer.

> I would approve this as is since its already ready but I think the build-tools-centos is not updated here which is what we care about anyways?

Oops, great catch! Fixed.

howardjohn (approved):

PR #3399, opened 2021-06-21
Explicitly define slack report config in each job

chaodaiG:
Since https://github.com/kubernetes/test-infra/pull/22486, prow has "fixed" the usage of `skip_report` to only work for reporting back to CI systems. This PR:
- removed global slack configs
- define slack report in individual jobs instead

Recommended files to review:
- prow/config.yaml
- prow/config/generate.go
- prow/config/istio-private_jobs/.defaults.yaml
- prow/config/jobs/.global.yaml

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -178,6 +178,8 @@ postsubmits:
       preset-override-envoy: "true"
     name: integ-telemetry-istiodless-mc-k8s-tests_istio_postsubmit_priv
     path_alias: istio.io/istio
+    reporter_config:
+      slack: {}
     skip_report: true
     spec:
       containers:
@@ -511,6 +513,8 @@ postsubmits:
       preset-override-envoy: "true"
     name: integ-pilot-istiodless-multicluster-tests_istio_postsubmit_priv
     path_alias: istio.io/istio
+    reporter_config:
+      slack: {}
     skip_report: true
     spec:
       containers:
@@ -710,6 +714,8 @@ postsubmits:
       preset-override-envoy: "true"
     name: integ-security-istiodless-multicluster-tests_istio_postsubmit_priv
     path_alias: istio.io/istio
+    reporter_config:
+      slack: {}
     skip_report: true
     spec:
       containers:
@@ -1854,6 +1860,8 @@ presubmits:
     name: integ-telemetry-istiodless-mc-k8s-tests_istio_priv
     optional: true
     path_alias: istio.io/istio
+    reporter_config:
+      slack: {}
     skip_report: true
     spec:
       containers:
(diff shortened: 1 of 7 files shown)


prow/cluster/jobs/istio-private/api/istio-private.api.master.gen.yaml
```diff
@@ -13,6 +13,14 @@ postsubmits:
       preset-override-envoy: "true"
     name: build_api_postsubmit_priv
     path_alias: istio.io/api
+    reporter_config:
+      slack:
+        channel: test-failures
```
howardjohn: private jobs should be in `#private-test-failures`

prow/cluster/jobs/istio/api/istio.api.master.gen.yaml
```diff
@@ -172,6 +204,14 @@ presubmits:
     decorate: true
     name: gencheck_api
     path_alias: istio.io/api
+    reporter_config:
```
howardjohn: I think this is a presubmit test? So it should not trigger slack

cjwagner:
> Since [kubernetes/test-infra#22486](https://github.com/kubernetes/test-infra/pull/22486), prow has "fixed" the usage of `skip_report` to only work for reporting back to CI systems. This PR:
> 
> * removed global slack configs
> * define slack report in individual jobs instead

I don't understand why the referenced change would preclude the use of the global slack configs and require slack reporting config to be specified on every job. My understanding is that the only effect the referenced change would have would be to make `skip_report: true` jobs start reporting to slack. If that is the case I would just expect those (presumably less common) `skip_report: true` jobs also specify an empty slack reporting config to opt out of slack reporting.

chaodaiG:
> > Since [kubernetes/test-infra#22486](https://github.com/kubernetes/test-infra/pull/22486), prow has "fixed" the usage of `skip_report` to only work for reporting back to CI systems. This PR:
> > 
> > * removed global slack configs
> > * define slack report in individual jobs instead
> 
> I don't understand why the referenced change would preclude the use of the global slack configs and require slack reporting config to be specified on every job. My understanding is that the only effect the referenced change would have would be to make `skip_report: true` jobs start reporting to slack. If that is the case I would just expect those (presumably less common) `skip_report: true` jobs also specify an empty slack reporting config to opt out of slack reporting.

Good point. Crier doesn't support this yet, see https://github.com/kubernetes/test-infra/blob/1dd3c44deb7433d620b57d901ae9118c0abcd181/prow/crier/reporters/slack/reporter.go#L149, empty slack config in prowjob has no effect. There will be upstream changes required

prow/cluster/jobs/istio/api/istio.api.master.gen.yaml
chaodaiG: good catch
/hold

chaodaiG:
/hold
This change will not work until https://github.com/kubernetes/test-infra/pull/22693 is in istio prow

chaodaiG:
Latest prow bump https://github.com/istio/test-infra/pull/3412 includes https://github.com/kubernetes/test-infra/pull/22693, so this PR should work now
/cc @cjwagner

howardjohn (approved):

PR #3402, opened 2021-06-25
Add postsubmit job for security fuzz tests

yangminzhu:


howardjohn (commented):
IIRC in the design doc it was specifically mentioned that this would not run routinely as part of each merged commit. I am fairly worried about the maintainability of this job over the long term, as it is bringing on a large number of additional dependencies.

It would be good to get more eyes from @istio/wg-test-and-release-maintainers before proceeding.

yangminzhu:
> IIRC in the design doc it was specifically mentioned that this would not run routinely as part of each merged commit. I am fairly worried about the maintainability of this job over the long term, as it is bringing on a large number of additional dependencies.
> 
> It would be good to get more eyes from @istio/wg-test-and-release-maintainers before proceeding.

I think in the design doc we will want to get this running periodically instead of running in the presubmit. If postsubmit is still considered too much, How about I change it to run once everyday? 

Regarding the maintenance, this is in a separate test target, I think the worst case is we disable it if we later find out that we do not have enough time to fix any potential flakes in it?

yangminzhu:
replaced by https://github.com/istio/test-infra/pull/3409

PR #3405, opened 2021-06-29
Update build tools image to 06-28.

bianpengyuan:
Update build tools image to include https://github.com/istio/tools/pull/1589. Not sure why automation does not update.


ericvn (changes requested):
The change is not initially made here, but would be made here: https://github.com/istio/common-files/blob/master/files/common/scripts/setup_env.sh#L66.

One that change is merged, automation will create PRs to update the files in test-infra as well in all the repos.

That said, we do have an issue to actually make the change when a new image is built (but that can cause issues since some like to have the image to test before pushing it to the total pipeline).

bianpengyuan:
@ericvn thanks for the pointer! Created https://github.com/istio/common-files/pull/439 for this.

PR #3409, opened 2021-06-30
Add periodic job for security fuzz tests

yangminzhu:
@howardjohn I added the job by following existing configurations with some small changes as this only needs the istio/istio (not istio/tools) repo, I'm not sure what's the best way to test such configurations so hope it will just work.


prow/cluster/jobs/all-periodics.yaml
```diff
@@ -398,6 +398,32 @@ periodics:
     nodeSelector:
       testing: test-pool
 
+# Security policy fuzz testing
+- cron: "0 7 * * *"  # starts every day at 07:00AM UTC
```
howardjohn: I think the all-periodics is sort of legacy file. We should probably clean them up long term. For this PR, it would be good to use the new way at least.
An example can be found in the  update-istio-ref job in prow/config/jobs/istio.io.yaml.

yangminzhu: thanks for the review, updated.

howardjohn (approved):

PR #3421, opened 2021-07-07
Generation of istio-private jobs. 

brian-avery:
This: 
* Implements branching for istio-private jobs using a new utility called `generate-transform-jobs`. It may be useful in the future to add diff and print functionality to this utility as well to bring our transformation job tooling in line with the prow job tooling. 
* Renames istio-private job files for consistency as well as updating of master files.
* Fixes an issue where `generate.go` doesn't check for errors
* Fixes an issue where `generate.go` hard codes paths
* Fixes an issue where we clean up public prow cluster configuration when running `make gen` but ignore private prow cluster configuration

@jacob-delgado @ericvn Please take a look. 

Part of https://github.com/istio/istio/issues/32242. @Monkeyanator I think you may be aware of other jobs that aren't automated?


prow/config/cmd/generate.go
```diff
@@ -109,6 +128,52 @@ func main() {
 		}); err != nil {
 			exit(err, "walking through the meta config files failed")
 		}
+
+		// istio-private
```
chizhg: AFAICT the tool in `prow/config/cmd` is supposed to be a generic tool for generating general Prow jobs, and the tool in `prow/genjobs/` is used for transforming the generated Prow jobs to use for private repos. So can we have this change in `prow/genjobs/` instead?

ericvn:
I'm wondering if you just change `branch` to create the new files in prow/config/istio-private_jobs. You'll still need to update the files there to support branching on a release as you have here. Then the `make gen` in a follow-on step does the `go run prow/genjobs/main.go --configs=./prow/config/istio-private_jobs` to create the new Prow jobs.

prow/config/cmd/generate.go
brian-avery: Thanks @chizhg. 

- `genjobs` is primarily for combining public jobs with private jobs (transforms) in order to produce cluster configs for istio-private jobs. 
- `prow/config/cmd/generate.go write` generates the public cluster config based on public cluster jobs. 
- `prow/config/cmd/generate.go branch` generates jobs for a new Istio branch, currently limited to only public jobs. 

This PR adds the ability to create private jobs for new Istio branches. After these are created, `make gen` can be run, which executes `genjobs` and `prow/config/cmd/generate.go write` in order to produce the cluster configs for both public and private jobs.

prow/config/cmd/generate.go
```diff
@@ -44,10 +46,112 @@ func GetFileName(repo string, org string, branch string) string {
...
+
+			for key, val := range transform.Labels {
+				transform.Labels[key] = strings.Replace(val, "master", branch, 1)
+			}
+
+			jobs.Transforms[key] = transform
+
+		}
+
+		dst := path.Join("..", "istio-private_jobs", name)
```
chizhg: It looks you could just use the `privateInputDir` flag value instead of hardcoding it here. The tool is now generic and is used by other projects using Prow, so it's better not adding Istio specific logic.

prow/config/cmd/generate.go
chizhg: Thanks @brian-avery .

It's still unclear to me why you need to update this tool and also import `genjobs` as a library... Is it going to generate the config files in `prow/config/istio-private_jobs`? If yes I feel it should be another tool to do this.

prow/config/cmd/generate.go
brian-avery: +1. Apologies, this is still a work in progress. Istio testing just removed the label earlier.

prow/config/cmd/generate.go
brian-avery: Yeah, this is to generate the config files in `prow/config/istio-private_jobs`. I'm fine creating separate tools to generate public and private config. Regardless, the tool will have to pull in the Configuration struct as a library as `genjobs` should probably stay for translating the cluster config. It just means we will have to call two tools in release builder when we create the config.

prow/config/cmd/generate.go
brian-avery: I wonder if we should make the jobs version of generate.go use the `inputDir` as well. This same value is currently hard-coded in master.

prow/generate-transform-jobs/main_test.go
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2019 Istio Authors
```
chizhg: Nit: the header is inconsistent with others.

chizhg:
Do you also need to run `generate-transform-jobs` in the `make gen` target?

brian-avery:
> Do you also need to run `generate-transform-jobs` in the `make gen` target?

No, `make gen` is to create prow cluster configs. That's taken care of by `genconfig`. This implements the same functionality that `generate.go branch` implements for normal prow job configs but for transformation configs.

prow/generate-transform-jobs/main_test.go
brian-avery: Thanks. Let me fix that one. At some point, we should probably go through the rest of this repo and fix the other copyrights similar to https://github.com/istio/istio/pull/23958.

prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
```diff
@@ -107,6 +109,53 @@ postsubmits:
     cluster: private
     decorate: true
     labels:
+      preset-enable-gomod-netrc: "true"
+      preset-enable-ssh: "true"
+      preset-override-deps: master-istio
+      preset-override-envoy: "true"
+      preset-service-account: "true"
+    name: benchmark-report_istio_postsubmit_priv
```
howardjohn: This doesn't seem like it should be added

howardjohn (commented):
LGTM aside from one

prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
brian-avery: We're currently including that in 

1.9: https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.9.gen.yaml#L117

and 1.10:https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio-private/istio/istio-private.istio.release-1.10.gen.yaml#L117

howardjohn (approved):

ericvn (approved):
LGTM

brian-avery:
@istio/operations-maintainers Can you please give this a stamp of approval for creating the `generate-transform-jobs` utility? Test and Release will own it (and the codeowners changes are in this PR), but because it's created in the prow directory, it requires your approval. Thanks!

cjwagner (approved):
Approving based on https://github.com/istio/test-infra/pull/3421#issuecomment-877233446

PR #3427, opened 2021-07-08
Update automator exclude check to exclude files in PR causing event

ericvn:
Currently, a common-files update will cause 16 automated PRs to be generated. Some of the PRs end up in a race condition and need some help to finish. This is caused by a recent change as more automation was added to try and limit an exposure of not having the latest commits when building a release.

The 16 automated jobs:
- There are automator jobs from common-files to cause PRs directly in istio.io, istio, api, tools, release-builder, pkg, client-go, gogo-genproto, and proxy. (9 jobs)
- The updates in api, pkg, client-go, gogo-genproto and proxy will cause automated PRs in istio to pick up their new commits (with only the common-files change). (5 jobs)
- The initial update in api, will also push an update to client-go and that will cause yet another update to istio. (2 jobs)

If we can prevent some of the updates from happening, namely those in api, pkg, client-go, gogogen-proto and proxy caused by the common-files updates, we remove 7 of the automated PRs.

What is the drawback to not doing this? The istio repo will be pointing to older commits (but only ignoring commits with common files updates) of the 5 repos. The istio `update_deps.sh` script will update to the latest commits. This means we are still seeing commits in istio from the 5 repos for non common-files changes, so still ahead of where we used to be (no automation in one repos). If we added a pre-submit job in `release-builder` to test for changes from an update_deps run, we could get back to making sure all repo references are up to date and limit the number of automated jobs using this PR.

Thoughts?

Note that I am replacing the behavior with the --git-exclude parameter with new behavior as it was added recently and only used in a few cases.

Slack discussion: https://istio.slack.com/archives/C6FCV6WN4/p1625688978099700

### DIFF
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -118,7 +118,7 @@ get_opts() {
       shift
       ;;
     --git-exclude)
-      git_exclude=(":^$2")
+      git_exclude="$2"
       shift 2
       ;;
     --)
@@ -194,10 +194,6 @@ validate_opts() {
   if [ -z "${email:-}" ] && ! $dry_run; then
     email="$(curl -sSfLH "Authorization: token $token" "https://api.github.com/user" | jq --raw-output ".email")"
   fi
-
-  if [ -z "${git_exclude:-}" ]; then
-    git_exclude=()
-  fi
 }
 
 evaluate_opts() {
@@ -237,7 +233,7 @@ add_labels() {
 
 commit() {
   if $dry_run; then
-    git diff --cached "${git_exclude[@]}"
+    git diff --cached
     return 0
   fi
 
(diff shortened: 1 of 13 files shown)


tools/automator/automator.sh
```diff
@@ -280,6 +276,18 @@ merge() {
   fi
 }
 
+validate_changes_exist_in_prior_pr() {
+  if [ -n "$git_exclude" ]; then
+    pr="$(curl -sSfLH "Authorization: token $token" -H "Accept: application/vnd.github.groot-preview+json" "https://api.github.com/repos/$user/$repo/commits/$sha/pulls" | jq ".[0].number")"
+    export GITHUB_TOKEN="$token"
+    changes=$(gh pr view "$pr" --repo "$user"/"$repo" --json files | jq -r '.files[].path' | grep -cvE '$git_exclude')
```
howardjohn: Why do we need to look at the PR instead of the commit locally? Aren't they the same?

howardjohn:
LGTM overall, just one thing I didn't fully understand about the implementation

tools/automator/automator.sh
ericvn: If we always expect the PR to be squashed, then we should be able to look at the last commit. I recall some issues with `git diff-tree` in the past.  I can try just using `git show --name-only`.

ericvn: @howardjohn Changed implementation to use files for the most recent commit. Format it name only with a one-liner as the header removed with the sed. Then the grep will remove the specified lines. If any changes are left, the job continues. Else ends.

tools/automator/automator.sh
```diff
@@ -280,6 +280,16 @@ merge() {
   fi
 }
 
+validate_changes_exist_in_prior_pr() {
```
brian-avery: nit: Should we make a similar change here to address the commit instead of PR?

brian-avery:
This LGTM, but will let @howardjohn provide the approval.

tools/automator/automator.sh
brian-avery: Would it make more sense to call this `validate_changes_exist_in_latest_commit` or something like that? The use of prior here made me think we're looking back at history, which we aren't really.

ericvn: Updated the function name and added a description.

brian-avery: Thanks :)

ericvn:
@howardjohn Any other comments on this?

howardjohn (approved):

PR #3460, opened 2021-07-22
Add optional job for new istiodremote tests.

ericvn:
This defines new optional jobs which test the changes in https://github.com/istio/istio/pull/34262 and it's cherry-pick to release-11.

Once the tests are passing, will remove the optional here and remove the original tests in istio/istio tested by the multicluster job.


stevenctl (approved):
Could we start this off with `skip` so it doesn't trigger on every job

frankbu:
Old central istiod test is already removed in https://github.com/istio/istio/pull/34331 because it doesn't work anymore with the framework changes. So it would be good to start running these new tests as soon as possible after 34331 merges, since otherwise we'll have no integration testing of external control plane at all.

ericvn:
Merge the optional tests so they can be run manually in the PR with the tests: https://github.com/istio/istio/pull/34331

PR #3486, opened 2021-08-05
run pilot and security tests with Mixed protocol LB service feature gate turned on

su225:
Follow up of https://github.com/istio/test-infra/pull/3481
Required for https://github.com/istio/istio/pull/33817
I had missed turning it on for presubmit and some postsubmit. Doing it in this PR

Signed-off-by: su225 <suchithjn22@gmail.com>


howardjohn (commented):
Maybe it makes sense in istio/istio to just always use this if the k8s version is new enough? Then we can easily add it to other jobs (say when we want to test QUIC telemetry), etc.

su225:
@howardjohn - Done. Changed it.

howardjohn (approved):

PR #3487, opened 2021-08-05
Un hide,skip,optional the istiodremote tests

ericvn:
Now that the tests are merged.


stevenctl (approved):

stevenctl:
Can we start with `optional` so that we can see that its consistently not-flaky across many runs without requiring `/retest` on a lot of PRs.

PR #3499, opened 2021-08-10
Use writable directory for cache

howardjohn:
Recommended by GKE team to use this one; /prow is not writable so will
not work


jacob-delgado (approved):

chizhg:
This also broke the internal Prow...
FYI @jasonwzm

PR #3506, opened 2021-08-13
Add job for externalistiod pilot tests

frankbu:


prow/config/jobs/istio.yaml
```diff
@@ -176,6 +176,21 @@ jobs:
...
+    command:
+      - entrypoint
+      - prow/integ-suite-kind.sh
+      - --topology
+      - MULTICLUSTER
+      - --topology-config
+      - prow/config/topology/external-istiod.json
+      - test.integration.pilot.kube
+    requirements: [kind]
+    modifiers: [presubmit_optional]
```
stevenctl: Should also start with skipped, or this will trigger for every PR

stevenctl (approved):

PR #3511, opened 2021-08-17
Remove multicluster label

frankbu:
It looks like @su225 sent a PR to remove the `multicluster` labels at the same time I sent a PR to add the new istiodremote job, which also had the label.

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -171,13 +171,11 @@ jobs:
       - prow/config/topology/external-istiod.json
       - test.integration.pilot.kube
     requirements: [kind]
+    resources: multicluster
     modifiers:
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -171,13 +171,11 @@ jobs:
       - prow/config/topology/external-istiod.json
       - test.integration.pilot.kube
     requirements: [kind]
+    resources: multicluster
```
frankbu, line 174: ^^^ @stevenctl, is this needed? What is it for? I noticed it's set for all the other mc tests.

howardjohn: Looks like it just increases CPU

howardjohn (approved):

PR #3517, opened 2021-08-24
Initial steps to replace `prow-bot`, `prow-deployer`, `istio-prow-test-job`, and `boskos` SA keys with Workload Identity.

cjwagner:
The Prow components using the SA key in the `prow-service-account` secret should be able to rely on the already configured workload identity. I updated the IAM rules so we can just remove those mounts entirely.

After this merges I'll bind the SAs and make another PR to start using them.

/assign @chaodaiG


chaodaiG (approved):
/hold
unhold when ready

PR #3518, opened 2021-08-24
Run periodic update jobs to update envoyproxy/envoy v1.19 branch

jacob-delgado:


Makefile.core.mk
```diff
@@ -31,6 +31,7 @@ generate-config:
 	@(cd prow/config/cmd; go run generate.go write)
 	@rm -fr prow/cluster/jobs/istio-private/*/*.gen.yaml
 	@go run prow/genjobs/main.go --configs=./prow/config/istio-private_jobs
+	@go run prow/genjobs/main.go --configs=./prow/config/istio-private_jobs
```
jwendell: why duplicate this?

prow/config/jobs/envoy-1.11.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: update-proxy
+  types: [periodic]
+  interval: 24h
+  command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=proxy
+  - "--title=Automator: update envoy@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+  - --labels=auto-merge
```
jwendell, line 49: Not sure if we want automatic stuff being merged in a micro release...

jwendell (changes requested):

Makefile.core.mk
jacob-delgado: Fixed. I was going to remove the `istio-private.envoy` file. Decided not to remove it.

prow/config/jobs/envoy-1.11.yaml
jacob-delgado: We do for 1.10. The @istio/wg-networking-maintainers group wanted Istio to no longer have to maintain istio/envoy.

ericvn:
Slightly different way in https://github.com/istio/test-infra/pull/3521 where I moved the job to proxy instead of using envoy. Not sure which way is better.

jacob-delgado:
Closed in lieu of Eric's PR

PR #3520, opened 2021-08-25
Continue migrating service account keys to workload identity.

cjwagner:
Follow up to https://github.com/istio/test-infra/pull/3517
I've already bound the `prowjob-default-sa` and `prow-deployer` service accounts.
Commits:
- Fixes an issue with the build cluster service accounts (they were pruned before they were applied due to a bad label value). And switches test-infra-trusted cluster to WI.
- Makes jobs using `prow-deployer` credentials switch to workload identity and stop using config generation (doesn't support `serviceAccountName` field and doesn't seem necessary here).
- Cleans up unused `deploy` genjobs requirements and `preset-prow-deployer-service-account`.

/assign @chaodaiG


chaodaiG:
please fix merge conflict

chaodaiG (approved):

cjwagner:
Fixed another merge conflict.

chaodaiG (approved):

prow/config/jobs/test-infra.yaml
```diff
@@ -31,62 +31,6 @@ jobs:
...
-    command:
-    - make
-    - -C
-    - boskos
-    - deploy
-    requirements: [deploy]
-    node_selector:
-        prod: prow
-
-  - name: deploy-boskos-config
```
howardjohn, line 76: Why did we move it? This may limit our auto-image updater, etc.

if there are limitations in the job generation we should address those IMO. If its just serviceAccountName missing we will probably need that anyways for our other (istio dev team owned) jobs currently using SA keys

fejta (approved):
/hold

prow/config/jobs/test-infra.yaml
cjwagner: I moved these jobs because they are infrastructure jobs that should be managed by our oncall team. We're not really familiar with Istio's job generator so I think it makes more sense for these jobs to be defined directly in Prow config like we're used to. Even if the job generation is updated to support serviceAccountName, I'd prefer to avoid the extra generation step; I don't think it is really adding value for these 3 jobs.

> This may limit our auto-image updater, etc.

Could you elaborate on this concern please? I wouldn't expect this to change to cause problems with auto image bumping, directly defined jobs should also be bumped. If that is not the case then that seems like the area to fix to me.

PR #3521, opened 2021-08-25
Fix periodic proxy update jobs

ericvn:
I noticed the periodic check to update the 1.11 release stream to the new envoy changes from yesterday. There is no `envoy-1.11` job, and I don't know that one is really needed since we don't have a envoy release-branch.  So, I moved the job from the envoy job to the proxy job, and then updated the proxy-1.11 job to include it as well. 

@istio/release-managers-1-11


jwendell (commented):
dupe of https://github.com/istio/test-infra/pull/3518 ?

ericvn:
Yep, although in a different way. I didn't want to add a envoy job for 1.11 since we don't have that branch.

ericvn:
Also, not sure if we should move this back to 1.10 as well, or just leave that release alone.

prow/config/jobs/proxy-1.11.yaml
```diff
@@ -124,6 +124,25 @@ jobs:
...
   - postsubmit
+- command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=proxy
+  - '--title=Automator: update envoy@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH'
+  - --labels=auto-merge
+  - --modifier=update_envoy_dep
+  - --token-path=/etc/github-token/oauth
+  - --cmd=scripts/update_envoy.sh $AUTOMATOR_SHA $AUTOMATOR_SHA_COMMIT_DATE
```
jwendell:
```suggestion
  - --cmd=UPDATE_BRANCH=release/v1.19 scripts/update_envoy.sh
```

prow/config/jobs/proxy.yaml
```diff
@@ -73,6 +73,23 @@ jobs:
...
+  interval: 24h
+  command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=proxy
+  - "--title=Automator: update envoy@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+  - --labels=auto-merge
+  - --modifier=update_envoy_dep
+  - --token-path=/etc/github-token/oauth
+  - --cmd=scripts/update_envoy.sh $AUTOMATOR_SHA $AUTOMATOR_SHA_COMMIT_DATE
```
jwendell:
```suggestion
  - --cmd=UPDATE_BRANCH=main scripts/update_envoy.sh
```

prow/config/jobs/proxy-1.11.yaml
```diff
@@ -124,6 +124,25 @@ jobs:
   timeout: 4h0m0s
   types:
   - postsubmit
+- command:
+  - ../test-infra/tools/automator/automator.sh
+  - --org=istio
+  - --repo=proxy
+  - '--title=Automator: update envoy@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH'
+  - --labels=auto-merge
```
jwendell, line 132: I'm not sure about this auto-merge thing. Do we really want to always use the latest stuff from envoyproxy/envoy - even if this comes from a "stable" branch? What do network maintainers think about this practice?

prow/config/jobs/proxy-1.11.yaml
ericvn: Fixed.

prow/config/jobs/proxy.yaml
ericvn: The script defaults to `main` if UPDATE_BRANCH is not defined. I'll add it here as I believe that will get copied to the new proxy at the branch cut and be a placeholder. Since we haven't change to main yet, 4Q item, the RMs may still miss changing the main to the actual envoy branch.

prow/config/jobs/proxy-1.11.yaml
ericvn: I'll defer to the @istio/wg-networking-maintainers-data-plane, but I thought that was what was decided.

jacob-delgado: I won't speak for them, but we wouldn't keep istio/envoy up to date with the latest patches for a branch. That made updating envoy difficult at times (not all bug fixes would get rolled up) and there's no longer the need for istio/envoy outside of security patches. Envoy is pretty strict on patches getting merged into older releases so I'm confident in using the head of the release branch personally.

ericvn: @istio/wg-networking-maintainers-data-plane or @istio/wg-product-security-maintainers Thoughts on whether or not we auto-merge? Otherwise I think this is ready.

prow/config/jobs/envoy.yaml
```diff
@@ -38,23 +38,6 @@ jobs:
     value: "/home/prow/go/src/istio.io/envoy"
   command: [./ci/do_ci.sh, bazel.release]
 
-- name: update-proxy
```
brian-avery, line 41: Does generate.go rely on this? Do we need to do something fancy so it gets added to to the branches but not run on master?

ericvn: This job was just moved from the envoy yamls to being run on the proxy yamls. ~~I do see I forgot to remove the one periodic from the generated envoy-10 so new commit coming for that.~~  Actually, I decided to not push the movement of the job from envoy to proxy in the earlier releases (1.10 and earlier) and just leave them the same. The envoy to proxy is just happening in master and 1.11 and will continue with 1.12 (assuming nothing changes).

The generator.go should copy the job from the proxy master to the new proxy-{release} yamls.

The one missing thing for the instructions that generate.go will use the `main` envoyproxy branch when the new release branches are cut. The UPDATE_BRANCH will need to be updated in the yaml, or it can be updated in the istio/proxy/scripts/update_envoy.sh from `main` to the correct branch once it's released.

brian-avery (approved):

PR #3523, opened 2021-08-25
Add automation to update common-files for a new build image

### LINKED ISSUE #2469: Add automator job for updating build tools tag in istio/common-files
Add automator job for updating build tools tag in [`istio/common-files`](https://github.com/istio/common-files/) (i.e. https://github.com/istio/common-files/pull/202)

While a **postsubmit** would be ideal, without strong pipeline support in Prow this is needlessly complex. There will be a race between the image build/push (i.e. `containers_tools_postsubmit`) and the availability of the image (also, the job could fail). Furthermore, the image tag is *dynamic* and computed as a timestamp at runtime. 

As an alternative, a daily **periodic** to query *gcr.io* for latest image tag for the branch (i.e. `$branch-latest`), then grab the corresponding timestamped tag for that image (e.g. `$branch-$date`) can be employed.

Another alternative is to use a tekton pipeline and combine this with the `containers_tools_postsubmit` although support in Prow is limited.

**e.g.**
```shell
$ gcloud container images list-tags gcr.io/istio-testing/build-tools  \
--filter="tags:master-latest" \
| grep -Pom1 'master-[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}-[0-9]{2}-[0-9]{2}'
```

ericvn:
Requires https://github.com/istio/common-files/pull/464. Once that PR and this PR merges, it will fix: https://github.com/istio/test-infra/issues/2469.

I'm not sure of the appropriate way to combine a job in the current repo and running automator against another repo.

In this case, I'm doing the existing post submit to create the build-image and then in the automator sh running in the common-files I grep the build output from the tools directory and call the script.

Another option would be to simply call the automator script in the common-files repo, which cd's back to tools and does the `make` and then runs the script back in the common-files directory. This keeps the total command in the automator --cmd.

### DIFF
--- a/prow/config/jobs/tools.yaml
+++ b/prow/config/jobs/tools.yaml
@@ -37,7 +37,7 @@ jobs:
     - --labels=release-notes-none
     - --modifier=update_tools_dep
     - --token-path=/etc/github-token/oauth
-    - --cmd=pushd ../tools && make containers-test |& tee make.out && popd && ./bin/update_build_image.sh --image $(grep "+ VERSION" ../tools/make.out | sed -e 's/.*=//')
+    - --cmd=pushd ../tools && make containers |& tee make.out && popd && ./bin/update_build_image.sh --image $(grep "+ VERSION" ../tools/make.out | sed -e 's/.*=//')
     resources: build
     regex: 'docker/.+|cmd/.+'
     requirements: [gcp, docker]
--- a/prow/config/jobs/tools.yaml
+++ b/prow/config/jobs/tools.yaml
@@ -30,12 +30,6 @@ jobs:
   - name: containers
     types: [postsubmit]
     command:
-    - make
-    - containers
-    - "|&"
-    - tee
-    - make.out
-    - "&&"
     - ../test-infra/tools/automator/automator.sh
     - --org=istio
     - --repo=common-files
@@ -43,7 +37,7 @@ jobs:
     - --labels=release-notes-none
     - --modifier=update_tools_dep
     - --token-path=/etc/github-token/oauth
-    - --cmd=bin/update_build_image.sh --image $(grep "+ VERSION" ../tools/make.out | sed -e 's/.*=//')
+    - --cmd=pushd ../tools && make containers-test |& tee make.out && popd && ./bin/update_build_image.sh --image $(grep "+ VERSION" ../tools/make.out | sed -e 's/.*=//')
     resources: build
     regex: 'docker/.+|cmd/.+'
     requirements: [gcp, docker]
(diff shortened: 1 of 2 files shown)


prow/config/jobs/tools.yaml
```diff
@@ -29,7 +29,21 @@ jobs:
 
   - name: containers
     types: [postsubmit]
-    command: [entrypoint, make, containers]
+    command:
+    - make
+    - containers
+    - "|&"
```
howardjohn: does this really work? isn't it a bash/shell thing, but this would pass it to `make`?

ericvn: I had tried it locally in a shell, `make containers-test |& tee make.out`. The idea here is to do the make and capture all the output to the `make.out` file as well as output it for the logs. Then within the automator command I find the version from the output. 

Note, that this all could be simplified by running the `make` and then the automator script defaulting it to use the latest found version.

howardjohn: I think `command` doesn't run in a shell though. Its actually more like `make containers-test '|&' tee make.out`?

ericvn: I think you are saying that I can create a new script, which will run in bash, or I can do as I mentioned in option#2 in the header and move the make inside of the automator script cmd (which is effectively copied to a .sh and run).

I'm speculating that the way prow works is that since this is a `tools` job that the tools repo will be cloned as well as the `test-infra` repo (the reason the script starts with ../test-infra. The automator script will also clone the specified `common files` repo so all 3 are available. The automator script will be running in the `common-files` directory so can cd and do the make capturing the output and then popd back and run the other script.

Let me work on that.

ericvn: This command does work when run from the command line (assuming my earlier script is available).

prow/config/jobs/tools.yaml
```diff
@@ -29,10 +29,18 @@ jobs:
 
   - name: containers
     types: [postsubmit]
-    command: [entrypoint, make, containers]
+    command:
+    - ../test-infra/tools/automator/automator.sh
```
howardjohn, line 33: I think you need `repos` clause to get test-infra checked out in the job

prow/config/jobs/tools.yaml
```diff
@@ -29,10 +29,18 @@ jobs:
...
+    - --org=istio
+    - --repo=common-files
+    - "--title=Automator: update build-tools image@$AUTOMATOR_SRC_BRANCH in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+    - --labels=release-notes-none
+    - --modifier=update_tools_dep
+    - --token-path=/etc/github-token/oauth
+    - --cmd=pushd ../tools && make containers |& tee make.out && popd && ./bin/update_build_image.sh --image $(grep "+ VERSION" ../tools/make.out | sed -e 's/.*=//')
```
howardjohn: is there some way we can make this a script or something? I would be impressed if all the shell/yaml/etc quoting all works out - it might be fine but certainly complex

ericvn: Fixed. Requires the script in PR: https://github.com/istio/common-files/pull/465

jacob-delgado (approved):

PR #3541, opened 2021-09-03
A few updates and improvements to Prow configgen tool

chizhg:
1.  Change `.global.yaml` to `.base.yaml`, and allow each sub-folder to overwrite the config which'll be shared by all config files under the same folder.

2. Improve the generated `trigger` field to match both the default trigger and the new configured trigger.

3. Allow `cron` and `interval` to be configured in the base config.


prow/config/README.md
```diff
@@ -78,6 +78,9 @@ requirement_presets:
       preset-service-account: "true"
 ```
 
+In each sub-folder, a `.base.yaml` file can also be added which'll overwrite the
```
nmittler: Need to clarify a bit more. Is it an overlay on top of the root base yaml, or does it completely replace it? Also, does this work recursively? For example:

```
./base.yaml (1)
./dir1/base.yaml (2)
./dir1/dir2/base.yaml (3)
```

For files under `dir1/dir2`, is the base just `dir1/dir2/base.yaml`? Or is it `./base.yaml` overlayed by `./dir1/base.yaml` overlayed by `./dir1/dir2/base.yaml`?

prow/config/cmd/generate.go
```diff
@@ -62,50 +63,61 @@ func main() {
...
 
-				name := file.Name()
-				ext := filepath.Ext(name)
-				name = name[:len(name)-len(ext)] + "-" + flag.Arg(1) + ext
+				src := filepath.Join(path, file.Name())
+				jobs := cli.ReadJobsConfig(src)
+				jobs.Jobs = config.FilterReleaseBranchingJobs(jobs.Jobs)
+
+				if jobs.SupportReleaseBranching {
+					tagRegex := regexp.MustCompile(`^(.+):(.+)-([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}-[0-9]{2}-[0-9]{2})$`)
```
nmittler: make this a constant?

prow/config/cmd/generate.go
```diff
@@ -62,50 +63,61 @@ func main() {
...
+					match := tagRegex.FindStringSubmatch(jobs.Image)
+					branch := "release-" + flag.Arg(1)
+					if len(match) == 4 {
```
nmittler, line 104: Can you add some docs to clarify the logic here a bit? It also might be useful to have a utility break some of this down into clear method names that aid in the reader's understanding.

prow/config/README.md
chizhg: It is an overlay on top of the root base yaml. Currently the overlay logic only applies to the `.base.yaml` files in the current folder and root folder since that'll work for us for now, but in the future I may change it to work recursively.
I've added more docs for clarification.

prow/config/cmd/generate.go
chizhg: Done.

prow/config/cmd/generate.go
chizhg: This change isn't quite related to my change (mostly due to unrelated refactoring/renaming), but I've added some comments here based on my understanding. It looks some improvements can also be done here in the future, e.g. reading the latest tag for the image of the new release branch, instead of adding a new tag for the image of the current branch.

nmittler (approved):

PR #3543, opened 2021-09-07
Make istiodremote pilot tests required

frankbu:


ericvn (commented):
Can we just unhide and unskip the tests and leave them optional based on the lack of testing shown at https://testgrid.k8s.io/istio_istio#integ-pilot-istiodremote-tests_istio. Once we have more test data, then we can make required.

ericvn (approved):
After looking at the tests running in post-submit: https://prow.istio.io/job-history/gs/istio-prow/logs/integ-pilot-istiodremote-tests_istio_postsubmit they do seem to behaving very well.

PR #3573, opened 2021-09-21
Adding cosign key env variables to 'release' preset 

stewartbutler:
This adds a new environment variable referencing a new secret on the prow clusters, which points to a gcpkms key that can be used to sign images via Cosign.


howardjohn (approved):

brian-avery (approved):

jwendell (commented):
releases 1.10 and 1.11 were not updated. Is this intentional?

stewartbutler:
> releases 1.10 and 1.11 were not updated. Is this intentional?

It was not. The 1.10 and 1.11 jobs overwrote the presets set by `.base.yaml`. I've removed the override (it was a noop prior to my change).

PR #3584, opened 2021-09-27
Revert "Update Prow to v20210924-1544aef1b6"

howardjohn:
Reverts istio/test-infra#3580

The timing of this exactly correlates to https://github.com/istio/istio/issues/35384; see https://testgrid.k8s.io/istio_istio#integ-security-k8s-tests_istio&exclude-non-failed-tests=20&width=20.

Note: it sounds very likely its NOT a prow issue but a GKE issue. However, since this occurred at literally the exact same time as the prow bump, revert seems like a reasonable first step. See https://github.com/kubernetes/test-infra/issues/23741

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -5,7 +5,7 @@ presubmits:
   run_if_changed: '^(\.prow|prow/(config|plugins|cluster/jobs/.*))\.yaml$'
   spec:
     containers:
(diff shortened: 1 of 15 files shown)


ericvn (approved):
At least as an experiment based on the timing, although sounds like there might be an underlying issue as well.

howardjohn:
Nevermind, this is confirmed to be a diffeent issue

PR #3605, opened 2021-10-07
Use full pilot istiodremote-multicluster tests

frankbu:
Remove `integ-istiodremote-k8s-tests` job in favour of `integ-pilot-istiodremote-multicluster-tests` for multicluster external control plane integration testing.


stevenctl:
lgtm – slight naming bikeshed: don't we call this "externalcontrolplane" or "externalcp" elsewhere? "istiodremote" is a bit confusing as it could refer to the "shared control plane" or "remote cluster" concept. 

I'll approve after you add the skip for the old multicluster job as well

frankbu:
> don't we call this "externalcontrolplane" or "externalcp" elsewhere?

No. `istiodremote` has been used consistently in all the jobs, for better or worse. We could do a global change in a followup PR, but not sure it's worth it.

stevenctl (approved):

PR #3663, opened 2021-11-16
Try moving some Prow jobs to use workload identity

chizhg:
If it works, I'll start moving all other Prow jobs to use workload identity and deprecate the service account key file.

### DIFF
--- a/prow/config/jobs/tools-1.12.yaml
+++ b/prow/config/jobs/tools-1.12.yaml
@@ -45,6 +45,7 @@ jobs:
   - --token-path=/etc/github-token/oauth
   - --script-path=../common-files/bin/create-buildtools-and-update.sh
   name: containers
+  service_account_name: prowjob-default-sa
   node_selector:
     testing: test-pool
   regex: docker/.+|cmd/.+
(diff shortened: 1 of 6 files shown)


prow/config/jobs/tools-1.12.yaml
```diff
@@ -81,13 +81,13 @@ jobs:
   - perf_dashboard
   - deploy
   name: deploy-perf-dashboard
+  service_account_name: prowjob-default-sa
```
howardjohn: One thing I am a bit confused about is the naming here. "default sa" seems to suggest that If I don't set any SA name I would get it - but in this case its actually giving extremely high privileges? 

Am I misunderstanding something or is the name just causing confusion?

chizhg: I can make two changes:
1. Remove https://github.com/istio/test-infra/blob/master/prow/config.yaml#L25
2. Change the name in https://github.com/istio/test-infra/blob/master/prow/cluster/build/prowjob_service_accounts.yaml#L10 to `prowjob-admin-sa`, and also update the names in my PR.

Does the change look right to you?

chizhg: Updated. Please take a look at the latest commit for delta changes.

howardjohn (approved):

chaodaiG (approved):

PR #3664, opened 2021-11-16
Move more Prow jobs to use workload identity

chizhg:
1. Move all jobs except release pipelines to use workload identity
2. Set a service account that only has storage object admin permissions as the default (to fix https://prow.istio.io/view/gs/istio-prow/logs/ci-test-infra-branchprotector/1460471385561239552)

### DIFF
--- a/prow/cluster/prowjob_service_accounts.yaml
+++ b/prow/cluster/prowjob_service_accounts.yaml
@@ -3,11 +3,21 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   annotations:
+    iam.gke.io/gcp-service-account: istio-prow-test-job-default@istio-testing.iam.gserviceaccount.com
+  labels:
+    app.kubernetes.io/part-of: prow
+  namespace: test-pods
+  name: prowjob-default-sa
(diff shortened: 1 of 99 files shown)


chaodaiG (approved):

ericvn (approved):

prow/cluster/prowjob_service_accounts.yaml
```diff
@@ -1,13 +1,23 @@
...
+apiVersion: v1
+kind: ServiceAccount
 metadata:
   annotations:
     iam.gke.io/gcp-service-account: istio-prow-test-job@istio-testing.iam.gserviceaccount.com
   labels:
     app.kubernetes.io/part-of: prow
   namespace: test-pods
-  name: prowjob-admin-sa
+  name: prowjob-advanced-sa
```
howardjohn, line 20: Can we comment somewhere what permissions this actually has? Its pretty hard to tell from this which will likely lead to incorrect usage

chizhg: Done in another PR.

PR #3665, opened 2021-11-16
Move release pipeline Prow jobs to use workload identity

chizhg:
I'm not sure if this works or not. Probably we have to use the secret key?

/hold

### DIFF
--- a/prow/cluster/build/prowjob_service_accounts.yaml
+++ b/prow/cluster/build/prowjob_service_accounts.yaml
@@ -35,3 +35,13 @@ metadata:
   # with a Googler on what permissions it has.
   # Please only use it when you do need them.
   name: prowjob-advanced-sa
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: 202987436673-compute@developer.gserviceaccount.com
+  labels:
+    app.kubernetes.io/part-of: prow-build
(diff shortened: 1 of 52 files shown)


prow/config/jobs/tools-1.12.yaml
```diff
@@ -184,19 +184,6 @@ requirement_presets:
       name: cgroup
     - emptyDir: {}
       name: docker-root
-  release:
-    annotations: {}
-    args: null
-    env:
-    - name: COSIGN_KEY
```
howardjohn, line 191: is t his replaced somehow?

prow/cluster/build/prowjob_service_accounts.yaml
```diff
@@ -35,3 +35,13 @@ metadata:
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: 202987436673-compute@developer.gserviceaccount.com
+  labels:
+    app.kubernetes.io/part-of: prow-build
+  namespace: test-pods
+  name: rel-pipeline-service-account
```
howardjohn, line 47: Can we add a comment when/why it should be used?

chizhg:
I just realized it's a bit tricky to make this change, since the release jobs are not using the default service account. I'll get back to it if I have more bandwidth. Closing it for now.

PR #3673, opened 2021-11-18
Simplify integ test definitions

howardjohn:
Depends on https://github.com/istio/istio/pull/35354/files. This just cleans things up mostly. There is 1 real change - operator test now runs in postsubmit as well, as all tests are expected to run in pre+postsubmit.

Justifications inline

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -30,39 +30,17 @@ jobs:
     command: [entrypoint, make, benchtest, report-benchtest]
     resources: benchmark
 
-  - name: integ-pilot-k8s-tests
-    types: [presubmit]
-    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube.presubmit]
-    requirements: [kind]
-    env:
-      - name: TEST_SELECT
-        value: "-postsubmit,-flaky"
-
   - name: integ-cni-k8s-tests
     types: [presubmit]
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -30,39 +30,17 @@ jobs:
     command: [entrypoint, make, benchtest, report-benchtest]
     resources: benchmark
 
-  - name: integ-pilot-k8s-tests
```
howardjohn, line 33: instead of define twice, once for postsubmit and once presubmit, we can share 1 definition.

prow/config/jobs/istio.yaml
```diff
@@ -30,39 +30,17 @@ jobs:
...
-      - name: TEST_SELECT
-        value: "-postsubmit,-flaky"
-
   - name: integ-cni-k8s-tests
     types: [presubmit]
     command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube.presubmit]
     requirements: [kind]
     env:
-      - name: TEST_SELECT
-        value: "-postsubmit,-flaky"
```
howardjohn, line 47: this is injected automatically now

prow/config/jobs/istio.yaml
```diff
@@ -123,15 +101,11 @@ jobs:
         value: "ipv6"
 
   - name: integ-operator-controller-tests
-    types: [presubmit]
```
howardjohn, line 126: real change: make it run in postsubmit as well

ericvn (approved):

PR #3703, opened 2021-12-08
Add periodic job to run bin/update_deps.sh in master

jacob-delgado:
Periodically we have to run this to get the latest version of distroless images as well as make sure some of the istio packages are up-to-date. Let's automate this so we no longer have to worry about this.

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
@@ -60,6 +60,64 @@ periodics:
       name: cgroup
     - emptyDir: {}
       name: docker-root
+- annotations:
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
As is I think we will have duplication here. At the very least we auto update proxy and api in other paths, not sure if there are others. These may conflict in some way

ericvn:
Agree that we already have changes to keep the other repos up to date in istio (with the exception that the common-files updates), and some of those are cross other repos so we don't want to remove them all. I'm not sure how often the distroless images are updated.

jacob-delgado:
They get updated about once every few weeks based on https://github.com/GoogleContainerTools/distroless/commits/main

What if we were to divide up the update_deps.sh script? Let me do a refactor and then update this to only update those images.

PR #3737, opened 2022-01-04
Restore CustomResourceDefinition to v1beta1

chases2:
Partial rollback of https://github.com/istio/test-infra/pull/3736

### DIFF
--- a/boskos/cluster/boskos-crds.yaml
+++ b/boskos/cluster/boskos-crds.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: dynamicresourcelifecycles.boskos.k8s.io
@@ -16,7 +16,7 @@ spec:
       served: true
       storage: true
 ---
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: resources.boskos.k8s.io
--- a/prow/cluster/kubernetes-external-secrets_crd.yaml
+++ b/prow/cluster/kubernetes-external-secrets_crd.yaml
@@ -1,6 +1,6 @@
 ---
 # From https://github.com/external-secrets/kubernetes-external-secrets/tree/f866e34e0319d9c54ea17d07f5d5818a28d9d0f6
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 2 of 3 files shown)


howardjohn (approved):
We should update them, v1beta1 is removed in 1.22+.

Fine with this for short term though

chases2:
Agree, but it looks like there's some field drift between the two versions, so it's not as easy as it should be:

```
error validating "cluster/kubernetes-external-secrets_crd.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "subresources" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): missing required field "versions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]; if you choose to ignore these errors, turn validation off with --validate=false
error validating "cluster/prowjob_customresourcedefinition.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "additionalPrinterColumns" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): missing required field "versions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]; if you choose to ignore these errors, turn validation off with --validate=false
```

howardjohn:
you can apply them to cluster, then do `kubectl get crd -oyaml ` and it
will convert it. There will be some extra junk added that can be manually
cleaned up (or https://github.com/itaysk/kubectl-neat can help with that)

Not sure any better approach.

Or find the source of the CRDs and copy them over, I assume we did not
hand-write these and upstream has updated - just not sure where

e-blackwelder (approved):

PR #3739, opened 2022-01-05
Setup envoy release-1.11, release-1.12 and release-1.13 branches

jacob-delgado:
Copied the envoy release-1.10 branch information, updated them for 1.11 and 1.12.


jacob-delgado:
Putting a DNM on this. Looking for feedback from for someone else I may need cc @ericvn @howardjohn

prow/config/jobs/envoy-1.12.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+  - name: test-release
+    types: [presubmit]
+    env:
+      - name: BAZEL_BUILD_EXTRA_OPTIONS
+        value: "--local_ram_resources=131072 --local_cpu_resources=42 --test_env=ENVOY_IP_TEST_VERSIONS=v4only --flaky_test_attempts=9"
+      - name: ENVOY_SRCDIR
+        value: "/home/prow/go/src/istio.io/envoy"
+    command: [./ci/do_ci.sh, bazel.release]
+
```
jacob-delgado, line 40: Removed the auto-update job

howardjohn (approved):

prow/config/jobs/envoy-1.12.yaml
ericvn: I had thought we wanted to keep the autoupdate job so RMs didn't have to know to pull newer releases and we said that any envoy PR in the release stream was good. Maybe we could leave the job and remove the auto-merge so the RM knows a change is there, but has the ability to say whether it merges or not.

jacob-delgado: This is for the envoy repo. We only need this for istio-private, not for istio/istio.

jacob-delgado: We don't want it to update istio(-private)/proxy with the istio(-private)/envoy repository. Hopefully that makes it clear.

PR #3742, opened 2022-01-06
Move Authentikos' ClusterRoleBinding to v1

chases2:
Safe migration for v1.22 per https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22
/assign @cjwagner @chaodaiG

### DIFF
--- a/authentikos/examples/authentikos-deployment.yaml
+++ b/authentikos/examples/authentikos-deployment.yaml
@@ -32,7 +32,7 @@ rules:
   - update
 ---
 kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


e-blackwelder (approved):

howardjohn (approved):

fejta:
Suspect that you won't simply be able to apply a version change to a resource (need to replace it)

cjwagner:
> Suspect that you won't simply be able to apply a version change to a resource (need to replace it)

We were wondering about that as well, but as far as I can tell this doesn't seem to be the case.

Either way, these are example and test manifests so they are applied to fresh clusters without existing resources.

howardjohn:
In my experience you could apply new versions on top of old ones (and even
back and forth)

PR #3748, opened 2022-01-06
Deploy schemaed ProwJob CRD with the rest of Prow.

cjwagner:
/assign @chaodaiG 
/cc @chases2


prow/Makefile
```diff
@@ -29,6 +29,8 @@ update-config: get-cluster-credentials
 		--wet
 
 deploy: get-cluster-credentials get-api-resources
+# Apply the ProwJob CRD with --server-side=true due to its size.
+	kubectl apply --server-side=true -f ./cluster/prowjob-crd/ --prune -l app.kubernetes.io/part-of=prow $(PRUNE_WL)
```
chases2, line 33: Doesn't this apply the new CRD twice? Once here, and again with the next kubectl apply (but unsuccessfully and with an error, since there's no server-side=true)?

cjwagner: No the next kubectl apply operates on the `./cluster/` directory not the `./cluster/prowjob-crd/` directory and kubectl does not recursively descend in to subdirectories. (See the other make targets in this file for examples.)

cjwagner:
PTAL

chaodaiG (approved):

PR #3755, opened 2022-01-10
Update go dependencies

howardjohn:


ericvn (approved):

stewartbutler (approved):

howardjohn:
@chaodaiG this seems like a regression from https://github.com/kubernetes/test-infra/issues/22888? `job_states_to_report` is removed again, which sounds like will break things: https://github.com/kubernetes/test-infra/pull/24168?

chaodaiG:
> @chaodaiG this seems like a regression from [kubernetes/test-infra#22888](https://github.com/kubernetes/test-infra/issues/22888)? `job_states_to_report` is removed again, which sounds like will break things: [kubernetes/test-infra#24168](https://github.com/kubernetes/test-infra/pull/24168)?

As mentioned in https://github.com/kubernetes/test-infra/pull/24168, that requiring `job_states_to_report` to be set on every prowjob is not the best UX. This problem was solved in https://github.com/kubernetes/test-infra/pull/24280 by introducing a `report` bool field that's computed from `job_states_to_report`

PR #3764, opened 2022-01-13
Remove `kubectl apply` pruning logic.

cjwagner:
The pruning logic was designed to prevent tech debt associated with resources being applied to the cluster without being checked in. However, this logic can be confusing and it is easy to mistakenly trigger pruning of resources. (We've done so or nearly done so multiple times.) This combined with the fact that we haven't encountered problems with that class of tech debt on our other Prow instances that don't use pruning, leads me to think this is probably more harm that it is worth.
If we change our minds in the future we should consider a more robust solution with better safeguards such as terraform or helm.

/assign @chaodaiG 
/cc @howardjohn 

I'll remove the `part-of` labels in a follow up PR.


chaodaiG (approved):
/hold

PR #3768, opened 2022-01-14
Remove unused app.kubernetes.io/part-of label from all components.

cjwagner:
Depends on #3764, only the latest two commits are unique to this PR and should be reviewed here.
/hold

This also cleans up the use of `--prune` in the boskos Makefile where it was also used (missed that before).
/assign @chaodaiG @listx

### DIFF
--- a/prow/cluster/build/buildcache.yaml
+++ b/prow/cluster/build/buildcache.yaml
@@ -2,7 +2,6 @@ apiVersion: v1
 kind: PersistentVolume
 metadata:
   labels:
-    app.kubernetes.io/part-of: prow-build
   name: fileserver
   namespace: test-pods
 spec:
@@ -18,7 +17,6 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
-    app.kubernetes.io/part-of: prow-build
   name: build-cache-claim
   namespace: test-pods
 spec:
--- a/prow/cluster/build/buildcache.yaml
+++ b/prow/cluster/build/buildcache.yaml
@@ -1,7 +1,6 @@
 apiVersion: v1
 kind: PersistentVolume
 metadata:
-  labels:
   name: fileserver
   namespace: test-pods
 spec:
@@ -16,7 +15,6 @@ spec:
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  labels:
   name: build-cache-claim
   namespace: test-pods
 spec:
--- a/prow/cluster/prow-controller-manager.yaml
+++ b/prow/cluster/prow-controller-manager.yaml
@@ -4,7 +4,6 @@ metadata:
   namespace: default
   name: prow-controller-manager
   labels:
-    app.kubernetes.io/part-of: prow
     app: prow-controller-manager
 spec:
   replicas: 1 # TODO(fejta) - delete plank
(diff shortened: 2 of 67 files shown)


listx:
Assuming `app.kubernetes.io/part-of` was only used by the pruning logic, this seems fine.

/lgtm

cjwagner:
> Assuming `app.kubernetes.io/part-of` was only used by the pruning logic, this seems fine.

That seems to be the case. I see no other occurrence of the string `part-of` in the repo.

listx (approved):
LGTM, thanks!

listx (approved):

prow/cluster/build/buildcache.yaml
```diff
@@ -18,7 +17,6 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   labels:
-    app.kubernetes.io/part-of: prow-build
```
jacob-delgado, line 21: if there are no other labels left, doesn't this fail due to being invalid yaml?

cjwagner: Oh, I think you're right. Let me fix that.

prow/cluster/prow-controller-manager.yaml
```diff
@@ -60,15 +59,13 @@ metadata:
   annotations:
     iam.gke.io/gcp-service-account: prow-control-plane@istio-testing.iam.gserviceaccount.com
   labels:
```
listx: Missed this one (and others in this file)

prow/cluster/build/buildcache.yaml
cjwagner: Thanks Jacob, I learned about multi-line sed and fixed things up. I've skimmed through and everything looks right to me now.

prow/cluster/velodrome/velodrome_configmap-nginx.yaml
```diff
@@ -4,7 +4,6 @@ metadata:
   name: nginx
   namespace: velodrome
   labels:
```
listx: This one too

listx (commented):
Also did you mean to remove the execute bit from tools/automator/automator.sh? (I wanted to comment there but I don't know how to in the GitHub file review UI where there is no textual diff)

cjwagner:
I have no idea how the script permission bits changed. I didn't know that sed would even modify that...

There were a few other places my sed script missed as well. I think I've fixed them. If I haven't I'm just going to remove the commit that removes the labels so I can merge pruning deprecation for boskos. The labels don't hurt anything so this is really just a waste of time.

listx (approved):

fejta:
Not sure if we can apply label changes. Will we need to manually recreate these resources?

fejta (approved):
/hold

cjwagner:
> Not sure if we can apply label changes. Will we need to manually recreate these resources?

I'm not sure. Seems like it might be the case that we cannot delete labels with `kubectl apply`. Would you prefer for me to drop the commits removing the label like I mentioned [here](https://github.com/istio/test-infra/pull/3768#issuecomment-1013499686)? This isn't worth the hassle of manually recreating the resources.

cjwagner:
PTAL, I'd like to get this in. If the `part-of` labels aren't actually removed from resources in the cluster it doesn't really matter.

listx (approved):

cjwagner:
/cc @jacob-delgado @howardjohn
PTAL

howardjohn (approved):

PR #3783, opened 2022-01-19
Truncate long job names

jacob-delgado:
There is a limit to job names (being 63 characters). This is preventing us from getting CI/CD done for the istio-private organization.

See https://github.com/istio/test-infra/pull/3739/ for the PR that is failing and the test grid job is here https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_test-infra/3739/pull-test-infra-check-testgrid-config/1483874476742414336/build-log.txt

Changes
* remove k8s-tests suffix from integration test names
* remove tests suffix from integration test names
* for long names that include "multicluster" change them to mc (e.g. integ-pilot-istiodremote-multicluster to integ-pilot-istiodremote-mc)

### DIFF
--- a/prow/config/jobs/istio-1.10.yaml
+++ b/prow/config/jobs/istio-1.10.yaml
@@ -80,7 +80,7 @@ jobs:
   env:
   - name: TEST_SELECT
     value: -postsubmit,-flaky,-multicluster
-  name: integ-pilot-k8s-tests
+  name: integ-pilot
   node_selector:
     testing: test-pool
   requirements:
@@ -96,7 +96,7 @@ jobs:
   env:
   - name: TEST_SELECT
     value: -postsubmit,-flaky,-multicluster
-  name: integ-security-k8s-tests
+  name: integ-security
   node_selector:
     testing: test-pool
   requirements:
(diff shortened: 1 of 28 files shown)


prow/config/jobs/istio-1.10.yaml
```diff
@@ -231,7 +231,7 @@ jobs:
   env:
   - name: TEST_SELECT
     value: -multicluster
-  name: integ-pilot-multicluster-tests
+  name: integ-pilot-mc-tests
```
howardjohn: remove `-tests` perhaps?

prow/config/jobs/istio-1.12.yaml
```diff
@@ -314,7 +314,7 @@ jobs:
   - presubmit_optional
   - presubmit_skipped
   - hidden
-  name: integ-pilot-istiodless-multicluster-tests
+  name: integ-pilot-less-mc-tests
```
howardjohn: would be nice to keep "istiodless" as "less" confused me. I think removing the `-tests` should give us room.

howardjohn (commented):
ok to approve this but we should add a test that asserts this?

jacob-delgado:
> ok to approve this but we should add a test that asserts this?

Added a unit test so that `make generate-config` fails before and we won't have to wait until CI/CD.

howardjohn (approved):

PR #3802, opened 2022-01-24
fix centos image

stevenctl:
Not sure how this ended up with the invalid tag, but it's breaking istio/proxy: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_proxy/3670/release-centos-test_proxy_release-1.13/1485692994966261760


ericvn (approved):

ericvn:
It seems that particular image was not built like the remainder of the images.

PR #3804, opened 2022-01-24
Fix the service account resource for the prow private build cluster

chizhg:
Not sure why these service accounts already exist in the private build cluster. @howardjohn I guess you manually applied them?


howardjohn (approved):
Yeah I applied it

cjwagner (approved):

PR #3832, opened 2022-02-05
Remove dedicated ipv6 pool

howardjohn:
GKE COS 1.21+, which we are on, can support the ipv6 tests now. No need
for dedicated pool. I manually verified this so I am fairly confident it
will work.


myidpt (approved):

myidpt:
Thanks John!

jacob-delgado (approved):

PR #3854, opened 2022-02-16
Major refactor and improvements to the prowgen tool

chizhg:
This PR is an effort to further generalize the prowgen tool so it can be used by more other projects. 

It involves multiple refactors and improvements, including:
1. Break the large `generate.go` file into multiple files and packages
2. Add `--pre-process-command` and `--post-process-command` to support preprocessing and postprocessing config files. These can be used when the users need customized config generation logic which cannot be supported by `prowgen`
3. Add `params` field to support configuring params with `$(params.key)` syntax
4. Clean up unused functions, improve README and comments, etc.
5. Make `istio.io/test-infra/tools/prowgen` a Go submodule so it's easier to be imported by other projects.


chizhg:
cc @chaodaiG @cjwagner

jacob-delgado:
Once @chaodaiG and @cjwagner take a look through this I can go through it and approve.

tools/prowgen/Dockerfile
```diff
@@ -12,13 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.17 as build-env
+FROM golang:1.16 as build-env
```
howardjohn, line 15: Why downgrade?

tools/prowgen/cmd/prowgen/main.go
```diff
@@ -133,17 +125,34 @@ func main() {
...
 
 			return nil
 		}); err != nil {
-			exit(err, "walking through the meta config files failed")
+			log.Fatalf("Walking through the meta config files failed: %v", err)
 		}
 	} else {
+		// Set INPUT and OUTPUT env vars for the pre-process and post-process
+		// commands to consume.
+		os.Setenv("INPUT", *inputDir)
```
howardjohn: would prefer to pass these to `cmd.Env`  directly instead of messing with the environment of this process

tools/prowgen/cmd/prowgen/main.go
```diff
@@ -133,17 +125,34 @@ func main() {
...
+		os.Setenv("OUTPUT", *outputDir)
+
+		if *preprocessCommand != "" {
```
howardjohn, line 145: why not just have user do `pre-process && prowgen`?

tools/prowgen/cmd/prowgen/main.go
```diff
@@ -152,20 +161,20 @@ func main() {
 		// Store the job config generated from all meta-config files in a cache map, and combine the
 		// job configs before we generate the final config files.
 		// In this way we can have multiple meta-config files for the same org/repo:branch
-		cachedOutput := map[ref]k8sProwConfig.JobConfig{}
+		cachedOutput := map[ref]*k8sProwConfig.JobConfig{}
```
howardjohn: nit: why switch to a pointer? if it can never be nil this seems to make things more confusing

tools/prowgen/cmd/prowgen/main.go
```diff
@@ -152,20 +161,20 @@ func main() {
...
 			}
 			if err != nil {
-				fmt.Printf("error: %s\n", err.Error())
+				log.Fatal(err)
 			}
 
 			baseConfig := bc
 			if _, err := os.Stat(filepath.Join(path, ".base.yaml")); !os.IsNotExist(err) {
-				baseConfig = pkg.ReadBase(baseConfig, filepath.Join(path, ".base.yaml"))
+				baseConfig = *pkg.ReadBase(&baseConfig, filepath.Join(path, ".base.yaml"))
```
howardjohn: if we are always going to dereference then make we should make ReadBase not return a pointer?

tools/prowgen/cmd/prowgen/main.go
```diff
@@ -198,25 +206,58 @@ func main() {
...
 			case "write":
-				pkg.Write(output, fname, bc.AutogenHeader)
-			case "diff":
-				existing := pkg.ReadProwJobConfig(fname)
-				pkg.Diff(output, existing)
-			default:
+				if e := pkg.Write(output, fname, bc.AutogenHeader); e != nil {
+					err = multierror.Append(err, e)
+				}
+				if *postprocessCommand != "" {
```
howardjohn, line 215: same comment as the preProcess

tools/prowgen/go.mod
```diff
@@ -0,0 +1,22 @@
+module istio.io/test-infra/tools/prowgen
```
howardjohn, line 1: is there value in splitting this?

tools/prowgen/pkg/decorator/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
```
howardjohn, line 1: outside of scope but seems like we can probably remove the bazel stuff at some point...

tools/prowgen/Dockerfile
chizhg: Mainly to keep it consistent with the Go version in go.mod.

tools/prowgen/cmd/prowgen/main.go
chizhg: Good point. Done.

tools/prowgen/cmd/prowgen/main.go
chizhg: The intent is to make it a built-in feature, so it's more clear to the users that it's a common practice to do pre-process/post-process when using `prowgen`. Otherwise they may consider adding non-generic feature to `prowgen`, or consider it hacky to do pre-process/post-process, neither of which will be ideal since `prowgen` itself cannot accommodate all use cases.

tools/prowgen/cmd/prowgen/main.go
chizhg: Ditto comment as preProcess

tools/prowgen/pkg/decorator/BUILD.bazel
chizhg: Agreed. It looks these .bazel files are not actually used for any critical jobs - since presubmit Prow jobs didn't fail when I didn't update them.

tools/prowgen/cmd/prowgen/main.go
chizhg: I was thinking using pointer is more efficient since these structs are relatively large, but I agree it's more readable when directly using values. I have removed the pointers here and elsewhere.

tools/prowgen/cmd/prowgen/main.go
chizhg: Removed the pointers.

tools/prowgen/go.mod
chizhg: Defining it into a separate Go module will make it easier and reduce the number of dependencies for users to import the structs in spec.go, when they need to implement their own pre-process command.

howardjohn (approved):

PR #3873, opened 2022-02-23
Remove dead tools and scripts

howardjohn:
Most of these are many many years old and clearly are not compatible
with current infra


chaodaiG (approved):
lgtm from my side

chases2 (approved):

jacob-delgado (approved):

PR #3875, opened 2022-02-24
private build: drop oauth-token

howardjohn:
private jobs do not need this
* Automator doesn't work in private
* Release builder doesn't need it after
  https://github.com/istio/release-builder/commit/a91e06df41114977ddbd9d0b3710ee35d3408bc5

This is done via denylist the automator jobs, and changing the real prod
release-builder jobs to drop the mount. This was added in the past when
we did in-line docker base image builds; now we do them as periodic jobs


jacob-delgado (approved):
still need the PRs approved for 1.12 and possibly 1.11 but we'll get those in before the next security patch anyway

PR #3877, opened 2022-02-24
Fix configurator command

howardjohn:
"Broken" in 701ef54495. We should probably not use `:latest` and get it
bumped by the bump script


howardjohn:
https://github.com/istio/test-infra/pull/3878

PR #3919, opened 2022-03-11
Update scripts for shellcheck 0.8

jacob-delgado:


ericvn (approved):

jacob-delgado:
cc @cjwagner We would like to upgrade shell-check used in our docker containers, but this PR being blocked prevents us from doing so. Anything I can do to get this job to work?

jacob-delgado:
cc @chaodaiG can you also take a look at this?

chaodaiG:
`integ-test-authentikos_test-infra` is broken, we should just disable it imo

chaodaiG (approved):
lgtm

might be better to delete `integ-test-authentikos` in a separate PR

PR #3929, opened 2022-03-15
Adding tag override to prowtrans

stewartbutler:
Complement to Hub override. Also modified hub override to use
go-containerregistry.pkg.name to parse the tag, to make sure
replacement *only* happens in the hub/repo.


tools/prowtrans/cmd/prowtrans/main.go
```diff
@@ -806,13 +812,29 @@ func updateExtraRefs(o options, job *config.UtilityConfig) {
 // updateHubs updates the docker hubs for container images
 func updateHubs(o options, job *config.JobBase) {
 	for i := range job.Spec.Containers {
+		tag, _ := dockername.NewTag(job.Spec.Containers[i].Image)
```
howardjohn, line 814: what is NewTag actually doing for us? Isn't Name() returning the same thing the input was?

stewartbutler: No, we are pulling `Context().Name()` which only pulls the registry part. Drops the SHA or tag, just gives the repo/hub/whatever. This way we can make sure we only do the replace on that part.

stewartbutler: I did this because I'm lazy and don't want to think about all the dumb docker edge cases, so I just pulled from a good library.

tools/prowtrans/cmd/prowtrans/main.go
```diff
@@ -806,13 +811,29 @@ func updateExtraRefs(o options, job *config.UtilityConfig) {
...
+		repostr := tag.Context().Name()
 		for in, out := range o.HubMap {
-			job.Spec.Containers[i].Image = strings.ReplaceAll(
-				job.Spec.Containers[i].Image,
+			repostr = strings.ReplaceAll(
```
coryrc: nit: I think `baseRef` would make a better name

like https://github.com/google/go-containerregistry/blob/63e6f4ed9c31bfefe26495cc9f3bd1f794174452/cmd/crane/cmd/append.go#L32

"RepositoryStr" (at least in the context of go-containerregistry) seems to only apply to a portion of this string, like https://github.com/google/go-containerregistry/blob/7e30746dde2f5a19b609f885dfef4e7b1d1503ad/pkg/name/repository.go#L51

stewartbutler: Good point. Fixed.

howardjohn (approved):

PR #3933, opened 2022-03-16
[WIP] Specify a pgp key to be used for helm provenance

jacob-delgado:
Do we wish to have helm provenance? See https://helm.sh/docs/topics/provenance/ for more details


howardjohn (commented):
Do they only support file path? Not reading from GCP KMS like cosign?

Can we just cosign our OCI helm package?

jacob-delgado:
Decided to go with using cosign for any artifact going into oci registry

PR #3963, opened 2022-03-28
Remove k8s 1.16-1.18 jobs for master

howardjohn:
There was some consensus between Jacob, Eric, Rama, and I to drop these
tests on master only. This would shift 1.14 to be testing 1.19-1.24. Our
official support policy would be 1.21-1.24, so we still test and
additional 2 versions beyond our official support.


howardjohn:
SSA PR is not going to work well without 1.20+, so maybe we wait

PR #3971, opened 2022-04-01
Run multicluster tests istiodless

frankbu:
Part of closing https://github.com/istio/istio/issues/27420

Depends on https://github.com/istio/istio/pull/38219


nmittler:
Instead of doing this, can we just make that flag default to true in istio/istio?

frankbu:
@nmittler

> Instead of doing this, can we just make that flag default to true in istio/istio?

Yes, that would be better, but I was thinking that it would be good to initially just leave it like this so that it's clearly visible that we are now switching to istiodless testing. We can then change the default and remove the extra arg in a future PR. I don't have a strong opinion on this, so I can change it now if you really want.

howardjohn (approved):

PR #3998, opened 2022-04-19
Add knative configgen as an example for using prowgen tool as a library

chizhg:


tools/prowgen/README.md
```diff
@@ -226,15 +226,7 @@ requirement_presets:
...
 
-## Generating the config
-
-### `make` command
-
-Istio Prow jobs config files can be generated with:
-
-```bash
-make generate-config
-```
```
ericvn: Not sure why this was removed. It is still valid.

chizhg: I was thinking this entry is specific to Istio, while this README is for general purpose. I have added it back for now, and will probably try to rework the README in the future.

ericvn (approved):

PR #4007, opened 2022-04-21
Add prow jobs for Istio 1.14

lei-tang:
Add prow jobs for Istio 1.14.

### DIFF
--- a/prow/config/jobs/api-1.14.yaml
+++ b/prow/config/jobs/api-1.14.yaml
@@ -1,0 +1,602 @@
+branches:
+- release-1.14
+image: gcr.io/istio-testing/build-tools:release-1.14-2022-04-18T18-19-40
+jobs:
+- command:
+  - make
+  - presubmit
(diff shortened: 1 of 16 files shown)


prow/config/jobs/api-1.14.yaml
```diff
@@ -0,0 +1,602 @@
+branches:
+- release-1.14
+image: gcr.io/istio-testing/build-tools:release-1.14-2022-04-18T18-19-40
+jobs:
+- command:
+  - make
+  - presubmit
+  image: gcr.io/istio-testing/build-tools:master-2022-04-18T18-19-40
```
ericvn, line 8: This is one of the image lines that need to be removed. If the inner line matches the outer except for the release, remove the inner..  Be careful in proxy as some inner images are different. tools-proxy vs tools-centos for example.

ericvn:
Making DNM so this doesn't go in prematurely.

lei-tang:
Replaced by https://github.com/istio/test-infra/pull/4008.

PR #4082, opened 2022-06-01
Add experimental ds branch ci/cd

jacob-delgado:
Discussed in the istio community meeting. @zhlsunshine will be working with others in the community to get experimental support. This will not only be done for experimental changes as the changes necessary for dual-stack support have major implications for the other parts of the system.


howardjohn (changes requested):
Need to turn off release test or we will push to master ovrewriting the real images

also wonder if we can/should use the job transformer we use for private jobs?

howardjohn:
maybe we should run ci for `experimental-*`

jacob-delgado:
> 

Who is allowed to create branches in istio? Don't you just have to be a member? I think that's reasonable if regex/globs are allowed.

prow/config/jobs/istio-experimental-ds.yaml
```diff
@@ -0,0 +1,403 @@
+org: istio
```
jacob-delgado: This is based on `istio.yaml` for `istio/istio`. Only changes are listed at the top.

zhlsunshine:
Hi @jacob-delgado & @howardjohn, if ci/cd is enabled for `experimental-dual-stack` branch, shall we get this branch image?  If yes, it will be very convenient for many users.

prow/config/jobs/istio-experimental-ds.yaml
```diff
@@ -0,0 +1,403 @@
...
+
+  - name: release-notes
+    types: [presubmit]
+    command:
+      - ../test-infra/tools/check_release_notes.sh
+      - --token-path=/etc/github-token/oauth
+    requirements: [github]
+    repos: [istio/test-infra@master,istio/tools@master]
+
+  # Recreate base images if the dockerfile changes (in postsubmit)
```
howardjohn: need to drop this as well

prow/config/jobs/istio-experimental-ds.yaml
```diff
@@ -0,0 +1,403 @@
...
+    resources: dedicated
+
+  - name: release-test
+    service_account_name: prowjob-advanced-sa
+    types: [presubmit]
+    command: [entrypoint, prow/release-test.sh]
+    requirements: [docker]
+    resources: dedicated
+
+  - name: release
```
howardjohn: need to drop this

howardjohn (commented):
It would definitely be nice to use the `tools/prowtrans/cmd/prowtrans/main.go` tool. Its meant for these purposes.

jacob-delgado:
@howardjohn I had to modify prowtrans to rename files that are using branches-out. 

Previously, this was generating 2 different `istio.istio.master.gen.yaml` file, however, in different directories. But the CI/CD was complaining that you cannot have 2 files of the same name checked-in to the repo.

So if a user specifies `branches-out` then it'll change the name of the first `branches` option with that of the first `branches-out` option  (e.g. prowtrans now generrates `istio.istio.master.gen.yaml` and `istio.istio.experimental-dual-stack.yaml` for the `experimental-dual-stack` branch).

jacob-delgado:
ping @howardjohn

jacob-delgado:
Any reason why we can't merge this in @howardjohn ?

prow/config/experimental/.defaults.yaml
```diff
@@ -0,0 +1,7 @@
+defaults:
+  mapping:
+    istio: istio
+  input: ./prow/cluster/jobs/
+  output: ./prow/cluster/jobs/
+  cluster: public
+  modifier: exp
```
howardjohn:
```suggestion
  modifier: exp_ds
```

since we may have other experiments?

Or changes branches-out to `experimental-*` so we can just always have this for all experiments? I think I prefer the later.

howardjohn: Might be `.*` not `*` though, I forget the syntax here

howardjohn (approved):

PR #4094, opened 2022-06-08
Remove presets from job configurations

howardjohn:
Presets are a great way to offer a higher level config model that expands to the full config internally. However, Istio already uses a higher level config model which we translate to the full config (and check into git).

By having two ways to do the same thing, things get a bit confusing and harder to manage. We have made mistakes in the past due to misunderstanding the job configs due to this.

This PR removes all presets.

... however, the private jobs, which work by transforming the generating configs, use presets extensively. Because these work on the post-generation config, our higher level config model doesn't apply here, but presets do. To clean this up, we move the private presets to a different file (so they will not accidentally be used by normal prow jobs), and add `resolve: true` to the transformer tool. This basically applies the preset so that its part of the checked in yaml instead of handled by prow at runtime.

The end result is that all of our job configs are fully "hydrated" and we can clearly see the resulting job definitions.

During this process I had a bunch of errors that were impossible to debug in the tooling due to ignoring errors, so I added some error printing.

There should be no impact to jobs as a result of this change; it is expected to behave identically. However, the generated YAML is a bit different.


ericvn (approved):
LGTM

PR #4108, opened 2022-06-16
Add `architectures` support to prowgen

howardjohn:
I originally wanted to use the `matrix` feature. However, given how
    common this will be, I think it makes sense to have a specialized field.
    This makes it easier to apply some of the bespoke customizations we
    want.

For https://github.com/istio/istio/issues/26652#issuecomment-1155682106

Commit 1 is the main change, I recommend reviewing it. The second commit just generates an architecture node selector for each. This should have no impact on any existing jobs but will be needed for when we actually use multi-arch


fejta (approved):

tools/prowgen/pkg/testdata/simple.yaml
```diff
@@ -44,9 +44,20 @@ jobs:
     resources: custom
     command: [prow/command.sh]
     repos: [istio/istio]
-    modifier: [presubmit_skipped]
```
ericvn, line 47: Not sure if there was a reason to remove this line, but noting it was removed in the test-data as well.

tools/prowgen/pkg/testdata/simple.yaml
```diff
@@ -44,9 +44,20 @@ jobs:
...
+    architectures: [amd64, arm64]
+    command: [prow/command.sh]
+    image: test
+
+  - name: multi-arch-param
+    types: [presubmit]
+    architectures: [amd64, arm64]
+    command: [prow/command.sh, $(params.arch)]
+    image: test
+
```
ericvn, line 60: Should we have an arm64 only test job?

tools/prowgen/pkg/testdata/simple.gen.yaml
```diff
@@ -332,6 +339,7 @@ presubmits:
           name: build-cache
           subPath: gocache
       nodeSelector:
+        kubernetes.io/arch: arm64
```
ericvn: Should this one be an amd64?

tools/prowgen/pkg/testdata/simple.yaml
howardjohn: it no longer is a real field - I made the yaml parser strict

tools/prowgen/pkg/testdata/simple.yaml
howardjohn: This means "make an arm job and make an amd job". So it's mostly syntax sugar - they are independent

tools/prowgen/pkg/testdata/simple.gen.yaml
howardjohn: there should be one arm one per the job config

ericvn: Correct, and I believe I am seeing one arm and one amd for the second job config. But I am seeing that the first job config yielded two arm64 jobs (line 376 and line 342 (342 should be amd)) which would be incorrect.

howardjohn: Good catch! was a bug, fixed in last commit

ericvn (approved):

PR #4109, opened 2022-06-16
Multi arch apply jobs

howardjohn:
For https://github.com/istio/istio/issues/26652#issuecomment-1155682106

This introduces new jobs for arm64

tools: new pre/postsubmit for container publishing
proxy: new presubmit jobs for all tests and release test; postsubmit job for release
istio: new integ test for arm, run unit tests on arm, change release to multi-arch (cross compiles)

These cannot merge until we have arm nodes... and we also should split them into three phases (since each depends on the former)


ericvn:
Make not stale.

howardjohn:
split into other PRs

PR #4110, opened 2022-06-16
Fix regression in preset.yaml removal, always set build env var

howardjohn:
Also had to fix an issue in the generator which repeated env vars a
bunch


ericvn (approved):
LGTM

PR #4115, opened 2022-06-22
Add a job for C++ PSM Echo test

yashykt:


prow/config/jobs/istio.yaml
```diff
@@ -95,6 +95,13 @@ jobs:
     command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube]
     requirements: [kind]
 
+  - name: integ-pilot-cpp
+    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube]
+    requirements: [kind]
+    env:
+      - name: GRPC_ECHO_IMAGE
+        value: "yashykt/istio:echo"
```
stevenctl: note: It's preferred that this is built from within the Istio repo and pushed to Istio image repositories rather than some somewhat untrusted image.

yashykt: agreed. Added a TODO

prow/config/jobs/istio.yaml
```diff
@@ -95,6 +95,14 @@ jobs:
     command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube]
     requirements: [kind]
 
+  - name: integ-pilot-cpp
```
howardjohn, line 98: Do we really want a whole new job? I was expecting a new deployment to existing job (eg `proxyless-go`, `proxyless-cpp`, etc). Then you can test communication between them as well. and its way cheaper

yashykt: I am open to suggestions. I believe the reasoning here was that we do not want to modify the tests themselves.

stevenctl: I figured we'd have one new (postsubmit) job, and that job would run all the deployments. 

I don't think presubmit makes sense. If something breaks in gRPC for c++ but not go, that's not really something that should block Istio PRs.

yashykt: made it postsubmit

costinm: How is it different from golang proxyless gRPC tests ? AFAIK the tests are not dependent on the C++ 'head' - but use a fixed image for the echo ( which should be build on each gRPC release ). 

But I am ok with postsubmit - we probably already run far too much in pre-submit and a failure on post-submit will still
uncover the bugs. 

I would like the proxyless gRPC tests to be a separate suite if possible, less noise.

prow/config/jobs/istio.yaml
costinm: I think the previous agreement was that it should be build in the gRPC repo - IMO the 'proxyless' mode ( as well as general interop ) should involve each framework and component building its own images. K8S Gateway tests are similar, each vendor handles its own images ( or infrastructure).

The client side is not under Istio control. 

A more interesting question is if gRPC should produce 'nightly' or 'pre-release' images reflecting their code changes
to be tested against Istio infra. At the very least their release process should involve producing the echo images in all languages and making sure the tests suites are passing ( as I suspect they already do for other XDS servers)

costinm (approved):
Not an expert in the test suites, but as post-submit and separate suite it seems low risk - if we observe flakiness we can fix/revert.

prow/config/jobs/istio.yaml
yashykt: I've moved the image to grpc's repository for now.

stevenctl (approved):

PR #4116, opened 2022-06-23
Drop bazel from this repo, following upstream

### LINKED ISSUE #1580: Remove bazel usage from the repo
No other repo in istio uses bazel. As far as I know there is no reason we need bazel here, and it makes the repo unfamiliar for Istio contributors.

If possible we should remove bazel.

I am not familiar enough with bazel to know how feasible this is and if its as simple as `rm all_bazel_stuff` and replace `bazel test` with `go test`, etc

howardjohn:
Follows https://github.com/kubernetes/test-infra/pull/26039

Fixes https://github.com/istio/test-infra/issues/1580


ericvn (approved):
LGTM

jacob-delgado (approved):

chaodaiG (approved):

PR #4117, opened 2022-06-23
Drop a number of unused Go tools

howardjohn:
I checked usage of these with Github codesearch across the org (ex: https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Aistio+release_note_collector)

I believe all of these are >4 years old and haven't been used in a long time


ericvn (approved):
Thanks for the cleanup

PR #4124, opened 2022-06-23
Add some presets back

howardjohn:
This partially reverts #4094

Basically, the issue is for `requirements` we pin older branches. So we were still using presets unexpected. We could go update all of those, but seems like a pain. Instead, we can just add the presets back only for those branches and remove them once the branches are pruned later.

We also had to remove `resolve` on the private jobs, or they would attempt to double-apply the preset (which fails) since `resolve` doesn't remove the preset label (maybe it should...)

Additionally, unit tests are added to ensure this never happens again.


ericvn (approved):
I'll leave fixing the double application of preset with resolve as a possible future work item.

PR #4173, opened 2022-07-20
Update periodic job for grabing envoy to the 1.23 branch

ericvn:
Temporary until after the 1.15 branch cut happens.

Prevent from making changes in istio/proxy main branch to change in envoy main branch and having to revert them in the 1.15 branch after the cut.


ericvn:
https://github.com/istio/proxy/pull/3948#issuecomment-1190510213 mentions to not make this change, create a new proxy branch ahead of time and then create jobs for that branch to update from 1.23 and push to master.  I'm not sure if that's an easy task. Maybe need to verify where the new release managers are relative to the cut, and not sure if we want to cut all repos yet to limit the number of cherry-picks...

ericvn:
For now, I may have convinced myself that we can fix the proxy 1.15 branch post cut, either manually or create a PR to actually revert the changes.

PR #4180, opened 2022-07-25
Add service accounts to `prowgen` tool

upodroid:
/cc @chaodaiG @howardjohn 

Hi, we are using this generator at knative/test-infra and I need urgent support for specifying service accounts. This change will allow us to safely add all PodSpec parameters except Container, Volumes.


chaodaiG:
/cc @chizhg 
who has more context on this
/uncc
please add me back if my approval is needed

tools/prowgen/pkg/spec/spec.go
```diff
@@ -143,6 +143,7 @@ type RequirementPreset struct {
 	Volumes      []v1.Volume       `json:"volumes,omitempty"`
 	VolumeMounts []v1.VolumeMount  `json:"volumeMounts,omitempty"`
 	Args         []string          `json:"args,omitempty"`
+	Spec         *v1.PodSpec       `json:"spec,omitempty"`
```
chizhg: Nit: I think it's better to name the json key as `podSpec`, and add a comment to explain when to use it - when you want to configure presets for fields in `v1.PodSpec` other than the `metadata` and `containers`.

chizhg (approved):

tools/prowgen/pkg/spec/spec.go
upodroid: Fixed

tools/prowgen/pkg/decorator/requirement.go
```diff
@@ -66,13 +67,13 @@ func ApplyRequirements(job *config.JobBase, requirements, excludedRequirements [
...
+			mergeRequirement(annotations, labels, spec, spec.Containers, &spec.Volumes, req)
 		}
 	}
 }
 
-// mergeRequirement will overlay the requirement on the existing job spec.
-func mergeRequirement(annotations, labels map[string]string, containers []v1.Container, volumes *[]v1.Volume,
+// mergeRequirement will overlay the requirement on the existing job spec. Use mergo for all keys except containers and metadata
+func mergeRequirement(annotations, labels map[string]string, spec *v1.PodSpec, containers []v1.Container, volumes *[]v1.Volume,
 	req spec.RequirementPreset) {
```
ericvn, line 77: Just pass `spec` and get containers from that?
```suggestion
func mergeRequirement(annotations, labels map[string]string, spec *v1.PodSpec, volumes *[]v1.Volume,
	req spec.RequirementPreset) {
	containers := spec.Containers
```

upodroid: That won't be helpful. The containers are processed by another function `createContainer` in `generate.go`. I would need to do a rewrite and modify the istio prow configuration.

upodroid:
Can I get this approved and merged please? Thanks

ericvn (approved):

PR #4188, opened 2022-07-28
Arm build cluster

stewartbutler:
Creating an arm64 build cluster in Prow.


prow/cluster/arm/kubernetes_external_secrets.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: kubernetes-client.io/v1
```
howardjohn: I think we actually don't need any of these. Since its only for the release job which arm never will. I can see any argument that they should be the same, but its easier to add secrets than remove them (since we never know blast radius), so I would be in favor of moving all of these. I believe we don't need any at all.

I'd keep the external secrets deployment though... not sure why, it just feels right :-). We may have some secrets we need eventually

howardjohn (approved):

prow/Makefile
```diff
@@ -46,22 +51,39 @@ deploy-monitoring: get-cluster-credentials
...
+deploy-arm: get-arm-cluster-credentials
+	kubectl apply -f ./cluster/arm
+
+query-arm: get-arm-cluster-credentials
+	kubectl cluster-info
+
 get-private-cluster-credentials: PROJECT=$(PROJECT_PRIVATE)
 get-private-cluster-credentials: CLUSTER=$(CLUSTER_PRIVATE)
-
+get-private-cluster-credentials: ZONE=$(ZONE_PRIVATE)
```
ericvn, line 74: nit: add an empty line after this

prow/cluster/arm/kubernetes-external-secrets_crd.yaml
```diff
@@ -0,0 +1,238 @@
...
+          required:
+            - spec
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                controllerId:
+                  description: The ID of controller instance that manages this ExternalSecret.
+                    This is needed in case there is more than a KES controller instances within the cluster.
```
ericvn, line 33: c/a/one

prow/cluster/arm/kubernetes-external-secrets_crd.yaml
```diff
@@ -0,0 +1,238 @@
...
+                  type: string
+                type:
+                  type: string
+                  description: >-
+                    DEPRECATED: Use spec.template.type
+                template:
+                  description: Template which will be deep merged without mutating
+                    any existing fields. into generated secret, can be used to
```
ericvn, line 41: grammer?

prow/cluster/arm/kubernetes-external-secrets_crd.yaml
stewartbutler: All of this is just copied verbatim over from prow/cluster/build, and I don't want to create a diff there. I think this is actually from upstream, too.

prow/cluster/arm/kubernetes-external-secrets_crd.yaml
stewartbutler: See above.

prow/Makefile
stewartbutler: Done.

chaodaiG (approved):
nothing controversial or potentially risky. Not depend on any other PRs, so LGTM.

Assuming that the build cluster itself is set up according to the create_build_cluster.sh script from upstream in terms of gencred refresher etc.

PR #4193, opened 2022-07-29
arm64: drop bootstrap image

howardjohn:
Hold until we update the image fully


ericvn (approved):
Looks like the using the old image(615...) instead of the new image(88a...).

PR #4208, opened 2022-08-01
Generate Prow jobs for Istio 1.15

Monkeyanator:
Automation did not work, could not find input path for prow jobs. Created this with `go run ./cmd/prowgen/main.go branch 1.15` and `go run ./main.go branch 1.15 --input-dir /home/samnaser/workspace/istio-workspace/test-infra/prow/config/istio-private_jobs`.


ericvn (approved):
Approving this job with a DNM to let the Release 1.15 team the ability to merge when they get the `make gen` and dashboard updates added. I notice the build image was not update to be release-15-...  Also note that the release builder code in a later step creates a new base image. I'm not sure how that code reads new image name and if it will be confused as we changed to using the branch-she.

PR #4210, opened 2022-08-02
Manual update_image of latest 1.14 build-tools image

ericvn:
The automated job runs and want to add a lot of license files (which then fails the gencheck test). Running the job locally does not want to add these files.

This is a manual PR updating the build tools image.

This supersedes https://github.com/istio/test-infra/pull/4209


jacob-delgado (approved):
approving, but will have 1.14 rms remove it

GregHanson:
LGTM

PR #4239, opened 2022-08-15
Remove velodrome

jacob-delgado:
This was removed in kubernetes, see https://github.com/kubernetes/test-infra/issues/16836

We should remove this and velodrome.istio.io should no longer work. This may require work from the operations team.


howardjohn (approved):
LGTM but is there an alternative?

chaodaiG:
> LGTM but is there an alternative?

Grafana is a possible alternative, if your team could figure out a way to deal with the AGPLv3 license https://grafana.com/blog/2021/04/20/grafana-loki-tempo-relicensing-to-agplv3/ which is not encouraged by Google(note that the project for Prow is still under Google).

Our team had migrated to use GKE managed prometheus + Gcloud monitoring solution. @cjwagner , I think we have set this up for Istio right?

jacob-delgado:
What is the purpose of velodrome? I just saw that it was removed in upstream k8s due to a possible security issue.

chaodaiG:
> What is the purpose of velodrome? I just saw that it was removed in upstream k8s due to a possible security issue.

iiuc it's Prow dashboard deprecated in favor of Prometheus + Grafana

howardjohn:
I think it was a prow dashboard implemented by Prometheus and grafana. It
was somewhat useful when it existed but not a big deal

howardjohn:
Actually maybe I don't know what I'm talking about.. I may have been
thinking of monitoring.prow.istio.io. Which is also on old grafana

chaodaiG:
since monitoring.prow.istio.io already exist then this PR LGTM to me

chaodaiG (approved):

PR #4243, opened 2022-08-16
Bump all golang images to latest version

howardjohn:
Older versions do not play well with go.sum updates


kzdv-pipeline:
Should the go.mod also be updated to reference go 1.19?

ericvn (approved):

PR #4259, opened 2022-08-26
Add k8s 1.25.0 testing to master

jacob-delgado:
https://github.com/istio/test-infra/pull/4255/ removed the jobs for 1.25 as 1.24 was missing. Not sure if this was an oversight. cc @howardjohn


howardjohn (commented):
we test 1.25 in presubmit, so we don't need it in postsubmit as well. Although the 1.24 one shouldn't be hidden

howardjohn:
Or rather we do test it in postsubmit already -- all of the non `*k8s*` jobs are 1.25 already

PR #4278, opened 2022-09-08
Run the same tests used for integ-ipv6 for dual-stack k8s clusters

jacob-delgado:
Adding the same level of testing that IPv6-only clusters do, but making it hidden and optional.

See https://github.com/istio/istio/pull/40634


howardjohn (approved):

zhlsunshine:
Relate to [issue#40394](https://github.com/istio/istio/issues/40394)

PR #4282, opened 2022-09-09
Do not report status of post-submit integration failures to slack

jacob-delgado:
The integ-ds postsubmit job is failing on the prow config sets it to report by default. This is an attempt to silence that until we can get it working. Since the map it creates for the `report_config.slack` is empty it's possible that we will need to make another change.

From https://github.com/kubernetes/test-infra/blob/master/prow/cmd/crier/README.md#slack-reporter:

> To silence notifications at the ProwJob level you can pass an empty slice to reporter_config.slack.job_states_to_report: postsubmits:

```yaml
  some-org/some-repo:
    - name: example-job
      decorate: true
      reporter_config:
        slack:
          job_states_to_report: []
      spec:
        containers:
          - image: alpine
            command:
              - echo
```

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -531,8 +531,7 @@ postsubmits:
     name: integ-ds_istio_postsubmit_pri
     path_alias: istio.io/istio
     reporter_config:
-      slack:
-        report: false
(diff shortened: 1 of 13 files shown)


jacob-delgado:
Not sure this will work, see https://github.com/kubernetes/test-infra/blob/f15829e24edce7451b15400f22ca0c88cff7fa21/prow/config/config.go#L2325

cc @chaodaiG

PR #4339, opened 2022-10-18
Generate Prow jobs for Istio 1.16

dhawton:
Generated using:

go run tools/prowgen/cmd/prowgen/main.go branch 1.16
go run ./tools/generate-transform-jobs/main.go branch 1.16


ericvn:
So the issue with this PR is that the images in the new yamls is still a `master` image. It should have been renamed to `release-1.16`. Hopefully, https://github.com/istio/test-infra/pull/4340 should fix that. However, this should cause us to go down the path where a `cloud` command is needed to retag an image in gcr: https://github.com/istio/test-infra/blob/master/tools/prowgen/cmd/prowgen/main.go#L114. This is the reason a Googler, or someone who has creds to the repository, needs to run it. In the past, I've also just changed that line to an echo locally, and feed that cloud command to someone with creds to run, if I can't find someone to run the whole thing.

ericvn (approved):

PR #4356, opened 2022-10-22
testgrid: dashboards must belong to dashboard groups

spiffxp:
The goal here is to prevent istio from littering testgrid.k8s.io with a bunch of top-level dashboards. 
Screen Shot 2022-10-21 at 5 21 48 PM

We follow a convention there of placing all dashboards under a group (see all the sig-foo groups at the bottom)

This PR adds a minimal version of [the testgrid config test used in kubernetes/test-infra](https://github.com/kubernetes/test-infra/blob/master/config/tests/testgrids/config_test.go). It verifies that testgrid config parses correctly, and then verifies the following conventions:

- dashboard and dashboard group names must start with "istio"
- dashboards must belong to a dashboard group
- dashboards should be in the group with the longest common prefix

With the test in place, the config failed, so the config was then modified to pass the test:

- add istio_release-1.16 group
- move other dashboards to appropriate groups

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module istio.io/test-infra
 go 1.18
 
 require (
+	github.com/GoogleCloudPlatform/testgrid v0.0.123
 	github.com/ghodss/yaml v1.0.0
 	github.com/google/go-cmp v0.5.8
 	github.com/google/go-containerregistry v0.8.1-0.20220216220642-00c59d91847c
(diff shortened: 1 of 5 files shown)


spiffxp:
/cc @dhawton 
https://github.com/istio/test-infra/pull/4339 is what caused me to scratch this itch

Probably the better long term fix is to update the prowjob generation tools you use to also autogenerate the testgrid config, but this is what I had time for.

howardjohn (approved):

spiffxp:
Thanks! Looks much better now

Screen Shot 2022-10-27 at 11 14 34 AM

PR #4357, opened 2022-10-22
Add new cron job for updating envoyproxy/go-control-plane

jacob-delgado:
In the process of adding this script

### DIFF
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
@@ -91,7 +91,8 @@ periodics:
       - --labels=auto-merge,release-notes-none
       - --modifier=update_go_control_planey_dep
       - --token-path=/etc/github-token/oauth
-      - --cmd=UPDATE_BRANCH=main bin/update_go-control-plane.sh
+      - --cmd=go get github.com/envoyproxy/envoy/go-control-plane@main && go mod tidy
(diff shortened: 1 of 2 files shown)


jacob-delgado:
cc @ericvn Modeled this after https://github.com/istio/test-infra/blob/a0fe48a9a8f10784b232fec6c53187cfea3b273b/prow/config/jobs/api.yaml#L23

howardjohn:
Is there something specific about g-c-p? Should we just have something
update all dependencies?

jacob-delgado:
If we could figure out a way to do all does that'd be ideal. I would say updating gcp is more vital to the project since it affects things like envoyfilter.

What do you use to update all of the dependencies? I use some hacky bash to update them.

howardjohn:
go-mod-upgrade but it's interactive - not sure it can be done automatically?

There is of course dependabot but it was kind of annoying when I look ~2
years ago. Not sure if it's improved since then

jacob-delgado:
I've used dependabot. It works works well for monorepos, but would probably creates more problems than it solves for istio.

howardjohn:
regardless of what tool I think once a week updating all deps would be
cool. I am sure we can do it - happy to start with this though. Maybe we
can just make the job "update deps" naming though so we can extend it
though?

jacob-delgado:
ping @howardjohn

howardjohn (approved):

PR #4373, opened 2022-11-02
Rotate test-infra-trusted build cluster kubeconfig

chaodaiG:
The test-infra-trusted cluster root CA is expiring soon, which will need to be rotated, and the old kubeconfig for Prow service cluster to authenticate with this build cluster will be invalid after the rotation. So a new kubeconfig is needed to avoid breakage of the connection to this build cluster.

Due to security reasons GKE doesn't allow long lived tokens any more, and the newly generated tokens are at most valid for 2 days, so there needs a way to rotate the token. This PR follows the strategy adopted by k8s prow

### DIFF
--- a/prow/cluster/gencred-config/gencred-config.yaml
+++ b/prow/cluster/gencred-config/gencred-config.yaml
@@ -1,0 +1,8 @@
+clusters:
+# test-infra-trusted build cluster(aka the test-pods ns on the prow service cluster)
+- gke: projects/istio-testing/locations/us-west1-a/clusters/prow
+  name: test-infra-trusted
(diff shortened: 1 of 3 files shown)


listx:
/lgtm
/hold

For additional reference behind how gencred works, see https://docs.prow.k8s.io/docs/build-clusters/.

howardjohn (approved):
RIP forever tokens

chaodaiG:
/unhold
note that these two new jobs will fail until the SA is WI binded and the new GCP SA is given the secret manager permission in the GCP project

listx (approved):

PR #4376, opened 2022-11-02
Add new test-infra-trusted build cluster kubeconfig to prow deployments

chaodaiG:
The `test-infra-trusted` kubeconfig entry will be duplicated once this PR is merged, both of them were generated from the old root CA, and both work.

/hold
The kubeconfig secret was synced to the wrong ns in #4374 , will fix before this PR can be merged


howardjohn (approved):

chaodaiG:
The secret was moved to default ns by #4377 and confirmed that the kubeconfig secret exists
/unhold

listx (approved):

PR #4426, opened 2022-12-02
Add break glass make target when test-infra-trusted cluster kubeconfi…

chaodaiG:
…g is out of sync

This is a chicken and egg problem. The job that rotates the kubeconfig secrets runs on test-infra-trusted build cluster, which is also rotated by the job itself, so when it's out of sync it won't auto-heal. Oncall can run this make target to break glass.

### DIFF
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -97,4 +97,17 @@ create-release-deps-configmap: get-private-cluster-credentials
 
 create-deps-configmaps: create-istio-deps-configmap create-release-deps-configmap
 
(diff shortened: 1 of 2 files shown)


chaodaiG:
note that these are from https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/oss/Makefile

michelle192837 (approved):
LGTM aside from the lint issues.

howardjohn:
gencred is no longer used

PR #4462, opened 2022-12-29
Fix cmd for update deps periodic jobs

jacob-delgado:
attempt to fix automated failure, see https://prow.istio.io/view/gs/istio-prow/logs/update-deps-istio_istio_periodic/1606832164815507456


jacob-delgado:
After some sleuthing, I found out that the repos needed to be changed. This is similar to https://github.com/istio/test-infra/blob/55336416241e7908d59bcf5bad220bc2d84ae943/prow/config/jobs/api.yaml#L25

prow/config/jobs/istio.yaml
```diff
@@ -431,7 +431,7 @@ jobs:
     - --token-path=/etc/github-token/oauth
     - --cmd=go get github.com/envoyproxy/go-control-plane@main && go mod tidy && make clean gen
     requirements: [github]
-    repos: [istio/istio@master]
+    repos: [istio/test-infra@master]
```
ericvn: Changing to the test-infra repo is correct since it's needed to run the automator script.

prow/config/jobs/istio.yaml
```diff
@@ -446,7 +446,7 @@ jobs:
     - --token-path=/etc/github-token/oauth
     - --cmd=go get github.com/envoyproxy/go-control-plane@main && go mod tidy && make clean gen
     requirements: [github]
-    repos: [istio/proxy@master]
+    repos: [istio/test-infra@master]
```
ericvn, line 434: Is there a reason to have this as two jobs, and not just have --repo=istio,proxy in the preceding job?

jacob-delgado: fixed

ericvn (approved):

PR #4467, opened 2023-01-04
Automated job to update proxy sha in istio/istio experimental branch

ericvn:
Add an automated job to update the proxy sha in the istio/istio experimental-ambient branch files when the proxy sha is updated.

The current issue is that the support_release_branching flag is at the file level so this job would be created at branch cut time which would cause changes in the new branch of istio/proxy to also update the proxy sha which is not desired. One could simply remove the job from the generated yaml, but that is a manual step (could be automated) that would need to be documented.

### DIFF
--- a/prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.gen.yaml
@@ -182,6 +182,81 @@ postsubmits:
           path: /var/tmp/prow/cache
           type: DirectoryOrCreate
         name: build-cache
+  - annotations:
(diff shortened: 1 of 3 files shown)


ericvn:
Found `disable_release_branching` which will prevent the job from being copied at branch cut.

ericvn:
Add DNM label until the istio/proxy code is updated to work in the experimental branch.

ericvn:
Remove the DNM as the Envoy is now updated.

ericvn:
Discussed in the Ambient call and noted that there is a lack of some testing and that the branch will be going away soon, so not to spend much time on setting this up now.

Because of that, I will close this PR for now.

PR #4468, opened 2023-01-05
Deploy common-files update to the experimental-ambient branch

ericvn:


ericvn:
DNM since we need some changes cherry-picked into the experimental-ambient branch for the go-sec updates in the common files.

ericvn:
Changes merged into the experimental-ambient branch so removing the DNM.

ericvn:
Discussed in the Ambient call and desire would be do to this more randomly than on each post-submit and also that the branch will be going away soon, so not to spend much time on setting this up now.

Because of that, I will close this PR for now.

PR #4483, opened 2023-01-13
benchmark: add back the benchmark job

poussa:
Add back the benchmarking jobs (start with single job) that were remove on  https://github.com/istio/test-infra/pull/3943

Contributes to: https://github.com/istio/istio/issues/41609

### DIFF
--- a/prow/cluster/jobs/all-periodics.yaml
+++ b/prow/cluster/jobs/all-periodics.yaml
@@ -28,3 +28,44 @@ periodics:
         - https://gcsweb.istio.io/gcs/istio-release/
     nodeSelector:
       testing: test-pool
+
+- cron: "0 8 * * *" # starts every day at 08:00AM UTC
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
Where does the benchmark run? AFAIK we shut down the boskos stuff

poussa:
> Where does the benchmark run? AFAIK we shut down the boskos stuff

The benchmark script is [modified ](https://github.com/istio/tools/pull/2312)to create and delete a cluster before and after run. I was thinking of using the `prowjob-advanced-sa` to create the cluster. That part is actually missing from the PR currently.

Would that work? It works in my testing prow env.

VillePihlava:
Depending on the plans for this pull request, https://github.com/istio/tools/pull/2340 should affect this

poussa:
Not working on this anymore.

PR #4509, opened 2023-01-23
community: dismiss stale PR reviews

howardjohn:
This is a trial run

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -274,7 +274,7 @@ branch-protection:
         community:
           branches: *blocked_branches
           required_pull_request_reviews:
-            dismiss_stale_reviews: false
+            dismiss_stale_reviews: true
             dismissal_restrictions:
               teams:
               - steering-committee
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -273,6 +273,12 @@ branch-protection:
           branches: *blocked_branches
         community:
           branches: *blocked_branches
+          required_pull_request_reviews:
+            dismiss_stale_reviews: false
```
ericvn: Just confirming as I thought you mentioned disabling reviews on a code change.  Notes say: `Set to true if you want to automatically dismiss approving reviews when someone pushes a new commit.`

ericvn: Behavior I see implies the default is false, so this would not change that behavior at all.

howardjohn: thanks, meant to be true

ericvn (approved):

howardjohn:
ping @istio/operations-maintainers

chaodaiG (approved):

PR #4519, opened 2023-01-25
Test splitting flakey gateway tests to their own test

ericvn:
In https://github.com/istio/istio.io/pull/12520, I am moving the flakey gtwapi tests to another test group, and this creates the test to run that group.

I did make this hidden and optional as it will fail without the matching PR moving the tests.


johnma14 (approved):
LGTM

PR #4553, opened 2023-02-05
remove update deps from istio 1.17 release

jacob-delgado:
this can get out of sync with go-controlplane and we do not want this automerged


ericvn (approved):

ericvn:
We should also throw a flag to not copy this in the future branching. That can be added when the master job gets updated as it's not working anyway.

PR #4566, opened 2023-02-09
Drop k8s 1.16-1.19 for master (Istio 1.18)

howardjohn:
These are *super* old, dropping support


ericvn (approved):
Per earlier discussion, removal of testing of versions before Kubernetes 1.20.

PR #4640, opened 2023-03-29
Allow org for automator to be configurable in prow jobs

Kmoneal:
Org for automator is currently hardcoded to `istio`. This change will allow the organization to be changed using an environment variable `ORG`.


ericvn:
Likely the reason we see some automator PRs pop up here from time to time. Not that his will directly fix it since they will need to add the overriding env var.

jjtischer:
> Likely the reason we see some automator PRs pop up here from time to time. Not that his will directly fix it since they will need to add the overriding env var.

Actually we did find the root cause of our prow posting PRs and removed it. 
config_updater was miss-configured ;) oops sorry about that

ericvn (approved):

PR #4642, opened 2023-03-29
Adding prow-arm-private config.

stewartbutler:


prow/cluster/private-arm/prow-private-istio-io_managedcertificate.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: networking.gke.io/v1
+kind: ManagedCertificate
```
howardjohn: we shouldn't need this

prow/cluster/private-arm/kubernetes_external_secrets.yaml
```diff
@@ -0,0 +1,6 @@
+# Empty
```
howardjohn, line 1: nit: if its empty do we really need prow/cluster/private-arm/kubernetes-external-secrets_deployment.yaml etc?

prow/cluster/private-arm/release-deps.yaml
```diff
@@ -0,0 +1,257 @@
+# Below contains the configmaps that configure the release-builder manifests in istio/istio and istio/release-builder.
+# This should be updated for each version
+apiVersion: v1
```
howardjohn: not needed, we don't run this job on arm and will make burden to sync it

prow/config.yaml
```diff
@@ -41,6 +41,10 @@ plank:
     config:
       default_service_account_name: "prowjob-default-sa" # Use workload identity
       gcs_credentials_secret: ""                         # rather than service account key secret
+  - cluster: prow-arm-private
+    config:
+      default_service_account_name: "prowjob-default-sa" # Use workload identity
+      gcs_credentials_secret: ""                         # rather than service account key secret
```
howardjohn: should be like the istio-private one

```
      default_service_account_name: "prowjob-private-sa"
      gcs_configuration:
        bucket: "istio-private-build"
        path_strategy: "explicit"
      ssh_key_secrets:
      - ssh-key-secret
```

stewartbutler: So the thing that confuses me is that the istio-private one does not specify a cluster, it specifies a 'repo'. What isn't clear is if that is decorating something else, or it is still applying to a cluster.

prow/cluster/private-arm/kubernetes_external_secrets.yaml
stewartbutler: So I'm copying the convention from `arm`. Should I likewise delete that?

prow/cluster/private-arm/prow-private-istio-io_managedcertificate.yaml
stewartbutler: Nuked.

prow/cluster/private-arm/release-deps.yaml
stewartbutler: Nuked.

jacob-delgado (approved):

prow/config.yaml
howardjohn: weird, need to look into the docs. Regardless prowjob-default-sa seems wrong , seems like we have a private SA?

howardjohn: you can match org or repo (confusing key, it's actually matching the org). https://github.com/kubernetes/test-infra/blob/c66af3fdd561/prow/config/config.go#L721

so we. probably want to remove this entirely

prow/cluster/crier_deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         env:
         # Use KUBECONFIG envvar rather than --kubeconfig flag in order to provide multiple configs to merge.
         - name: KUBECONFIG
-          value: "/etc/kubeconfig/istio-config:/etc/kubeconfig-build-test-infra-trusted/kubeconfig"
+          value: "/etc/kubeconfig/istio-config:/etc/kubeconfig-build-test-infra-trusted/kubeconfig:/etc/kubeconfig-prow-arm-private/kubeconfig"
```
howardjohn, line 38: I don't think we want private here, based on the other private note being there.

prow/cluster/private-arm/kubernetes_external_secrets.yaml
howardjohn: I guess so, maybe we do it later though

prow/cluster/crier_deployment.yaml
howardjohn: Oh the `istio-config` has all of them today, I see

howardjohn (approved):

PR #4643, opened 2023-03-29
Add v1.27.0.rc0 testing

jacob-delgado:


prow/config/jobs/istio.yaml
```diff
@@ -302,6 +302,21 @@ jobs:
...
         value: " --istio.test.retries=1 "
 
+  - name: integ-k8s-127
+    types: [postsubmit]
+    modifiers: [hidden]
+    command:
+      - entrypoint
+      - prow/integ-suite-kind.sh
+      - --node-image
+      - jacodelg/node:v1.27.0-rc.0
```
jacob-delgado: @howardjohn can you either push your own or copy this to gcr.io?

howardjohn: I already did, I had this same PR open but forgot to submit it :-) Should be in GCR

ericvn (approved):
Was thinking we could leave it un-hidden, but let's merge and see how it does.

PR #4675, opened 2023-04-21
1.18 branch cut - step 3

pmerrison:
This PR completes the step 3 instructions for the 1.18 branch cut.  There were two parts of the instructions that didn't make sense to me though, so please review this PR for completeness:

1 - Manually update BASE_VERSION in the istio-1.18.yaml from master to 1.18 - I couldn't find anything called BASE_VERSION to update
2 - Update test-infra code to handle duplicate entries as well as master image changes done manually above for upcoming 1.19 branch cut - I didn't understand what this meant

Other potential nits:

I based this PR off @ericvn's 1.17 version [here](https://github.com/istio/test-infra/pull/4493/files#top).  As such, I have just added the 1.18 content to `testgrid/config.yaml` and it's getting a bit big, maybe we should remove the 1.16 ones?


howardjohn (approved):

ericvn:
I do notice some duplication of env var and image fields. This shouldn't really have any effect, other than making the files bigger and a little tougher to change the image used by a file as you'll need to do it across all jobs. If you do remove them, be careful to check the images as some jobs are meant to have them to override the default (mainly in proxy, and I have been burnt here before (they want tools-proxy but it was removed and uses tools instead by mistake)).

Other than that, looks like a few dashboard changes.

PR #4676, opened 2023-04-21
Allow configurable automator org

Kmoneal:
Allow automator prow jobs to use organization environment variable and update git clone to use username and token for private repos. This PR was missing the automator portion to not just use `$ORG` as the org name.


jacob-delgado:
This may make it easier for us to also use for `istio-private` too. Thoughts @ericvn ?

ericvn:
@Kmoneal I think this looks good, and I think I know why you added the `Fix dry run git command` commit, but can you articulate that here. Thanks.

Kmoneal:
`Fix dry run git command` was to enable users to not have to configure github tokens locally but if this is a concern. I can switch it to use tokens as well and add instructions on how to do the dry-run

ericvn (approved):

PR #4684, opened 2023-05-01
Fix go control plane update from updating in release-1.18 branch

jacob-delgado:


howardjohn (approved):
LGTM, do we need to update the branching behavior?

ericvn (approved):

ericvn:
> LGTM, do we need to update the branching behavior?

Coming in another PR.

PR #4718, opened 2023-05-16
Use the make target instead of calling the release scripts directly

jwendell:
This will allow the script to reuse the `VERSION` variable defined in the Makefile.


jwendell:
No longer needed as of https://github.com/istio/istio/pull/44965

PR #4763, opened 2023-06-15
jobs: deny usage of underscore in name

howardjohn:
This is to enforce consistency in job names. Underscore is used as a
seperator for auto-generated suffixes, so using it in the base name is
problematic.


ericvn (approved):
Thanks.

PR #4769, opened 2023-06-15
Terraform: better document GCS buckets and cleanup empty ones

howardjohn:
These buckets are unused and empty so have been removed. Also add a
table so we can have a single place to reference for all the storage
going on in the project. Currently with it spread everywhere its hard to
grok.


ericvn (approved):
Thanks for the added documentation.

PR #4777, opened 2023-06-21
Benchmark Ztunnel Command

Stevenjin8:
Still WIP. This PR adds a command to run benches which is in https://github.com/istio/ztunnel/pull/561. I'm not sure how to upload artifacts.

I want to have the html generated by criterion be an artifact. Do I ust need to add it to /logs/artifacts/ and it will be automatically uploaded to a GCP bucket?

Also, I can't find a way to create a custom response that shows the results of the tests.

### DIFF
--- a/prow/cluster/jobs/istio/ztunnel/istio.ztunnel.master.gen.yaml
+++ b/prow/cluster/jobs/istio/ztunnel/istio.ztunnel.master.gen.yaml
@@ -52,6 +52,49 @@ postsubmits:
     branches:
     - ^master$
     decorate: true
+    name: bench_ztunnel_postsubmit
+    path_alias: istio.io/ztunnel
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - command:
+        - ./scripts/benchtest.sh
+        env:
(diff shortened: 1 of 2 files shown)


howardjohn (approved):
LGTM. I was thinking it would be good to merge `./scripts/benchtest.sh` first but I guess this is optional, so merging early may help you test out the script. So lets merge now

MorrisLaw:
> LGTM. I was thinking it would be good to merge `./scripts/benchtest.sh` first but I guess this is optional, so merging early may help you test out the script. So lets merge now

My recent PR to ztunnel fails because the script does not exist yet: https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_ztunnel/571/bench_ztunnel/1673373264027586560

howardjohn:
Ah, this should be presubmit_skipped as well so its run only on demand. @MorrisLaw in the meantime it will _not_ block your PR - its optional, even though its always a bit hard to tell that in the UI.

Stevenjin8:
@howardjohn I am working on the fix.

PR #4782, opened 2023-06-21
prow: migrate from gencred to gke-gcloud-auth-plugin/WI

howardjohn:
Currently, we have a job the runs every so often and updates our kubeconfig Secrets with a new token. This is not ideal from a security and reliability POV.

This PR migrates off of this onto `gke-gcloud-auth-plugin` which uses workload identity in the prow control plane components to request access tokens on-demand.

This PR:
* Adds the new kubeconfig configmap. It has no private info so its no longer a Secret
* Removes the old ExternalSecrets, gencred job, gencred config
* Updates the prow control plane to reference the new configs

I did a dry run of this by updating one of the clusters to this method to validate permissions are setup correctly, etc.


listx:
I don't have much to say wrt the config changes (moving away from gencreds sounds like a great idea to me).

One nit I have is that it would be safer to make the changes to 1 component's deployment file at a time (or at least, to just 1 component initially) to help spread the risk over multiple deployments (instead of just 1 big change pushed out to all deployments).

prow/cluster/build_kubeconfig.yaml
```diff
@@ -0,0 +1,101 @@
+# Kubeconfig provides access to each of our Kubernetes build clusters.
```
jjtischer, line 1: This is a really nice change, I like the single KUBECONFIG
You might consider documenting how someone might generate/update these credentials. If a cluster has to get rebuilt or another one added.

jjtischer (approved):

ericvn (approved):

jacob-delgado (approved):

stewartbutler (approved):

PR #4783, opened 2023-06-21
prow: add highlighter config

howardjohn:
Motivation: we currently use `FAIL:` in logs to trigger the highlight,
when the line did *not* cause a failure. This allows us to highlight
without faking log lines. It may also be useful for future things we
want to add, now that we have the basis for it.

This should be the defaults + the SUSPICIOUS part. See
https://github.com/kubernetes/test-infra/blob/e7ccf79861d3c53dfaa6c784c0c8537922c4c0cc/prow/spyglass/lenses/buildlog/lens.go#L95
for the defaults.

Note this overrides the defaults, so we need to include the default
parts.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -68,8 +68,8 @@ deck:
           highlight_regexes:
             - "timed out"
             - "ERROR:"
-            - "(FAIL|Failure \[)\b"
-            - "panic\b"
+            - "(FAIL|Failure \\[)\\b"
+            - "panic\\b"
             - "SUSPICIOUS:"
       required_files:
       - build-log.txt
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -64,6 +64,13 @@ deck:
       - podinfo.json
     - lens:
         name: buildlog
+        config:
+          highlight_regexes:
+            - "timed out"
+            - "ERROR:"
+            - "(FAIL|Failure \[)\b"
```
cjwagner: IIRC you need to escape `\` when using quoted strings. So this should be 
```suggestion
            - "(FAIL|Failure \\[)\\b"
```

howardjohn: Thanks! that is probably why the job failed too

cjwagner (approved):

PR #4784, opened 2023-06-21
Setup job configuration to allow reading GCP secrets directly

howardjohn:
Pairs with https://github.com/istio/tools/pull/2526, see that PR for
more context.

This adds a new preset type of secrets that will be fetched at job
startup. Currently, no jobs use this - but the `github` and `release`
presets will be migrated to this, eventually.

### DIFF
--- a/prow/config/jobs_test.go
+++ b/prow/config/jobs_test.go
@@ -239,7 +239,7 @@ func TestJobs(t *testing.T) {
 		}
 
 		if j.Type == Presubmit {
-			return fmt.Errorf("jobs with secrets must cannot be presubmits")
+			return fmt.Errorf("jobs with secrets cannot be presubmits")
 		}
 		if !hasEntrypoint {
 			return fmt.Errorf("jobs with secrets must use entrypoint")
--- a/prow/config/jobs_test.go
+++ b/prow/config/jobs_test.go
@@ -220,6 +220,32 @@ func TestJobs(t *testing.T) {
 		}
 		return nil
 	})
+
+	RunTest("secret access", func(j Job) error {
+		secrets := false
+		hasEntrypoint := false
+		for _, c := range j.Base.Spec.Containers {
+			if len(c.Command) > 0 && c.Command[0] == "entrypoint" {
+				hasEntrypoint = true
+			}
+			for _, e := range c.Env {
+				if e.Name == "GCP_SECRETS" {
+					secrets = true
+				}
+			}
+		}
+		if !secrets {
+			return nil
+		}
+
+		if j.Type == Presubmit {
+			return fmt.Errorf("jobs with secrets must cannot be presubmits")
+		}
+		if !hasEntrypoint {
+			return fmt.Errorf("jobs with secrets must use entrypoint")
+		}
+		return nil
+	})
 }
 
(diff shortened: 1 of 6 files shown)


prow/config/jobs_test.go
```diff
@@ -220,6 +220,32 @@ func TestJobs(t *testing.T) {
...
+					secrets = true
+				}
+			}
+		}
+		if !secrets {
+			return nil
+		}
+
+		if j.Type == Presubmit {
+			return fmt.Errorf("jobs with secrets must cannot be presubmits")
```
ericvn: nit:
```suggestion
			return fmt.Errorf("jobs with secrets cannot be presubmits")
```

tools/prowgen/pkg/decorator/requirement.go
```diff
@@ -62,6 +63,29 @@ func ApplyRequirements(job *config.JobBase, requirements, excludedRequirements [
...
+	if len(secrets) == 0 {
+		return
+	}
+	marshal, err := json.Marshal(secrets)
+	if err != nil {
+		log.Fatalf("failed to marshal secrets: %v", err)
+	}
+	if len(job.Spec.Containers) != 1 {
+		// We could support more but it may expand permissions, just keep it safe for now
+		log.Fatalf("secrets only works with 1 container")
```
ericvn: nit:
```suggestion
		log.Fatalf("secrets only work with 1 container")
```

ericvn (approved):
Set DNM in case you want to fix the nits. Feel free to remove.

tools/prowgen/pkg/testdata/.base.yaml
```diff
@@ -105,3 +105,8 @@ requirement_presets:
     - "common"
     - "args"
     - "that are reusable"
+  secrets:
+    secrets:
```
jwendell, line 109: two levels?

tools/prowgen/pkg/decorator/requirement.go
```diff
@@ -62,6 +63,29 @@ func ApplyRequirements(job *config.JobBase, requirements, excludedRequirements [
...
+	secrets := []spec.Secret{}
+	for _, req := range presets {
+		secrets = append(secrets, req.Secrets...)
+	}
+	if len(secrets) == 0 {
+		return
+	}
+	marshal, err := json.Marshal(secrets)
+	if err != nil {
+		log.Fatalf("failed to marshal secrets: %v", err)
```
jwendell, line 79: return?

prow/config/jobs_test.go
```diff
@@ -220,6 +220,32 @@ func TestJobs(t *testing.T) {
...
+		}
+		if !hasEntrypoint {
+			return fmt.Errorf("jobs with secrets must use entrypoint")
+		}
+		return nil
+	})
```
jwendell, line 248: do you want to test if the pod has more than 1 container too?

tools/prowgen/pkg/testdata/.base.yaml
howardjohn: The first one is the requirement name, in non-test it will be something like `release:`. But probably could have better naming

PR #4787, opened 2023-06-22
terraform: add initial scoped service accounts

howardjohn:
This gets us ready to use more fine-scoped service accounts using WI to restrict access to only minimal set of resources required. This PR:

* Adds a new module to setup WI, heavily copied from Kubernetes with some tweaks for our own needs.
* Adds 2 new accounts, one for release jobs and one for RBE.
* The release permissions are currently a small stub so we can do some testing on fake secrets, after testing will give it the real ones. The RBE one _should_ be sufficient, but I will do some manual testing before moving those jobs over

### DIFF
--- a/infra/gcp/istio-prow-build/iam.tf
+++ b/infra/gcp/istio-prow-build/iam.tf
@@ -24,3 +24,17 @@ resource "google_service_account" "prow_internal_storage" {
   display_name = "Prow Internal Storage"
   project      = "istio-prow-build"
 }
+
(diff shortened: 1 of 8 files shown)


ericvn (approved):
Thanks. Lots of work across a number of PRs.

stewartbutler (approved):

howardjohn:
`Apply complete! Resources: 8 added, 0 changed, 0 destroyed`

PR #4788, opened 2023-06-22
Remove cosign 'secret'

howardjohn:
This is not a secret, its just a simple value.

To keep things simple, we remove the secret and inline it.

This is part of an effort to remove all secrets from the cluster. While
this one technically is fine, I think it causes confusion that it is
confidential when it isn't.


ericvn:
While this looks good to me, just wondering if @stewartbutler has any concerns.

ericvn (approved):

stewartbutler:
Nah, looks good to me -- I just put it there for the sake of matching convention elsewhere. Approving.

stewartbutler (approved):

PR #4790, opened 2023-06-22
prow: move deck-private to use WI

howardjohn:
Will do this after https://github.com/istio/test-infra/pull/4782 to avoid too much going on at once

### DIFF
--- a/prow/cluster/deck_private_rbac.yaml
+++ b/prow/cluster/deck_private_rbac.yaml
@@ -1,6 +1,8 @@
 kind: ServiceAccount
 apiVersion: v1
 metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: prow-control-plane@istio-testing.iam.gserviceaccount.com
   namespace: default
   name: deck-private
 ---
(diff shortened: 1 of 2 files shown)


howardjohn:
ping @stewartbutler when you have a chance

prow/cluster/deck_private_rbac.yaml
```diff
@@ -1,6 +1,8 @@
 kind: ServiceAccount
 apiVersion: v1
 metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: prow-control-plane@istio-testing.iam.gserviceaccount.com
```
jjtischer, line 5: make sure you check/allow prow-control-plane workload identity mapping for each service that needs it.
```
  workload_members = [
    "serviceAccount:${local.project_id}.svc.id.goog[default/crier]",
    "serviceAccount:${local.project_id}.svc.id.goog[default/deck]",
    "serviceAccount:${local.project_id}.svc.id.goog[default/hook]",
    "serviceAccount:${local.project_id}.svc.id.goog[default/prow-controller-manager]",
    "serviceAccount:${local.project_id}.svc.id.goog[default/tide]"
  ]
```

stewartbutler (approved):

prow/cluster/deck_private_rbac.yaml
howardjohn: Good call. I am adding this to terraform

PR #4797, opened 2023-06-22
terraform: fix new WI SAs to have write access to prow logs

howardjohn:
This is required for prowjobs to run

### DIFF
--- a/infra/gcp/modules/workload-identity-service-account/main.tf
+++ b/infra/gcp/modules/workload-identity-service-account/main.tf
@@ -34,7 +34,9 @@ resource "google_service_account" "serviceaccount" {
 }
 data "google_iam_policy" "workload_identity" {
   binding {
-    members = ["serviceAccount:${local.cluster_project_id}.svc.id.goog[${var.cluster_namespace}/${local.cluster_serviceaccount_name}]"]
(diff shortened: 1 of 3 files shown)


jacob-delgado (approved):

infra/gcp/modules/workload-identity-service-account/main.tf
```diff
@@ -50,3 +52,10 @@ resource "google_project_iam_member" "project_roles" {
   role     = each.value.role
   member   = "serviceAccount:${google_service_account.serviceaccount.email}"
 }
+// If this is going to be used for prowjobs, then we need to give it access to gs://istio-prow to write logs.
+resource "google_storage_bucket_iam_member" "member" {
+  count  = var.prowjob ? 1 : 0
+  bucket = "istio-prow"
+  role   = "roles/storage.objectAdmin"
```
jjtischer, line 59: this should be sufficient
"roles/storage.legacyBucketWriter"

howardjohn: I thought storage.objects.update was needed a well? and legacyBucketWriter only has {create,delete,list}?

PR #4819, opened 2023-06-26
Ztunnel benches only run on demand

Stevenjin8:
Fixes #4777 so that benchmark tests only run on demand


howardjohn (approved):
/ok-to-test

thanks1

Stevenjin8:
I think these are the only options needed because the istio benchmarks only have these two options https://github.com/istio/test-infra/blob/5dbd49128b0d642c43a15760d04d53e254087d6a/prow/config/jobs/istio.yaml#L26.

PR #4828, opened 2023-06-27
Benchmarking Ztunnel is a presubmit job

Stevenjin8:
Should fix failing postsubmit CI on ztunnel by making benches presubmit only.


Stevenjin8:
Even if the benches were merged, the script would have failed because of missing environment variables.

howardjohn (approved):

PR #4835, opened 2023-06-27
release-builder: use files instead of env

howardjohn:
Env is a bit prone to leaking accidentally, such as
https://prow.istio.io/view/gs/istio-prow/logs/build-release_release-builder_postsubmit/1673714939367264256#

Needs https://github.com/istio/tools/pull/2541


jacob-delgado (approved):
approved, but added DNM to wait for istio/tools repo changes

PR #4838, opened 2023-06-27
private: remove automator jobs which we don't have permission for

howardjohn:


prow/config/istio-private_jobs/api-1.16.yaml
```diff
@@ -17,8 +17,9 @@ transforms:
   - presubmit
   labels:
     preset-enable-ssh: "true"
-- job-denylist:
-  - update_api_dep_release-1.16
+- job-allowlist:
+  - build
+  - gencheck
```
ericvn, line 22: I notice the release flavor of the jobs are noted above. Is the non-release version needed as well?

howardjohn: Its a regex so no, doesn't need to match the full thing. I think the full version is used only because its auto generated

ericvn (approved):

PR #4839, opened 2023-06-27
terraform: create new istio-prow-private GCP account

howardjohn:
Contains https://github.com/istio/test-infra/pull/4838; PR will be smaller once that lands and I rebase.

Currently, our infra for private releases is tangled with istio-prow-build.

This is not great from an understanding perspective, security, etc.

This PR creates a new project from the ground up to replace the private components

The following build locations are changed:
* gcr.io/istio-prow-build => New Artifact Registry
* gs://istio-private-build => gs://istio-build-private
* gs://istio-private-prerelease => gs://istio-prerelease-private

There were a few others used in the old private process, but they are all obsolete and will not be carried forward.

GCP Secret access is not needed; this is only used for docker/grafana/github, which the private infra does not need. This is another benefit - I can say with confidence these are NOT used once we split this; today its much fuzzier.

RBE is disabled in private, so no permissions needed there.

Followups needeD:
* Point prow to the new cluster, of course
* Add humans to GCS reader access. Ideally a Google Group for PSWG, though.
* Manually set arm64 node pool to cgroupsv1, which terraform cannot handle currently.
* Grant prow control plane access to the cluster
* Setup the Kubernetes YAMLs for the cluster. This is just the namespace, service account, and release-deps.yaml (I think)

cc @jacob-delgado @stewartbutler @jjtischer @ericvn

### DIFF
--- a/infra/gcp/istio-prow-private/ar.tf
+++ b/infra/gcp/istio-prow-private/ar.tf
@@ -1,0 +1,29 @@
+resource "google_artifact_registry_repository" "main" {
+  location      = "us"
+  repository_id = "istio-prow-private"
+  description   = "registry to host private Istio release artifacts"
+  format        = "DOCKER"
+
+  docker_config {
+    immutable_tags = true
+  }
+
+  lifecycle {
+    prevent_destroy = true
+  }
+}
+
+data "google_iam_policy" "artifact_registry" {
+  binding {
+    role    = "roles/artifactregistry.createOnPushWriter"
+    members = [
+      "serviceAccount:${module.prowjob_private_account.email}",
+    ]
+  }
+}
+resource "google_artifact_registry_repository_iam_policy" "policy" {
+  project     = google_artifact_registry_repository.main.project
+  location    = google_artifact_registry_repository.main.location
+  repository  = google_artifact_registry_repository.main.name
+  policy_data = data.google_iam_policy.artifact_registry.policy_data
+}
--- a/infra/gcp/istio-prow-private/ar.tf
+++ b/infra/gcp/istio-prow-private/ar.tf
@@ -15,7 +15,7 @@ resource "google_artifact_registry_repository" "main" {
 
 data "google_iam_policy" "artifact_registry" {
   binding {
-    role    = "roles/artifactregistry.createOnPushWriter"
+    role = "roles/artifactregistry.createOnPushWriter"
     members = [
       "serviceAccount:${module.prowjob_private_account.email}",
     ]
--- a/infra/gcp/istio-prow-private/cluster.tf
+++ b/infra/gcp/istio-prow-private/cluster.tf
@@ -1,0 +1,76 @@
+module "prow_cluster" {
+  source       = "../modules/gke-cluster"
+  cluster_name = "prow"
+  project_name = local.project_id
+  # ARM node pools are (currently) only available here, so ensure we use this region.
+  cluster_location = "us-central1-f"
+}
+module "prow_node_test" {
+  source = "../modules/gke-nodepool"
+  name   = "test"
+
+  project_name = local.project_id
+  cluster_name = module.prow_cluster.cluster.name
+  location     = module.prow_cluster.cluster.location
+
+  machine_type  = "e2-standard-16"
+  initial_count = 1
+  min_count     = 1
+  max_count     = 15
+
+  disk_size_gb = 256
+  disk_type    = "pd-ssd"
+
+  service_account = module.prow_cluster.cluster_node_sa.email
+
+  labels = {
+    "testing" = "test-pool"
+  }
+}
+module "prow_node_build" {
+  source = "../modules/gke-nodepool"
+  name   = "build"
+
+  project_name = local.project_id
+  cluster_name = module.prow_cluster.cluster.name
+  location     = module.prow_cluster.cluster.location
+
+  machine_type  = "n1-highmem-64"
+  initial_count = 0
+  min_count     = 0
+  max_count     = 10
+
+  disk_size_gb = 2000
+  disk_type    = "pd-ssd"
+
+  service_account = module.prow_cluster.cluster_node_sa.email
+
+  labels = {
+    "testing" = "build-pool"
+  }
+}
+module "prow_node_arm" {
+  source = "../modules/gke-nodepool"
+  name   = "arm"
+
+  project_name = local.project_id
+  cluster_name = module.prow_cluster.cluster.name
+  location     = module.prow_cluster.cluster.location
+
+  machine_type  = "t2a-standard-16"
+  initial_count = 2
+  max_count     = 60
+  min_count     = 2
+
+  disk_size_gb = 256
+  disk_type    = "pd-ssd"
+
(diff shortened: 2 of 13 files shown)


infra/gcp/istio-prow-private/cluster.tf
```diff
@@ -0,0 +1,75 @@
+module "prow" {
```
jjtischer: There are lots of things named "prow". might i suggest something else like 
- prow-cluster
- prow-gke-cluster
- prow-primary

infra/gcp/istio-prow-private/cluster.tf
```diff
@@ -0,0 +1,75 @@
...
+  initial_count = 2
+  max_count     = 60
+  min_count     = 2
+
+  disk_size_gb = 256
+  disk_type    = "pd-ssd"
+
+  service_account = module.prow.cluster_node_sa.email
+
+  spot = true
```
jjtischer, line 78: why spot instances here but not on the build-pool?

infra/gcp/istio-prow-private/ar.tf
```diff
@@ -0,0 +1,25 @@
+resource "google_artifact_registry_repository" "main" {
+  location      = "us"
+  repository_id = "istio-prow-private"
+  description   = "registry to host private Istio release artifacts"
+  format        = "DOCKER"
+
+  docker_config {
+    immutable_tags = true
+  }
```
jjtischer, line 9: all repos and buckets should have prevent destroy. this will prevent a terraform apply or delete from removing the resource
```suggestion
  }
    lifecycle {
    prevent_destroy = true
  }
```

infra/gcp/istio-prow-private/providers.tf
```diff
@@ -0,0 +1,17 @@
+terraform {
+  backend "gcs" {
+    bucket = "istio-terraform"
+    prefix = "istio-prow-private" // project name
+  }
+
```
jjtischer, line 6: what version of terraform?

infra/gcp/istio-prow-private/storage.tf
```diff
@@ -0,0 +1,16 @@
+# Mirrors the gs://istio-build bucket
+resource "google_storage_bucket" "istio_build_private" {
+  name          = "istio-build-private"
+  location      = "US"
+  storage_class = "STANDARD"
+
+  uniform_bucket_level_access = true
```
jjtischer, line 7: I recommend looking into retention policies, because I dont think buckets can use the TF lifecycle prevent delete

infra/gcp/modules/gke-cluster/outputs.tf
```diff
@@ -13,6 +13,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+output "cluster" {
+  description = "The cluster"
+  value       = google_container_cluster.cluster
+}
```
jjtischer, line 19: You might consider outputting ca_certificate, and endpoint for the GKE cluster so you can setup helm or kubectl from in TF later on.

prow/config/istio-private_jobs/release-builder-1.16.yaml
```diff
@@ -34,6 +34,7 @@ transforms:
     GCS_BUCKET: istio-private-prerelease/prerelease
     HELM_BUCKET: istio-private-build/dev/charts
     PRERELEASE_DOCKER_HUB: gcr.io/istio-prow-build
+    GCP_SECRETS: null # Don't use their secrets. We only need to push to GCS and GCR which do not need secrets
```
jjtischer: I would expect release-builder to require github creds to fetch repos and dependencies

infra/gcp/modules/gke-cluster/outputs.tf
howardjohn: I _think_ they can access from `outputs.cluster.xyz`? Not sure though

prow/config/istio-private_jobs/release-builder-1.16.yaml
howardjohn: Good call, forgot about that. Looks like its setup to use `ssh-key-secret`. This is handled by prow's containers so we cannot use the GCP_SECRETS stuff. Not a big deal though, we can use a secret

infra/gcp/istio-prow-private/cluster.tf
howardjohn: will put a comment in the code, but - GCP is only allowing non-trivial quotas for t2a nodes using spot instances

infra/gcp/istio-prow-private/storage.tf
howardjohn: I think you cannot delete a bucket with items unless you explicitly allow it:

```
[force_destroy](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#force_destroy) - (Optional, Default: false) When deleting a bucket, this boolean option will delete all contained objects. If you try to delete a bucket that contains objects, Terraform will fail that run.
```

prow/config/istio-private_jobs/release-builder-1.16.yaml
howardjohn: Still thinking about how to handle this best. I will handle in a followup if necessary, I am sure we will need a bit of iteration before this is working end to end

infra/gcp/istio-prow-private/cluster.tf
```diff
@@ -0,0 +1,76 @@
...
+  source = "../modules/gke-nodepool"
+  name   = "test"
+
+  project_name = local.project_id
+  cluster_name = module.prow_cluster.cluster.name
+  location     = module.prow_cluster.cluster.location
+
+  machine_type  = "e2-standard-16"
+  initial_count = 1
+  min_count     = 1
```
stewartbutler, line 19: I'd let it scale down to zero, the private build infra is very infrequently used.

infra/gcp/istio-prow-private/cluster.tf
```diff
@@ -0,0 +1,76 @@
...
+  name   = "arm"
+
+  project_name = local.project_id
+  cluster_name = module.prow_cluster.cluster.name
+  location     = module.prow_cluster.cluster.location
+
+  machine_type  = "t2a-standard-16"
+  initial_count = 2
+  max_count     = 60
+  min_count     = 2
```
stewartbutler: Likewise -- should we scale this down to zero?

infra/gcp/istio-prow-private/cluster.tf
howardjohn: GKE cannot scale to zero

jjtischer: I think it can, the initial nodes can be removed from the cluster and the node pools can scale to 0. A challenge with this is prow job lifetime, we have our cluster set to remove jobs after 3 days. Jobs being kept will keep nodes alive and then the pool will not scale to zero

stewartbutler (approved):

PR #4857, opened 2023-06-29
Use new proxy override path

howardjohn:
Goes with https://github.com/istio/proxy/pull/4757


ericvn (approved):
Approve and mark DNM so can be merged as the mentioned PR is ready.

PR #4890, opened 2023-07-08
Revert "Update community job to use new tool"

dhawton:
Reverts istio/test-infra#4881 -- fixes failing sync-org postsubmit job: https://prow.istio.io/view/gs/istio-prow/logs/sync-org_community_postsubmit/1677529337726767104

The image that contains the new tool was reverted in #4887, so we may need to revert this to use the old tool until the new image is ready.


dhawton:
Closing since new image with new tool should be done relatively soon

PR #4912, opened 2023-07-18
Add github-readonly token to istio.io gencheck job

ericvn:
https://github.com/istio/istio.io/pull/13595 adds some automation which needs to convert from a short sha to the full sha in the gateway-api repo. This requires the GitHub read-only token.


ericvn:
Seems that we can ignore the sha in that job. Closing for now.

PR #4924, opened 2023-07-25
Remove 1.16 jobs

ericvn:
As there is one 1.16 periodic job that is failing, let's remove the 1.16 jobs. I had originally thought I would only remove the periodic ones, but decided to remove them all.


howardjohn:
Lets merge the 1.16 cve patch in first

jacob-delgado (approved):

PR #4931, opened 2023-07-27
Remove references to envoy repo building

jacob-delgado:
I noticed we haven't updated this in a few years.


jacob-delgado:
Maybe we can just remove this altogether and not do any CI/CD whether istio-private or istio? The building is still done within the proxy repo anyway.

howardjohn (commented):
yeah I don't know that we actually use this.. I certainly did not in the last CVE

jacob-delgado:
Ok. Removed the references for this.

howardjohn (approved):

PR #4949, opened 2023-08-03
Add back script used for tests

ericvn:
This file was removed in #4830 but is used in the enhancement's repo validate-schema job. If we don't want to validate, we need to remove the jobs.

### DIFF
--- a/scripts/validate_schema.sh
+++ b/scripts/validate_schema.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+
+# Copyright Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


ericvn:
See https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_enhancements/154/validate-features_enhancements_release-1.19/1687162980476653568

scripts/validate_schema.sh
```diff
@@ -0,0 +1,103 @@
...
+}
+
+# Curl the GitHub API to get a list of files for the specified PR. If files are
+# found, exit. We might eventually want to validate the data here.
+validate_schema() {
+    echo "Validating ${DOCUMENT_PATH} against ${SCHEMA_PATH}"
+
+    pushd "${REPO_PATH}"
+
+    VALIDATE_SCHEMA_PATH=../tools/cmd/schema-validator
```
howardjohn, line 83: the tool is gone too

ericvn:
As the failing test isn't required and it will require a PR to restore code in the tools repo, I'll delay this until after the branch cut. The test was passing before the files were removed, so it makes sense to add the validation back in some form.

ericvn:
Not stale.

ericvn:
Mentioned by users of istio/enhancements again.

ericvn:
I'm going to close this PR and open a feature request to add the schema validation some other code similar to what was done to remove the tool to begin with.

ericvn:
Issue: https://github.com/istio/istio/issues/48211

PR #4986, opened 2023-08-28
Run pilot tests in dual stack/ipv6 mode and not just security integration tests

jacob-delgado:
Putting in a DNM until we have understand if we want to test this and before any tests are changed in `tests/integration/pilot`.

Follow-up to https://github.com/istio/test-infra/pull/4985#pullrequestreview-1598682451

Right now, IPv6 and Dual-stack integration are only being run on the `tests/integration/security` package. We should have wider coverage to see if/what is broken.

There is a lot of work to be done for either to make sure that both are in a better spot, most largely around unit testing. However, this could help us figure out what to prioritize as there is a lot of work.

With https://github.com/istio/istio/commit/73ddc6b927a67acda8effbe09e98b0004d714659 we saw that 1.19 was broken before release had some of my colleagues not manually tested the RC.

Looking for any feedback. I understand the matrix of different environments is overwhelming, but the pilot tests are probably some of the most important tests we have that can help us further validate IPv6 and dual-stack. I see that we run them for including a different echo image so it's not clear to me when it is appropriate to test things and when it isn't.

See https://github.com/istio/test-infra/blob/8f3388405b3f40c10e5568650f5f866481f5aae5/prow/config/jobs/istio.yaml#L123

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -129,6 +129,26 @@ jobs:
       - name: GRPC_ECHO_IMAGE
         value: "grpctesting/istio_echo_cpp"
 
+  - name: integ-pilot-ipv6
+    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube]
+    requirements: [kind]
+    types: [postsubmit]
+    env:
+      - name: DOCKER_IN_DOCKER_IPV6_ENABLED
+        value: "true"
+      - name: IP_FAMILY
+        value: "ipv6"
+
+  - name: integ-pilot-ds
+    command: [entrypoint, prow/integ-suite-kind.sh, test.integration.pilot.kube]
+    requirements: [kind]
+    types: [postsubmit]
+    env:
+      - name: DOCKER_IN_DOCKER_IPV6_ENABLED
+        value: "true"
+      - name: IP_FAMILY
+        value: "dual"
+
   - name: integ-pilot-multicluster
     command:
       - entrypoint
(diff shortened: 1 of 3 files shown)


stevenctl:
Can you add optional when we first turn it on, and then make it required after we're sure it passes in CI?

jacob-delgado:
This is only for post-submits. We would need to make sure they are working and are not flaky before we turn them on as a presubmit (if we ever do). I'd rather they be a post-submit and maybe for a while they should be `hidden`.

prow/config/jobs/istio.yaml
```diff
@@ -129,6 +129,26 @@ jobs:
       - name: GRPC_ECHO_IMAGE
         value: "grpctesting/istio_echo_cpp"
 
+  - name: integ-pilot-ipv6
```
howardjohn, line 132: I don't think this really does what we want. The tests already run a large subset in presubmit and all tests in postsubmit. I think the DS ones are just broken since we have this logic only in one place: https://github.com/istio/istio/blob/3e8f29a160c06d57dc0436f8d9ce7db72f3de4b1/tests/integration/security/main_test.go#L52

howardjohn:
> 

They are not postsubmit only because they are flaky, but because we cannot have 100s tests running in presubmit. see https://github.com/istio/istio/issues/34387

prow/config/jobs/istio.yaml
kfaseela: @jacob-delgado I think we just need to set the experimental flag in pilot tests, and see how it goes. We are running the tests even now, but without the flag enabled, I guess.

kfaseela:
@jacob-delgado : https://github.com/istio/istio/pull/46766
Let us see if the tests need any modifications/any bug fix needed. Based on that we can take it forward

PR #5049, opened 2023-10-12
Add integration testing of cni in dual stack mode

jacob-delgado:
as a post-submit job. Istio 1.18 and Istio 1.19 were released with CNI failing. See https://github.com/istio/istio/issues/46625


howardjohn (changes requested):
I am not sure we need this.

1. The design of the `environments` test was specifically to make sure we only have 1 test per environment. If its not meeting our needs, we should holistically fix it
2. CNI runs in postsubmit in the integ-ds (and all environment) tests already

jacob-delgado:
But without the integration test flags I wouldn't have thought

PR #5050, opened 2023-10-13
add peer metadata discovery test job

kyessenov:


howardjohn (approved):
Do we know it will pass? If not, we should set presubmit_optional+hidden or presubmit_skipped so we don't block everyone while testing it.

if its already tested then looks fine, feel free to remove hold

kyessenov:
I tested it already. It should pass.

kyessenov:
I'll watch istio to make sure I don't break everyone.

PR #5061, opened 2023-10-19
Drop support for Kubernetes 1.21 and 1.22

howardjohn:
We added startupProbe by default, so these no longer run. IMO its a good
time to drop them, these are super old versions that we likely have no
users of anymore.

Followup PR will update istio.io


linsun (approved):
agree they are super old

PR #5106, opened 2023-12-04
Add dual-stack job for docs test

kfaseela:
Needed for https://github.com/istio/istio.io/pull/14257/files

### DIFF
--- a/prow/config/jobs/istio.io.yaml
+++ b/prow/config/jobs/istio.io.yaml
@@ -50,6 +50,7 @@ jobs:
       - name: IP_FAMILY
         value: "dual"
     regex: ^(go.mod$|tests/|pkg/test/|prow/|content/en/boilerplates/snips/|content/en/docs/.*(test\.sh|/snips\.sh)$)
+    modifiers: [presubmit_optional]
 
   - name: update-ref-docs-dry-run
     types: [presubmit]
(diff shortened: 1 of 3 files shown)


prow/config/jobs/istio.io.yaml
```diff
@@ -41,6 +41,16 @@ jobs:
...
 
+  - name: doc.test.dualstack
+    command: [entrypoint, prow/integ-suite-kind.sh, doc.test.dualstack]
+    requirements: [kind]
+    env:
+      - name: DOCKER_IN_DOCKER_IPV6_ENABLED
+        value: "true"
+      - name: IP_FAMILY
+        value: "dual"
+    regex: ^(go.mod$|tests/|pkg/test/|prow/|content/en/boilerplates/snips/|content/en/docs/.*(test\.sh|/snips\.sh)$)
```
ericvn, line 52: Let's also make this optional until we've done some testing with the new tests:
```suggestion
    regex: ^(go.mod$|tests/|pkg/test/|prow/|content/en/boilerplates/snips/|content/en/docs/.*(test\.sh|/snips\.sh)$)
    modifiers: [presubmit_optional]
```

ericvn (commented):
You'll also need to run `make gen` to create the generated files.

ericvn:
The change will require another `make gen` to put the optional in the generated files.

kfaseela:
> The change will require another `make gen` to put the optional in the generated files.

Done

ericvn (approved):
This job is optional and should be merged to test the instio.io test. We may need to revisit and tweak as that test is updated.

PR #5135, opened 2024-01-02
Sort all generated jobs for istio repos in ascending order

jacob-delgado:
Follow up to https://github.com/istio/test-infra/pull/4844

Code is a bit ugly, but most of prowgen is a bit hacky. This will always have the generated files for the istio org branches be in ascending order based on the job name. Hopefully this can help reduce some of the churn we see when we add or remove jobs.

There's no need for a flag or any default. If someone sees it fit it should be done in a follow-up PR.


tools/prowgen/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.19 as build-env
+FROM golang:1.20 as build-env
```
ericvn, line 15: Is there a reason to not go to 1.21 here? I know 1.19 is EOL and 1.20 should be EOL soon as well.

howardjohn (approved):

tools/prowgen/Dockerfile
jacob-delgado: It's not really used for anything. It's only used when we are debugging prowgen anyway.

PR #5139, opened 2024-01-04
Add Kubernetes 1.29 support for Istio 1.20 and master

keithmattix:
Kubernetes 1.29 is out, so we should test against it. Because there's a mismatch between k8s releases and Istio releases (the former has 3 per year while the latter has 4), we run these tests for Istio 1.20 and the future 1.21


MorrisLaw (approved):
lgtm!

prow/config/jobs/istio.yaml
```diff
@@ -319,6 +319,22 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: " --istio.test.retries=1 "
 
+  - name: integ-k8s-129
```
howardjohn: We don't need this one. We test, in postsubmit, all versions except the "primary" version. All other tests run with the "primary" version.

Since 1.29 is already the primary version, no need for a redundant job here.

keithmattix: Ah good catch; removed

howardjohn (approved):
LGTM to add the job, we should make sure @ericvn approves to actually declare "support" though

PR #5181, opened 2024-01-26
Enable blobless fetch for git repos

howardjohn:
This should be quite a bit faster


prow/config.yaml
```diff
@@ -18,6 +18,7 @@ plank:
       gcs_configuration:
         bucket: "istio-prow"
         path_strategy: "explicit"
+      blobless_fetch: true
```
zirain, line 21: can you details about this?

howardjohn: https://github.com/kubernetes/test-infra/issues/26590

tl;dr `git clone` faster, should have zero downsides I think (or if I am wrong we can just revert)

linsun (commented):
cc @dhawton who is our prow expert to review

prow/config.yaml
zirain: TIL, thanks!

zirain (approved):

dhawton (approved):

linsun (approved):

kfaseela (approved):

ericvn:
@stewartbutler Would you take a look?

stewartbutler (approved):

PR #5183, opened 2024-01-30
add missing release-deps cm for 1.20 and 1.21

dhawton:
The release-deps configmaps weren't created during branch cutting for 1.20 and 1.21. Add them to the istio-private cluster.

### DIFF
--- a/prow/cluster/private/release-deps.yaml
+++ b/prow/cluster/private/release-deps.yaml
@@ -149,9 +149,6 @@ data:
       ztunnel:
         git: https://github.com/istio/ztunnel
         auto: deps
-      pkg:
-        git: https://github.com/istio/pkg
-        auto: modules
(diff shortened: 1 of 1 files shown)


prow/cluster/private/release-deps.yaml
```diff
@@ -133,3 +133,73 @@ kind: ConfigMap
...
+        auto: modules
+      proxy:
+        git: https://github.com/istio-private/proxy
+        auto: deps
+      ztunnel:
+        git: https://github.com/istio/ztunnel
+        auto: deps
+      pkg:
+        git: https://github.com/istio/pkg
+        auto: modules
```
howardjohn: I think we can drop this now, same with 1.20

howardjohn (approved):

PR #5186, opened 2024-01-31
Create generate-deps-configmaps tool

### LINKED ISSUE #5184: release-1.xx-release-deps config map not being created during branch cutting
https://github.com/istio/test-infra/pull/5183 -- these are being missed during branch cutting. These should be added to the tooling to alleviate issues in the istio-private cluster

dhawton:
The purpose of this script is to generate the istio-deps and release-deps configmaps (and any other we may need in the future) that are needed for istio-private work.

Currently these are not generated and are caught after-the-fact when the PSWG team is building security releases when jobs in Prow fail to schedule since the ConfigMaps are missing. We should, ideally, have this as part of the branch cutting process. Cleanup will still be manual, but can occur during the regular EOL cleanup flow.

After merging, the last step is to update the release branch cutting issue template to ensure this is completed during Step 3.

Closes #5184


tools/generate-deps-configmaps/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```bash
+python3 tools/generate-deps-ConfigMaps/generate.py [options]...
+```
+
+You can run this within `make shell` to ensure you have the appropriate Python and PyYAML versions.
+
+### Options
+
+The following is a list of supported options. If an option is optional, than its _default_ value will be used if not provided. If an option is required
```
ilrudie: sp.
```suggestion
The following is a list of supported options. If an option is optional, then its _default_ value will be used if not provided. If an option is required
```

tools/generate-deps-configmaps/README.md
```diff
@@ -0,0 +1,37 @@
...
+This should be run during the steps of Step 3 where the release branch jobs are being generated, though specific order does not matter.
+
+## Prerequisites
+
+- [Python 3.6+](https://www.python.org/downloads/)
+- [PyYAML](https://pyyaml.org/wiki/PyYAMLDocumentation)
+
+## Templates
+
+The templates utilize [Template strings](https://docs.python.org/3.4/library/string.html#template-strings) to generate the ConfigMap dependency data.
```
ilrudie: Any reason we point to 3.4 when we're asking for a 3.6+ prereq?

tools/generate-deps-configmaps/generate.py
```diff
@@ -0,0 +1,140 @@
...
+import sys
+import yaml
+import argparse
+from string import Template
+
+
+# Get a list of templates in the directory. We remove the suffix here as the name of the
+# template is also used in the name of the ConfigMap.
+def get_templates(path):
+    extension = ".template"
```
ilrudie: convey intent that this is meant to be constant. no practical effect to the code though so feel free to disregard
```suggestion
    EXTENSION = ".template"
```

tools/generate-deps-configmaps/generate.py
```diff
@@ -0,0 +1,140 @@
...
+
+    for template in templates:
+        print("Generating ConfigMap for " + template)
+        cm = create_configmap(args.minor_version, template, args.namespace)
+        release_deps.append(cm)
+
+    if not args.dry_run:
+        print("Writing to " + args.destination)
+        with open(args.destination, "w") as stream:
+            generate_yaml(cm, stream)
```
ilrudie:
```suggestion
            generate_yaml(release_deps, stream)
```

ilrudie (changes requested):
over all looks helpful, couple of suggestions and 1 issue with python happily allowing you to use an uninitialized variable which results in outputting a mangled file on the non-`--dry-run` codepath...

tools/generate-deps-configmaps/README.md
dhawton:
```suggestion
The templates utilize [Template strings](https://docs.python.org/3.6/library/string.html#template-strings) to generate the ConfigMap dependency data.
```

tools/generate-deps-configmaps/generate.py
```diff
@@ -0,0 +1,140 @@
...
+
+# Get a list of templates in the directory. We remove the suffix here as the name of the
+# template is also used in the name of the ConfigMap.
+def get_templates(path):
+    EXTENSION = ".template"
+    templates = []
+    for _, _, files in os.walk(path):
+        for file in files:
+            if file.endswith(extension):
+                templates.append(file.replace(extension, ""))
```
ilrudie: sorry, the previous suggestion was a partial which breaks things if accepted as it was. that's on me
```suggestion
            if file.endswith(EXTENSION):
                templates.append(file.replace(EXTENSION, ""))
```

ilrudie (changes requested):
this one's on me, my "convey that this is a const" suggestion should not have been accepted as is... it broke things. I didn't make that at all clear in my comment though

ilrudie:
A bit of an aside: Do we have any lint/formatting requirements for python? in other python project I contribute to [black](https://pypi.org/project/black/) is required. I'm really not an expert on python TBH but I think if we're going to have some python kicking around then introducing black to our linting is probably a good shout.

I don't think it's required for this to merge (IMO at least) but I would say maybe run black against your code and commit the results.

Also maybe we should open an issue to track/consider adding black formatter.

dhawton:
> A bit of an aside: Do we have any lint/formatting requirements for python? in other python project I contribute to [black](https://pypi.org/project/black/) is required. I'm really not an expert on python TBH but I think if we're going to have some python kicking around then introducing black to our linting is probably a good shout.
> 
> I don't think it's required for this to merge (IMO at least) but I would say maybe run black against your code and commit the results.
> 
> Also maybe we should open an issue to track/consider adding black formatter.

We use autopep8 to apply the pep 8 styling against it.

ilrudie:
> > A bit of an aside: Do we have any lint/formatting requirements for python? in other python project I contribute to [black](https://pypi.org/project/black/) is required. I'm really not an expert on python TBH but I think if we're going to have some python kicking around then introducing black to our linting is probably a good shout.
> > I don't think it's required for this to merge (IMO at least) but I would say maybe run black against your code and commit the results.
> > Also maybe we should open an issue to track/consider adding black formatter.
> 
> We use autopep8 to apply the pep 8 styling against it.

I was not aware of autopep8... thanks for the info and disregard previous comment then

ilrudie:
Side note... this script strips out the comments at the top of the yaml file I think. Might not be a big deal but something to consider

ilrudie (approved):
Potential issue with the yaml comments aside, this seems fit for purpose. All error handling isn't perfect but I don't think the imperfect error handling is dangerous and using this seems less error prone than manually adjusting the yaml.

ilrudie:
well, you can strike my comment about error handling now 😄

ericvn (approved):

PR #5215, opened 2024-02-16
add integ-calico job

ilrudie:
Adds an ambient test for Calico CNI as suggested [here](https://github.com/istio/istio/pull/49255#issuecomment-1932467482).

This addition will be used to help make sure that our inpod capture method functions as expected with Calico. Other CNIs may be considered/added later as well as tests which rely on these CNIs for configurations which aren't possible with the basic KinD CNI (example, K8s Network Policy).

PRs this depends on:

- https://github.com/istio/istio/pull/49255
- https://github.com/istio/common-files/pull/990

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -77,6 +77,7 @@ jobs:
         value: --istio.test.skipVM
 
   - name: integ-ambient-calico
+    modifiers: [presubmit_optional]
     requirements: [kind]
     env:
     - name: INTEGRATION_TEST_FLAGS
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -77,7 +77,7 @@ jobs:
         value: --istio.test.skipVM
 
   - name: integ-ambient-calico
-    modifiers: [presubmit_optional]
+    modifiers: [presubmit_optional, presubmit_skipped]
     requirements: [kind]
     env:
     - name: INTEGRATION_TEST_FLAGS
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -76,6 +76,20 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: --istio.test.skipVM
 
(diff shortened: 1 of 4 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -76,6 +76,20 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: --istio.test.skipVM
 
+  - name: integ-ambient-calico
+    requirements: [kind]
```
dhawton, line 80:
```suggestion
  - name: integ-ambient-calico
    modifiers: [presubmit_optional]
    requirements: [kind]
```

I think we should make it optional until we know it's working as expected

prow/config/jobs/istio.yaml
```diff
@@ -76,6 +76,20 @@ jobs:
       - name: INTEGRATION_TEST_FLAGS
         value: --istio.test.skipVM
 
+  - name: integ-ambient-calico
```
dhawton, line 79:
```suggestion
  - name: integ-ambient-calico
    modifiers: [presubmit_optional]
```

I think it should be optional until we know it's working as expected, then we can make it required.. as well as any other bugs that may need addressing without full stopping other work.

ilrudie: Sounds like a good plan. TY

dhawton:
LGTM. @howardjohn you had a fair amount of involvement with this in the Istio PR, wdyt?

prow/config/jobs/istio.yaml
```diff
@@ -76,6 +76,21 @@ jobs:
...
+    modifiers: [presubmit_optional]
```
howardjohn: I don't think we should run this on presubmit. Its a useful test but IMO we don't want to bloat up our test matrix this much (esp since we will be adding cilium, ..., in the future). can we make it `[presubmit_optional, presubmit_skipped]`?

ilrudie: makes sense. thanks for the suggestion

howardjohn (approved):

PR #5234, opened 2024-02-29
Initial jobs for sail-operator repo

jwendell:


prow/config/jobs/sail-operator.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - entrypoint
+    - ../test-infra/tools/automator/automator.sh
+    - --org=$AUTOMATOR_ORG
+    - --repo=sail-operator
+    - "--title=Automator: Update dependencies in $AUTOMATOR_ORG/$AUTOMATOR_REPO@$AUTOMATOR_BRANCH"
+    - --labels=auto-merge
+    - --modifier=update_deps
+    - --token-env
+    - --cmd=./tools/update_deps.sh
+    requirements: [github-istio-testing]
```
howardjohn, line 47: This gives (very) high level privilege here. This is acceptable, as its a periodic job that runs only on approved+merged code, rather than arbitrary presubmit code. We do this in Istio in many places. However, all Istio code like this is protect by CODEOWNERS where only trusted maintainers have approval access.

So I think this is fine as long as we make sure the repo maintains a similar trust level. In aprticular, do we need the 3 third party Admin tools here https://github.com/istio-ecosystem/sail-operator/settings/access? And will a CODEOWNERS be established? When it is, can we ensure that the owners for the `tools/` is scoped only to Istio maintainers?

prow/config/jobs_test.go
```diff
@@ -40,6 +40,7 @@ func TestJobs(t *testing.T) {
 				return fmt.Errorf("private org must use private cluster, got %v", j.Base.Cluster)
 			}
 		case "istio":
+		case "istio-ecosystem":
```
howardjohn: case doesn't fallthrough

```suggestion
		case "istio", "istio-ecosystem":
```

prow/config.yaml
```diff
@@ -409,6 +409,19 @@ branch-protection:
           branches:
             master:
               protect: true
+        sail-operator:
+          required_status_checks:
+            contexts:
+            - EasyCLA
```
howardjohn: I think this just means "CLA must pass". But we need the CLA to actually run which is owned by LF outside of Istio.

@therealmitchconnors do you know how to get that setup?

jwendell: +1. I'll remove it while we get the CI configured at least for the temporary branch.

prow/config/jobs_test.go
jwendell: too much C lately :/

howardjohn (approved):

prow/config.yaml
jwendell: @therealmitchconnors Do you know how we can get easyCLA installed in istio-ecosystem/sail-operator repo?

PR #5255, opened 2024-03-13
terraform: give various users access

howardjohn:
For now, this sets up "viewers" for private infra, and admins that
consist of myself. This is definitely not the long-term desired state
but should be good first steps

### DIFF
--- a/infra/gcp/common.tf
+++ b/infra/gcp/common.tf
@@ -1,0 +1,18 @@
+locals {
+  # Members with view access to private infrastructure.
+  # Anyone here MUST be in the PSWG
+  # Comment indicates github username
+  private_infra_viewers = toset([
+    "howardjohn@google.com",     # howardjohn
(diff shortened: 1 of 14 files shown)


infra/gcp/common.tf
```diff
@@ -0,0 +1,18 @@
+locals {
+  # Members with view access to private infrastructure.
+  # Anyone here MUST be in the PSWG
+  # Comment indicates github username
+  private_infra_viewers = toset([
+    "howardjohn@google.com",     # howardjohn
+    "johnbhoward96@gmail.com",   # howardjohn
+    "ketihmattix@microsoft.com", # keithmattix
+    "daniel.hawton@solo.io",     #dhawton
```
dhawton:
```suggestion
    "daniel.hawton@solo.io",     # dhawton
```

everyone else has a space, so this was bugs me :D

dhawton (approved):

PR #5266, opened 2024-03-21
Sail: Fix the path to automator repo

jwendell:


prow/config/jobs/sail-operator.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
     cron: "0 23 * * 1-5" # run Mon-Fri at 11:00PM UTC
     command:
     - entrypoint
-    - ../test-infra/tools/automator/automator.sh
+    - ../../test-infra/tools/automator/automator.sh
```
ericvn: Should this be
```suggestion
    - ../istio/test-infra/tools/automator/automator.sh
```

I'm thinking you are in Istio-ecosystem/sail-operator and the tooling would have istio/test-infra at the same level.

jwendell: I think it's istio.io

ericvn (approved):

PR #5272, opened 2024-03-27
Add --signoff option to automator script

dgn:
We need this because istio-ecosystem/sail-operator uses DCO to verify contributions - and that requires signoffs, even if it's just the bot creating an automated PR

### DIFF
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -28,7 +28,7 @@ cleanup() {
 }
 
 get_opts() {
-  if opt="$(getopt -o '' -l branch:,org:,repo:,title:,match-title:,body:,labels:,user:,email:,modifier:,script-path:,cmd:,token-path:,token:,token-env,merge-repository:,merge-branch:,git-exclude:,strict,dry-run,verbose -n "$(basename "$0")" -- "$@")"; then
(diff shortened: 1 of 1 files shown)


tools/automator/automator.sh
```diff
@@ -253,7 +257,7 @@ commit() {
 
   local src_branch="${AUTOMATOR_SRC_BRANCH:-none}"
   fork_name="$src_branch-$branch-$modifier-$(hash "$title")"
-  git -c "user.name=$user" -c "user.email=$email" commit --message "$title" --author="$user <$email>"
+  git -c "user.name=$user" -c "user.email=$email" commit --message "$title" --author="$user <$email>" "${commit_args[@]}"
```
jwendell, line 260: Wouldn't this fail (`set -u`) if this variable is not defined?

dgn: nope, not a problem because it's accessed using `[@]`

dgn: compare `script_args` and `shell_args` in this same file. they work the same way

jwendell: cool :)

jwendell (approved):

PR #5275, opened 2024-03-28
Add a job to verify MacOS builds still work

### LINKED ISSUE #49879: Add a pre-submit test that builds the Mac binaries.
(This is used to request new product features, please visit <https://github.com/istio/istio/discussions> for questions on using Istio)

**Describe the feature request**

See https://github.com/istio/istio/pull/49858, but occasionally a PR may require a small tweak in the MacOS/Linux code to continue to build on a Mac. Create a test that just builds the Mac binaries to verify a change isn't needed here.

Arguably, this doesn't happen often, but it should a smaller test to just try building the binaries.

**Describe alternatives you've considered**

**Affected product area (please put an X in all that apply)**

[ ] Ambient
[ ] Docs
[ ] Dual Stack
[ ] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

ericvn:
Fixes https://github.com/istio/istio/issues/49879

Initially make the test optional. Will make required once it's verified to pass.

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -402,6 +402,12 @@ jobs:
     regex: '^samples/bookinfo/src/'
     requirements: [docker]
 
+  # Try to create the Mac binaries. Ocassionally a change breaks the ability to build on MacOS.
+  - name: macbuildcheck
+    types: [presubmit]
+    modifiers: [presubmit_optional]
+    command: [TARGET_OS=darwin, make, build]
+
   - name: update-go-control-plane
     types: [periodic]
     cron: "0 2 * * 0" # run each Sunday at 02:00AM UTC
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -406,7 +406,7 @@ jobs:
   - name: macbuildcheck
     types: [presubmit]
     modifiers: [presubmit_optional]
-    command: [TARGET_OS=darwin, make, build]
+    command: [make, -e, "TARGET_OS=darwin", build]
 
   - name: update-go-control-plane
     types: [periodic]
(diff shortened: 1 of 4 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -402,6 +402,12 @@ jobs:
     regex: '^samples/bookinfo/src/'
     requirements: [docker]
 
+  # Try to create the Mac binaries. Ocassionally a change breaks the ability to build on MacOS.
+  - name: macbuildcheck
+    types: [presubmit]
+    modifiers: [presubmit_optional]
+    command: [TARGET_OS=darwin, make, build]
```
jwendell: would this work? perhaps `[make, -e, "TARGET_OS=darwin", build]` ?

also, does it make sense to restrict this job to changes to Makefiles?

ericvn: > also, does it make sense to restrict this job to changes to Makefiles?

No, as changes in code typically cause this to happen. Especially changes in one of the Linux only files which may require a change in the MacOS only file for compatibility.

ericvn: `make -e TARGET_OS=darwin build` does also seem to build the files if run locally, as well as the prior command. I suspect this might be better since the common-files script could maybe not pass TARGET_OS in the future.

jwendell: Sorry, my suggestion above was just to rearrange the order of the arguments, as the final yaml will be rendered having `TARGET_OS=darwin` as the first item in the array, not sure if it would work, as it's not a binary.

jwendell (approved):

PR #5284, opened 2024-04-04
prow: protect all branches in repos

howardjohn:
This changes our configuration to protect *all* branches in repos, not
just release ones.

This means users will need to work in forks. When operating from the UI,
GitHub should automatically do this for them. For development with
standard CLI tooling, users will need to adopt their workflows to
accomodate.

The `branches` config should not be required anymore, but I left it just
in case this doesn't work how I expect it to. Once we validate it, we
can remove those as well.

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -148,17 +148,23 @@ branch-protection:
   protect: false
   orgs:
     istio:
+      # Enable protection for all repos
+      protect: true
+      # Every repo requires CLA to pass
       required_status_checks:
         contexts:
         - EasyCLA
-      required_pull_request_reviews: &protection_required_pull_request_reviews
+      # By default, required 1 approval from a CODEOWNER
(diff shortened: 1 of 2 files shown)


howardjohn:
Put a hold to get multiple reviews as this is fairly sensitive

dhawton (approved):

jwendell (approved):

bleggett (approved):

cjwagner (approved):

ericvn:
One question. How will this affect the branch cut when the RMs need to create the new release branch in the repo? I know there is a post release branch creation to update the branch protection.

I think the `git push --set-upstream origin release-1.x` might be the issue.

howardjohn:
hmm good point. TBH not 100% sure the way around it but I am sure if we play around with it we can figure it out?

PR #5298, opened 2024-04-11
jobs: add Kubernetes 1.30 job

howardjohn:
Part of https://github.com/istio/istio/issues/50401


dhawton (approved):
Approving since it's DNM. LGTM except: I don't see the tag in gcr.io/istio-testing/kind-node yet, assuming that's still in progress.

PR #5328, opened 2024-04-24
automator: push scripted commits

dgn:
Previously, we would only push and create a PR if there were uncommitted changes in the repository. That doesn't work for scripts that create multiple commits, though, like a script that performs an automatic merge.

### DIFF
--- a/tools/automator/automator.sh
+++ b/tools/automator/automator.sh
@@ -257,7 +257,11 @@ commit() {
 
   local src_branch="${AUTOMATOR_SRC_BRANCH:-none}"
   fork_name="$src_branch-$branch-$modifier-$(hash "$title")"
-  git -c "user.name=$user" -c "user.email=$email" commit --message "$title" --author="$user <$email>" "${commit_args[@]}"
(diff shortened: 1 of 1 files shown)


tools/automator/automator.sh
```diff
@@ -257,7 +257,11 @@ commit() {
 
   local src_branch="${AUTOMATOR_SRC_BRANCH:-none}"
   fork_name="$src_branch-$branch-$modifier-$(hash "$title")"
-  git -c "user.name=$user" -c "user.email=$email" commit --message "$title" --author="$user <$email>" "${commit_args[@]}"
+  if ! git diff --cached --quiet --exit-code; then
+    git -c "user.name=$user" -c "user.email=$email" commit --message "$title" --author="$user <$email>"
+  else
+    echo "No changes to commit. Assuming the commit was already made"
```
ericvn: Nit:
```suggestion
    echo "No changes to commit. Assuming the commit was already made."
```

ericvn (approved):
LGTM.

PR #5395, opened 2024-06-12
prow: Add new C3 pool

howardjohn:
- **Bumping Prow**
- **Add new c3 node pool**

### DIFF
--- a/infra/gcp/istio-prow-build/cluster-prow.tf
+++ b/infra/gcp/istio-prow-build/cluster-prow.tf
@@ -174,3 +174,32 @@ module "prow_test" {
     "testing" = "test-pool"
   }
 }
+
+# Prow 'test' node pool is used for most jobs
+# Consists of c3-22 nodes.
+# Note: despite the naming "build" vs "test", a lot of build jobs use the test pool.
+module "prow_test_c3" {
+  source = "../modules/gke-nodepool"
+
+  name         = "istio-test-pool-c3"
+  cluster_name = "prow"
+  location     = "us-west1-a"
+
+  machine_type  = "c3-standard-22"
+  initial_count = 1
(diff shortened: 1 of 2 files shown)


prow/istio-autobump-config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-gitHubLogin: "istio-testing"
-gitHubToken: "/etc/github-token/oauth"
+gitHubLogin: "xxxxxxxxxxxxxxxxxxxxxxxxxx"
+gitHubToken: "/home/john/.secrets/github"
```
jwendell: Is this right?

howardjohn: nope, thanks

infra/gcp/istio-prow-build/cluster-prow.tf
```diff
@@ -173,4 +173,33 @@ module "prow_test" {
...
+  spot = true
+
+  disk_size_gb = 256
+  disk_type    = "pd-ssd"
+
+  project_name    = local.project_id
+  service_account = "istio-prow-jobs@istio-prow-build.iam.gserviceaccount.com"
+
+  labels = {
+    "testing" = "test-pool"
```
dhawton, line 203: Is this in addition to the existing e2 pool?

howardjohn: For now the plan was to trial run it side by side and if it is working well we can maybe move everything over. I only set max=10 here for that reason

dhawton: Thanks, that's what I thought just wanted to verify

dhawton (approved):

PR #5401, opened 2024-06-19
add ambient ipv6 integ-test job

### LINKED ISSUE #51222: Ambient IPv6 integration test runs
**Describe the feature request**
Right now we do not run any `ipv6` integration tests against `ambient` (as far as I can tell). IPv6 support is still experimental.

We should consider running `integ_ambient` with an IPv6 config as well as an IPv4 config (`integ_ambient_ipv6` or whatever)

This _should_ be relatively straightforward and help us de-experimentalize IPv6 support, just some test boilerplate work.

[x] Ambient
[ ] Docs
[x] Dual Stack
[ ] Installation
[ ] Networking
[ ] Extensions and Telemetry
[ ] Security
[x] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

kfaseela:
Fixes https://github.com/istio/istio/issues/51222

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -238,6 +238,23 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-ipv6
+    modifiers: [presubmit_optional]
+    requirements: [kind]
+    env:
+    - name: INTEGRATION_TEST_FLAGS
+      value: --istio.test.ambient
+    - name: DOCKER_IN_DOCKER_IPV6_ENABLED
+      value: "true"
+    - name: IP_FAMILY
+      value: "ipv6"
+    command:
+    - entrypoint
+    - prow/integ-suite-kind.sh
+    - --kind-config
+    - prow/config/ambient-sc.yaml
+    - test.integration.ambient.kube
+
   - name: integ-helm
     command: [entrypoint, prow/integ-suite-kind.sh, test.integration.helm.kube]
     requirements: [kind]
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -239,11 +239,11 @@ jobs:
     - test.integration.ambient.kube
 
   - name: integ-ambient-ipv6
-    modifiers: [presubmit_optional]
+    modifiers: [presubmit_optional, presubmit_skipped]
     requirements: [kind]
     env:
     - name: INTEGRATION_TEST_FLAGS
-      value: --istio.test.ambient
+      value: " --istio.test.kube.helm.values=cni.ambient.ipv6=true --istio.test.ambient "
     - name: DOCKER_IN_DOCKER_IPV6_ENABLED
       value: "true"
     - name: IP_FAMILY
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -96,6 +96,97 @@ postsubmits:
     - ^master$
     cluster: private
     decorate: true
+    name: integ-ambient-ipv6_istio_postsubmit_pri
+    path_alias: istio.io/istio
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - command:
+        - entrypoint
+        - prow/integ-suite-kind.sh
+        - --kind-config
+        - prow/config/ambient-sc.yaml
+        - test.integration.ambient.kube
+        env:
+        - name: BAZEL_BUILD_RBE_INSTANCE
+        - name: BUILD_WITH_CONTAINER
+          value: "0"
+        - name: DOCKER_IN_DOCKER_IPV6_ENABLED
+          value: "true"
+        - name: GOMAXPROCS
+          value: "5"
+        - name: INTEGRATION_TEST_FLAGS
+          value: --istio.test.ambient
+        - name: IP_FAMILY
+          value: ipv6
+        - name: DEPENDENCIES
+          valueFrom:
+            configMapKeyRef:
+              key: dependencies
+              name: master-istio-deps
+        - name: ISTIO_ENVOY_BASE_URL
+          value: https://storage.googleapis.com/istio-build-private/proxy
+        image: gcr.io/istio-testing/build-tools:master-89ff97db972e1ec7d61a7831bf9408f509e17d73
+        name: ""
+        resources:
+          limits:
+            cpu: "5"
+            memory: 24Gi
+          requests:
+            cpu: "5"
+            memory: 3Gi
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /home/prow/go/pkg
+          name: build-cache
+          subPath: gomod
+        - mountPath: /gocache
+          name: build-cache
+          subPath: gocache
+        - mountPath: /lib/modules
+          name: modules
+          readOnly: true
+        - mountPath: /sys/fs/cgroup
+          name: cgroup
+          readOnly: true
+        - mountPath: /var/lib/docker
+          name: docker-root
+        - mountPath: /home/.netrc
+          name: netrc
+          readOnly: true
+          subPath: .netrc
+      nodeSelector:
+        kubernetes.io/arch: amd64
+        testing: test-pool
+      volumes:
+      - hostPath:
+          path: /var/tmp/prow/cache
+          type: DirectoryOrCreate
+        name: build-cache
+      - hostPath:
+          path: /lib/modules
+          type: Directory
+        name: modules
+      - hostPath:
+          path: /sys/fs/cgroup
+          type: Directory
+        name: cgroup
+      - emptyDir: {}
+        name: docker-root
+      - name: netrc
+        secret:
+          items:
+          - key: secret
+            mode: 384
+            path: .netrc
+          secretName: netrc-secret
+  - annotations:
+      testgrid-create-test-group: "false"
+    branches:
+    - ^master$
+    cluster: private
+    decorate: true
     name: integ-ambient_istio_postsubmit_pri
     path_alias: istio.io/istio
     spec:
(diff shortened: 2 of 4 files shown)


kfaseela:
@bleggett do we want to have this as part of presubmit? Or we want to first run this locally and make sure we raise all the bugs, fix them, and then get this merged?

bleggett:
Is this enabling the ambient IPv6 support flag?

https://github.com/istio/istio/blob/b9aa2e993604619c6fe80c8c293ef0c80e189924/manifests/charts/istio-cni/values.yaml#L55

(on mobile/limited access, cannot really tell)

kfaseela:
> Is this enabling the ambient IPv6 support flag?
> 
> https://github.com/istio/istio/blob/b9aa2e993604619c6fe80c8c293ef0c80e189924/manifests/charts/istio-cni/values.yaml#L55
> 
> (on mobile/limited access, cannot really tell)

not really, let me fix that... currently this was just making the kind cluster ipv6 and then run ambient profile..did not know that there is one other explicit flag

bleggett:
> > Is this enabling the ambient IPv6 support flag?
> > 
> > https://github.com/istio/istio/blob/b9aa2e993604619c6fe80c8c293ef0c80e189924/manifests/charts/istio-cni/values.yaml#L55
> > 
> > (on mobile/limited access, cannot really tell)
> 
> not really, let me fix that... currently this was just making the kind cluster ipv6 and then run ambient profile..did not know that there is one other explicit flag

Yeah there are a bunch of non-Istio flags - ambient _should_ really only have 1 "IPv6" flag, to begin to fix the obfuscatory mess of "dualstack" flags and options we have in sidecar.

kfaseela:
@bleggett I am enabling the ipv6 flag as you suggested, and was checking the cni logs to make sure it is really enabled, I do see yet another ipv6 flag as well, does that also need to be turned on?

```
$ k logs -n istio-system istio-cni-node-429dx | grep -i ipv6
AmbientIPv6: true
ENABLE_INBOUND_IPV6=false
ENABLE_INBOUND_IPV6=false
ENABLE_INBOUND_IPV6=false
$
```

bleggett:
> @bleggett I am enabling the ipv6 flag as you suggested, and was checking the cni logs to make sure it is really enabled, I do see yet another ipv6 flag as well, does that also need to be turned on?
> ENABLE_INBOUND_IPV6=false

No, that is a sidecar legacy flag, not used for any ambient IPv6 logic or rules addition.

Another question - 

Can you confirm these tests fail if https://github.com/istio/istio/blob/b9aa2e993604619c6fe80c8c293ef0c80e189924/manifests/charts/istio-cni/values.yaml#L55 is turned off? I would expect them to, and they need to to be useful.

leosarra:
I tried to run with cni.ambient.ipv6=true in an IPv6-only KinD cluster and I'm getting error in istio-validation init containers

```
go test -p 1 -tags=integ -run TestTraffic ./tests/integration/ambient -timeout 30m --log_output_level=tf:debug -v --istio.test.work_dir=out/linux_amd64 --istio.test.hub=gcr.io/istio-testing  --istio.test.tag=latest --istio.test.skipWorkloads=tproxy,vm,proxyless --istio.test.ambient=true --istio.test.kube.helm.values=cni.ambient.ipv6=true

...

kubectl logs sidecar-v2-cbc6c454f-8884b -n echo-1-74656 -c istio-validation
2024-06-25T12:14:00.626364Z	info	Starting iptables validation. This check verifies that iptables rules are properly established for the network.
2024-06-25T12:14:00.626407Z	info	Listening on [::1]:15001
2024-06-25T12:14:00.626562Z	info	Listening on [::1]:15006
2024-06-25T12:14:05.630890Z	error	iptables validation failed; workload is not ready for Istio.
When using Istio CNI, this can occur if a pod is scheduled before the node is ready.

If installed with 'cni.repair.deletePods=true', this pod should automatically be deleted and retry.
Otherwise, this pod will need to be manually removed so that it is scheduled on a node with istio-cni running, allowing iptables rules to be established.
```

Did I make a mistake in running the tests? I tried with cni.ambient.ipv6=false as well but the problem persist. I have no issue in running non-ambient tests.

bleggett:
> I tried to run with cni.ambient.ipv6=true in an IPv6-only KinD cluster and I'm getting error in istio-validation init containers
> 
> ```
> go test -p 1 -tags=integ -run TestTraffic ./tests/integration/ambient -timeout 30m --log_output_level=tf:debug -v --istio.test.work_dir=out/linux_amd64 --istio.test.hub=gcr.io/istio-testing  --istio.test.tag=latest --istio.test.skipWorkloads=tproxy,vm,proxyless --istio.test.ambient=true --istio.test.kube.helm.values=cni.ambient.ipv6=true
> 
> ...
> 
> kubectl logs sidecar-v2-cbc6c454f-8884b -n echo-1-74656 -c istio-validation
> 2024-06-25T12:14:00.626364Z	info	Starting iptables validation. This check verifies that iptables rules are properly established for the network.
> 2024-06-25T12:14:00.626407Z	info	Listening on [::1]:15001
> 2024-06-25T12:14:00.626562Z	info	Listening on [::1]:15006
> 2024-06-25T12:14:05.630890Z	error	iptables validation failed; workload is not ready for Istio.
> When using Istio CNI, this can occur if a pod is scheduled before the node is ready.
> 
> If installed with 'cni.repair.deletePods=true', this pod should automatically be deleted and retry.
> Otherwise, this pod will need to be manually removed so that it is scheduled on a node with istio-cni running, allowing iptables rules to be established.
> ```
> 
> Did I make a mistake in running the tests? I tried with cni.ambient.ipv6=false as well but the problem persist. I have no issue in running non-ambient tests.
> 

Right now these are only testing ambient ipv6, not ambient plus sidecar. Sidecar has its own v6 problems, which will probably need additional tests and other fixes, such as the ones John is currently working on.

kfaseela:
@bleggett shall we merge this PR, and trigger the test on some dummy istio PR to fix issues one by one? As per Leonardo's run, the ambient test with ipv6 enabled is failing on basic run.

bleggett:
> @bleggett shall we merge this PR, and trigger the test on some dummy istio PR to fix issues one by one? As per Leonardo's run, the ambient test with ipv6 enabled is failing on basic run.

That should only be with sidecars (which have their own preexisting ipv6 issues unrelated to ambient) - we should disable sidecar tests for this, this suite should test 

- just ambient
- just ipv6
- and it should fail if `ambient.ipv6=false` and we run it in an _ipv6-only context_ (we don't have to automate this, but it should be tested manually to verify - these tests must run in a pure ipv6-only context, or we will get false positives because something falls back to ipv4 and "dualstack" support masks it)

as a baseline.

We can fix the sidecar + ambient + ipv6 interop stuff separately in followup PRs,  and append tests for that separately.

@leosarra can you reproduce that outside of the test suite here? enabling `ambient.ipv6` shouldn't affect init containers or validation containers, those are only for sidecar, and we should categorically ignore those for ambient. If you can, raise a separate issue, that shouldn't have anything to do with the PR here.

So @kfaseela once you confirm that these tests are running+passing in a pure ipv6 context (not dualstack) and that they fail for you locally in that pure ipv6 test context if you set `ambient.ipv6=false`, that's all we need to do here.

kfaseela:
> > @bleggett shall we merge this PR, and trigger the test on some dummy istio PR to fix issues one by one? As per Leonardo's run, the ambient test with ipv6 enabled is failing on basic run.
> 
> That should only be with sidecars (which have their own preexisting ipv6 issues unrelated to ambient) - we should disable sidecar tests for this, this suite should test 
> 
> - just ambient
> - just ipv6
> - and it should fail if `ambient.ipv6=false` and we run it in an _ipv6-only context_ (we don't have to automate this, but it should be tested manually to verify - these tests must run in a pure ipv6-only context, or we will get false positives because something falls back to ipv4 and "dualstack" support masks it)
> 
> as a baseline.
> 
> We can fix the sidecar + ambient + ipv6 interop stuff separately in followup PRs,  and append tests for that separately.
> 
> @leosarra can you reproduce that outside of the test suite here? enabling `ambient.ipv6` shouldn't affect init containers or validation containers, those are only for sidecar, and we should categorically ignore those for ambient. If you can, raise a separate issue, that shouldn't have anything to do with the PR here.
> 
> So @kfaseela once you confirm that these tests are running+passing in a pure ipv6 context (not dualstack) and that they fail for you locally in that pure ipv6 test context if you set `ambient.ipv6=false`, that's all we need to do here.

As you know the ambient test framework has sidecar enabled echo services as well. So we can merge this PR, this is anyways optional test, and will be triggered explicitly on PRs as of now. So I can raise a test PR on Istio and we can keep running this and modify the PR as and when we discover ipv6 issues

prow/config/jobs/istio.yaml
```diff
@@ -238,6 +238,23 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-ipv6
+    modifiers: [presubmit_optional, presubmit_skipped]
+    requirements: [kind]
+    env:
+    - name: INTEGRATION_TEST_FLAGS
+      value: " --istio.test.kube.helm.values=cni.ambient.ipv6=true --istio.test.ambient "
```
howardjohn, line 246: ipv6 is default enabled now

howardjohn (approved):

prow/config/jobs/istio.yaml
kfaseela: @bleggett thinks it is better to explicitly enable this, so that people understand the various knobs that are in use, in case the defaults change later on

bleggett:
> As you know the ambient test framework has sidecar enabled echo services as well. So we can merge this PR, this is anyways optional test, and will be triggered explicitly on PRs as of now. So I can raise a test PR on Istio and we can keep running this and modify the PR as and when we discover ipv6 issues

That's fair enough. Removing sidecars from the picture requires a new suite that strips out the sidecar bits, like in https://github.com/istio/istio/pull/51574 - I have a slight preference for testing the feature in isolation (without sidecars) _and_ testing the combined confluence of features (sidecar + ipv6 + ambient + ipv6 + cni sidecar repair + cni ambient ).

Having multiple tests that test the feature in isolation _and_ test the confluence of all those makes diagnosis and triaging of failures much simpler when something breaks - if the _only_ test you have is testing a confluence of 4-5 different independently toggleable, relatively complex features, it can be quite challenging to figure out where the problem actually is (and can, as we've seen with CNI repair, occasionally mask genuine issues as well).

But - any coverage is better than none.

PR #5402, opened 2024-06-19
make ambient docs test job mandatory

kfaseela:
This was previously added as optional for initial testing purpose, now let us make it mandatory.


howardjohn (approved):
thanks!

PR #5420, opened 2024-07-02
add an optional ambient dual-stack test job

kfaseela:
This will enable us run the tests on demand, and see if ambient mode works on a dual stack cluster

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -96,6 +96,97 @@ postsubmits:
     - ^master$
     cluster: private
     decorate: true
+    name: integ-ambient-dual_istio_postsubmit_pri
+    path_alias: istio.io/istio
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - command:
+        - entrypoint
+        - prow/integ-suite-kind.sh
+        - --kind-config
+        - prow/config/ambient-sc.yaml
+        - test.integration.ambient.kube
+        env:
+        - name: BAZEL_BUILD_RBE_INSTANCE
+        - name: BUILD_WITH_CONTAINER
+          value: "0"
+        - name: DOCKER_IN_DOCKER_IPV6_ENABLED
+          value: "true"
+        - name: GOMAXPROCS
+          value: "5"
+        - name: INTEGRATION_TEST_FLAGS
+          value: ' --istio.test.kube.helm.values=cni.ambient.ipv6=true --istio.test.ambient '
+        - name: IP_FAMILY
+          value: dual
+        - name: DEPENDENCIES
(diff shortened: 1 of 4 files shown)


kfaseela:
cc @bleggett

bleggett:
This is still testing "sidecar dualstack" right? Because we aren't creating a suite that factors out sidecars, like in https://github.com/istio/istio/pull/51574?

I mention this because we already have the ` integ-ds` job which tests dualstack for sidecars, so we have some overlap here we probably do not want.

If we want to keep sidecars as part of the ambient test, we should just extend `integ-ds`. If we want a separate test, we should probably construct a separate suite to run in this job that doesn't overlap with `integ-ds`.

kfaseela:
> This is still testing "sidecar dualstack" right? Because we aren't creating a suite that factors out sidecars, like in [istio/istio#51574](https://github.com/istio/istio/pull/51574)?
> 
> I mention this because we already have the ` integ-ds` job which tests dualstack for sidecars, so we have some overlap here we probably do not want.
> 
> If we want to keep sidecars as part of the ambient test, we should just extend `integ-ds`. If we want a separate test, we should probably construct a separate suite to run in this job that doesn't overlap with `integ-ds`.

bleggett:
We may also need to create a dedicated "ambient dualstack" suite (or reuse `integ-ds`) because I think we need to ensure we are explicitly creating some K8S Services with _both_ kinds of IPs, and I do not think that will happen by default just by setting IP_FAMILY to `dual` during cluster creation - I could be wrong tho. We need to check.


https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-options-on-new-services

indicates that if you create new Services in a dualstack cluster, they will be by-default single-stack, unless you specifically set `ipFamilyPolicy` on the Service - we would need to do that for these tests to be useful.

bleggett:
(that being said we can merge this as an optional job and iterate)

bleggett (approved):
We will need to actually create/use a customized suite for this to test it properly as mentioned in the comments, but I am OK with adding an optional job to experiment with as a first step.

jacob-delgado (approved):

kfaseela:
> We may also need to create a dedicated "ambient dualstack" suite (or reuse `integ-ds`) because I think we need to ensure we are explicitly creating some K8S Services with _both_ kinds of IPs, and I do not think that will happen by default just by setting IP_FAMILY to `dual` during cluster creation - I could be wrong tho. We need to check.
> 
> 
> https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-options-on-new-services
> 
> indicates that if you create new Services in a dualstack cluster, they will be by-default single-stack, unless you specifically set `ipFamilyPolicy` on the Service - we would need to do that for these tests to be useful.

If the cluster is dualstack, we set all services with IP family for dual, as part of the integration test framework already.

PR #5456, opened 2024-08-07
Another new Slack URL

craigbox:
Updated with an inviter URL that will last for 2000 uses.

### DIFF
--- a/infra/gcp/istio-io/run.tf
+++ b/infra/gcp/istio-io/run.tf
@@ -18,8 +18,8 @@ resource "google_cloud_run_v2_service" "redirector" {
       image = "gcr.io/istio-testing/redirector"
       env {
         name = "REDIRECT_URL"
-        # expires 2034-07-21 or after 400 uses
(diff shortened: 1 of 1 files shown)


infra/gcp/istio-io/run.tf
```diff
@@ -18,8 +18,8 @@ resource "google_cloud_run_v2_service" "redirector" {
       image = "gcr.io/istio-testing/redirector"
       env {
         name = "REDIRECT_URL"
-        # expires 2034-07-21 or after 400 uses
-        value = "https://join.slack.com/t/istio/shared_invite/zt-2n6a8whhg-8IB9cGVezfL5Kqyy9Dfbxg"
+        # expires 2024-07-30 or after 2000 uses
```
kfaseela: This does not seem right?

craigbox: thanks, meant 2025

kfaseela (approved):

PR #5484, opened 2024-09-12
postsubmit coverage

Stevenjin8:
Coverage postsubmit job.

### DIFF
--- a/prow/config/jobs/ztunnel.yaml
+++ b/prow/config/jobs/ztunnel.yaml
@@ -14,6 +14,11 @@ jobs:
     types: [presubmit]
     requirements: [cratescache]
 
+  - name: coverage
+    command: [entrypoint, make, coverage]
+    types: [postsubmit]
+    requirements: [cratescache]
+
   - name: release
     architectures: [amd64, arm64]
     service_account_name: prowjob-testing-write
--- a/prow/cluster/jobs/istio/ztunnel/istio.ztunnel.master.gen.yaml
+++ b/prow/cluster/jobs/istio/ztunnel/istio.ztunnel.master.gen.yaml
@@ -7,6 +7,53 @@ postsubmits:
       testgrid-num-failures-to-alert: "1"
     branches:
     - ^master$
+    decorate: true
+    name: coverage_ztunnel_postsubmit
+    path_alias: istio.io/ztunnel
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - command:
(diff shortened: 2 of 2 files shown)


Stevenjin8:
hopefully i configured this right...

prow/config/jobs/ztunnel.yaml
```diff
@@ -14,6 +14,11 @@ jobs:
     types: [presubmit]
     requirements: [cratescache]
 
+  - name: coverage
+    command: [entrypoint, make, coverage]
+    types: [postsubmit]
+    requirements: [cratescache]
```
Stevenjin8, line 20: Will this break the cache since `make converage` will compile with some extra RUSTC flags?

bleggett: I don't know enough about prow or `cratescache` to know.

Unless `cratescache` is doing something very strange, it should cache pulled crates individually - so if we have 80% precached from previous jobs and `coverage` pulls a net new 20%, we should be fine, so I seriously doubt it will cause significant problems.

cc @dhawton tho for prow context

bleggett:
xref: https://github.com/istio/ztunnel/pull/1283

prow/config/jobs/ztunnel.yaml
bleggett: Oh it's just a persistently mounted `.cargo` dir, and not something with any build intelligence.

Meh, I am inclined to not care - this is a postsubmit job anyway. If we want to make it a presubmit job, execution time matters more and we can solve for that when we have that problem, but we aren't doing that here.

bleggett: @Stevenjin8 if you're feeling like it, you can create a dedicated `cratescoveragecache` mount and use that here instead. That will use considerably more disk space on the build box tho, so again I'd consider this optional for a postsubmit job.

bleggett (approved):

prow/config/jobs/ztunnel.yaml
dhawton: It is a host mounted .cargo dir, so should only be a cache of the crates to save from redownload. I don't believe it stores anything additional.

howardjohn: Yeah its not a build cache. Would love to have one but it doesn't work well with cargo since the builds take an exclusive lock on the cache..

howardjohn (approved):

PR #5554, opened 2024-12-18
Update prow to latest version - 12/18/2024

jacob-delgado:
We haven't updated prow and its configs for a while (see https://github.com/istio/test-infra/pull/5515). I did this manually.


howardjohn (approved):
LGTM but put a hold until end of day -- we are doing a release today and I don't want to risk anything until after

PR #5588, opened 2025-01-24
nil check os.DirEntry before checking d.IsDir in callback

mikemorris:
Under some circumstances, [WalkDirFunc](https://pkg.go.dev/io/fs#WalkDirFunc) will call the callback with a nil os.DirEntry argument.

Alternatively, moving the `err` check above the `d.IsDir()` check should also fix the issue, as `err` is set non-nil in both documented cases where `d` would be nil.

This is intended to avoid the [segfault](https://github.com/istio/istio/issues/54707) in release automation step 3.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1300a88]

goroutine 1 [running]:
main.main.func1({0x17715a1, 0x12}, {0x0?, 0x0?}, {0x1a75ae0, 0x40007660c0})
        /tmp/tmp.kuMqMgoaHb/branch/work/src/istio.io/test-infra/tools/prowgen/cmd/prowgen/main.go:71 +0x38
path/filepath.WalkDir({0x17715a1, 0x12}, 0x40006fbf18)
        /usr/local/go/src/path/filepath/path.go:398 +0xa0
main.main()
        /tmp/tmp.kuMqMgoaHb/branch/work/src/istio.io/test-infra/tools/prowgen/cmd/prowgen/main.go:70 +0x254
exit status 2
Error: failed to branch: failed to setup prow: failed to generate new prow config: exit status 1
exit status 1
```

### DIFF
--- a/tools/prowgen/cmd/prowgen/main.go
+++ b/tools/prowgen/cmd/prowgen/main.go
@@ -44,7 +44,7 @@ var (
 	preprocessCommand   = flag.String("pre-process-command", "", "command to run to preprocess the meta config files")
 	postprocessCommand  = flag.String("post-process-command", "", "command to run to postprocess the generated config files")
 	longJobNamesAllowed = flag.Bool("allow-long-job-names", false, "allow job names that are longer than 63 characters")
-	skipGarTagging = flag.Bool("skip-gar-tagging", false, "skip tagging gar images since that is permitted by few folks")
+	skipGarTagging      = flag.Bool("skip-gar-tagging", false, "skip tagging gar images since that is permitted by few folks")
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


tools/prowgen/cmd/prowgen/main.go
```diff
@@ -44,7 +44,7 @@ var (
 	preprocessCommand   = flag.String("pre-process-command", "", "command to run to preprocess the meta config files")
 	postprocessCommand  = flag.String("post-process-command", "", "command to run to postprocess the generated config files")
 	longJobNamesAllowed = flag.Bool("allow-long-job-names", false, "allow job names that are longer than 63 characters")
-	skipGarTagging = flag.Bool("skip-gar-tagging", false, "skip tagging gar images since that is permitted by few folks")
+	skipGarTagging      = flag.Bool("skip-gar-tagging", false, "skip tagging gar images since that is permitted by few folks")
```
kfaseela, line 47: do we need this?

mikemorris: It's just my editor running gofmt again, this time it's minor and i don't think it hurts through, rest of the file is already formatted.

kfaseela (approved):

PR #5589, opened 2025-01-30
Fix the default --input-dir and --cluster-dir for prowgen

kfaseela:
Release Automation Step 3 fails at this step as 

```
2025-01-30T21:47:06.126583Z	info	Running command: git tag --no-sign 1.25
2025-01-30T21:47:06.139620Z	info	Fetched all sources and setup working directory at /tmp/tmp.sXnNuBmun6/branch/work
2025-01-30T21:47:06.140969Z	info	*** Updating prow config for new branches.
2025-01-30T21:47:06.141009Z	info	Running command: go run ./cmd/prowgen/main.go branch 1.25 --skipGarTagging
2025/01/30 22:47:12 lstat ./prow/config/jobs: no such file or directory
```

The dir is actually `test-infra/prow/config/jobs`

### DIFF
--- a/tools/prowgen/cmd/prowgen/main.go
+++ b/tools/prowgen/cmd/prowgen/main.go
@@ -39,8 +39,8 @@ var (
 	// regex to match the test image tags.
 	tagRegex = regexp.MustCompile(`^(.+):(.+)-([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}-[0-9]{2}-[0-9]{2}|[0-9a-f]{40})$`)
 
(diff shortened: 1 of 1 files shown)


kfaseela:
cc @mikemorris

howardjohn (commented):
Why not just explicitly set these when you call it? 

I think other usages are running from the root of the repo

kfaseela:
> Why not just explicitly set these when you call it?
> 
> I think other usages are running from the root of the repo

Makes sense, raised the respective review in release-builder repo. See https://github.com/istio/release-builder/pull/2025/files

PR #5590, opened 2025-01-31
Fix command parsing for prow commands

kfaseela:
The parsing logic has some issues which breaks the release automation STEP 3.
This is needed for https://github.com/istio/release-builder/pull/2025/files


kfaseela:
cc @mikemorris

howardjohn (approved):

PR #5592, opened 2025-01-31
actually update branched image tag in all jobs, not just parent config

mikemorris:
This appears to be an oversight where the image was not being updated to the branched tag in all jobs, only in the parent config. I believe this should replace the need for the following manual part of Step 3 as described in https://github.com/istio/istio/issues/54707

> In the job configs prow/**/*-1.25.yaml, replace any image build-tools(-centos|-proxy):master with build-tools(-centos|-proxy):release-1.25. This is easy to do in VSCode, like this:

Additionally, filters out duplicate env vars in all jobs to replace the need for the following other manual part of Step 3, 
refs https://pkg.go.dev/k8s.io/api/core/v1#EnvVar

> Remove duplicated env entries of the following in the configs (you can use a similar search in VSCode as above, but beware, it'll only match the indentation level that you specify if you have strict matching enabled.):
```
- name: BUILD_WITH_CONTAINER
  value: "0"
```

Finally, adds logging for commands needed to retag images when running with `--skip-gar-tagging` and updates some minor deprecated filesystem calls.

/cc @kfaseela @dhawton

### DIFF
--- a/tools/prowgen/cmd/prowgen/main.go
+++ b/tools/prowgen/cmd/prowgen/main.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
@@ -84,7 +83,7 @@ func main() {
 
 			imagesToTag := make(map[string]string)
 
-			files, _ := ioutil.ReadDir(path)
+			files, _ := os.ReadDir(path)
 			for _, file := range files {
 				if file.IsDir() {
 					continue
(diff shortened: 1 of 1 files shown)


mikemorris:
/hold

Going to add the duplicate env removal on top of this since it's now actually iterating over all jobs.

mikemorris:
/unhold

Tested locally and this appears to be working as intended now.

kfaseela (approved):

kfaseela:
@mikemorris : this LGTM. I just added a hold in case you have anything else to verify or want @howardjohn to do a review before merging. Feel free to remove the dnm when you want to merge it

zirain (approved):

kfaseela (approved):

PR #5597, opened 2025-02-06
[release-1.25] Automated branching step 3

mikemorris:
Does https://github.com/istio/test-infra/pull/5593 need to be merged first?

This appears to still be missing some changes as compared to #5513:
- `prow/cluster/jobs/istio-private/*`
  - gencheck CI is expecting these, probably easiest to debug
  - Looks like we might not be running the `prowtrans` commands which would be manually called by `make gen`
    - https://github.com/istio/test-infra/blob/master/Makefile.core.mk#L33-L34
- `prow/config/private-presets.yaml`
  - ???


mikemorris:
Replaced by #5598

PR #5615, opened 2025-03-03
Add windowsbuildcheck

keithmattix:
Now that we're splitting out the linux specific builds to separate files, we need to add a build check for a non linux OS. Windows seems like a good choice

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -481,9 +481,9 @@ jobs:
     types: [presubmit]
     command: [make, -e, "TARGET_OS=darwin", build]
 
-  - name: windowscnibuildcheck
+  - name: windowsbuildcheck
     types: [presubmit]
-    command: [make, -e, "TARGET_OS=windows", build-cni]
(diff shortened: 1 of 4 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -481,6 +481,10 @@ jobs:
     types: [presubmit]
     command: [make, -e, "TARGET_OS=darwin", build]
 
+  - name: windowscnibuildcheck
+    types: [presubmit]
+    command: [make, -e, "TARGET_OS=windows", build-cni]
```
howardjohn: Should we try to build more? At least istioctl should build. I guess we find that from release test. The OSX one is largely to ensure developers don't get broken I guess. I suspect we could make windows work more broadly, maybe we start with this though. I would name the job windowsbuildcheck to future proof it from the details

howardjohn (approved):
LGTM, put a hold for one naming nit

prow/config/jobs/istio.yaml
keithmattix: Yeah, there are some other refactors that are needed to make windows build more generally. Good call out on the naming though. Will ix

PR #5627, opened 2025-03-25
Allow release from experimental branch

howardjohn:
This adds a *manual* release step for experimental branch. The current go code doesn't support this so I updated it to allow this option.

This will be use for the experimental inference serving.

A manual release was chosen due to lax review requirements on these branches, while the postsubmit job offers high privileges. Making this a manual operation ensures an admin can review the code is safe before triggering the release job.


LiorLieberman:
Thanks @howardjohn ! 
/lgtm fwiw

dhawton (approved):

PR #5648, opened 2025-04-17
Add jobs for ambient multicluster

keithmattix:
Depends on istio/istio#55991 so adding a hold

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -250,6 +250,20 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-multicluster
+    requirements: [kind]
+    env:
+    - name: INTEGRATION_TEST_FLAGS
+      value: --istio.test.ambient --istio.test.ambient.multinetwork
+    command:
+    - entrypoint
+    - prow/integ-suite-kind.sh
+    - --kind-config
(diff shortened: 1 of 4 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -250,6 +250,20 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-multicluster
```
dhawton:
```suggestion
  - name: integ-ambient-mc
```

Thoughts on using -mc instead of -multicluster? The other jobs use -mc instead of multicluster so this would be more consistent.

keithmattix: Yeah works for me

prow/config/jobs/istio.yaml
```diff
@@ -250,6 +250,20 @@ jobs:
...
+    requirements: [kind]
```
dhawton, line 255:
```suggestion
    modifiers: [presubmit_optional]
    requirements: [kind]
```

Also a thought, make this optional while the testing is added to make sure everything is working as expected before we make it required. If we don't, and the tests aren't ready/perfect, we will block PRs.

keithmattix: Completely forgot that was an option; thanks for the catch

dhawton (approved):

PR #5685, opened 2025-06-24
Automate kind image node building

keithmattix:
Something like this or a bash script would be useful to automate the copying of kind images from kindest to gcr (to avoid rate-limits). Putting this up just to give an example of what TF would cost; happy to convert to bash if that's preferred


jacob-delgado:
Adding @howardjohn as he currently does this manually.

howardjohn (commented):
I chatted with Keith about this offline, I am not sure it provides much benefit.

Before: run a bash script on each version we want
After: run a bash script on each version we want AND a terraform module

unless I am mistaken

keithmattix:
Where is the bash script? If it already exists, we should totally just use that. Or if it doesn't exist, could you point me to a good place for it to live?

howardjohn:
tools/build-kind-image.sh

keithmattix:
Ok I'm just now realizing that you're referring to the istio repo and not this one. Closing

PR #5705, opened 2025-07-28
Indicate new expiration date for token

howardjohn:
No terraform change needed, this is pulling from GCP secret manager which we write outside TF


keithmattix (approved):

infra/gcp/istio-prow-build/secrets.tf
```diff
@@ -6,7 +6,7 @@ locals {
     # Access token for "istio" dockerhub account
     "release_docker_istio",
     # Fine grained PAT in the Istio org, "github/istio-release/release". Has write access to "Contents" and "Workflows".
-    # Expires 6/26/2025.
+    # Expires 7/29/2026.
```
keithmattix, line 9: Can you update L15 as well? Going to set up a Slack reminder in test-and-release so we don't get broken again

jacob-delgado (approved):

PR #5706, opened 2025-07-29
Istio Owned CNI Config Presubmit Tests

jaellio:
Optional tests

Part of https://github.com/istio/istio/pull/56972

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -236,6 +236,7 @@ jobs:
 
   - name: integ-ambient-owned-cni
     modifiers: [presubmit_optional]
+    optional: true
     requirements: [kind]
     env:
     - name: INTEGRATION_TEST_FLAGS
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -236,7 +236,6 @@ jobs:
 
   - name: integ-ambient-owned-cni
     modifiers: [presubmit_optional]
-    optional: true
     requirements: [kind]
     env:
     - name: INTEGRATION_TEST_FLAGS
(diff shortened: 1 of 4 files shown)


prow/config/jobs/istio.yaml
```diff
@@ -234,6 +234,19 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-owned-cni
+    modifiers: [presubmit_optional]
+    requirements: [kind]
```
dhawton, line 239:
```suggestion
    optional: true
    requirements: [kind]
```

Strongly encourage setting optional until the tests are passing as expected/validated, or this will block PRs

dhawton (changes requested):

prow/config/jobs/istio.yaml
```diff
@@ -234,6 +234,20 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-owned-cni
+    modifiers: [presubmit_optional]
+    optional: true
```
therealmitchconnors: this change broke the test.
```suggestion
```

therealmitchconnors (approved):

prow/config/jobs/istio.yaml
jaellio: Great thanks! Is this covered by presubmit_optional?

prow/config/jobs/istio.yaml
dhawton: it shouldn't break the test, how did it do that?

dhawton: Go Docs showing the field: https://pkg.go.dev/sigs.k8s.io/prow/pkg/config#Presubmit, we really shouldn't add a non-optional test without knowing it will work because it will block PRs from merging. We did this just a couple weeks ago with Keith's ambient-mc integ tests

dhawton (approved):

PR #5707, opened 2025-07-30
Seperate IstioOwnedCNI flag to avoid helmValues read

jaellio:
HelmValues are challenging to read in istio tests to determine whether or not to skip a test. Adding a separate setting makes this check easier with our current design.


dhawton (approved):

leosarra:
I see some reproducible failures on all istio pre/post-submit runs which seems related to this PR (one of the failed runs: https://prow.istio.io/view/gs/istio-prow/logs/integ-ambient-owned-cni_istio_postsubmit/1951877024855363584).
Can you take a look @jaellio ?

PR #5738, opened 2025-09-12
Add job for Ambient mode with nftables backend

### LINKED ISSUE #57653: Add a CI job for validating Ambient use-cases with native nftables
The ambient integration tests cover multiple use-cases and rather than duplicating them a better option is to create a CI job that allows us to run all the existing tests with the nftables backend. This issue is to track the associated work.

sridhargaddam:
Related to: https://github.com/istio/istio/pull/57524
Fixes: https://github.com/istio/istio/issues/57653

### DIFF
--- a/prow/config/jobs/istio.yaml
+++ b/prow/config/jobs/istio.yaml
@@ -234,6 +234,19 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-nftables
+    modifiers: [presubmit_optional]
+    requirements: [kind]
+    env:
+      - name: INTEGRATION_TEST_FLAGS
+        value: "--istio.test.ambient --istio.test.nativeNftables "
+    command:
+      - entrypoint
+      - prow/integ-suite-kind.sh
+      - --kind-config
+      - prow/config/ambient-sc.yaml
+      - test.integration.ambient.kube
+
   - name: integ-ambient-owned-cni
     modifiers: [presubmit_optional]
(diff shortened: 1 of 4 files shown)


keithmattix (approved):

kfaseela (approved):

prow/config/jobs/istio.yaml
```diff
@@ -234,6 +234,19 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-ambient-nftables
+    modifiers: [presubmit_optional]
```
kfaseela, line 238: I merged this, and then started thinking - do we need this on presubmit as optional? Better to have only as postsubmit?

sridhargaddam: > I merged this, and then started thinking - do we need this on presubmit as optional?

Yes, this was intentional. Since nftables is a new feature the idea was to give it some soak time before making it a required job. JFYI, the tests are now passing fine with the Nftables Ambient PR - https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/57524/integ-ambient-nftables_istio/1967111276312137728

> Better to have only as postsubmit?

I'm afraid no. As you know, nftables is an alternative to iptables for setting up traffic redirection rules, and it provides a more modern interface that many Linux distributions are moving toward. To catch any regressions early and ensure new features work with the nftables backend, we would need a presubmit job.

keithmattix: +1 to presubmit

PR #5762, opened 2025-10-29
infra: move to faster c4d machines

howardjohn:
These are very very fast relative to c3d and even n4. Same price. Might as well move everything over, we have plenty of quota.

### DIFF
--- a/infra/gcp/istio-prow-build/cluster-prow.tf
+++ b/infra/gcp/istio-prow-build/cluster-prow.tf
@@ -186,9 +186,9 @@ module "prow_test_c3" {
   location     = "us-west1-a"
 
   machine_type  = "c3-standard-22"
-  initial_count = 1
-  max_count     = 10
-  min_count     = 1
+  initial_count = 0
+  max_count     = 0
+  min_count     = 0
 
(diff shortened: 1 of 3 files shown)


infra/gcp/istio-prow-build/cluster-prow.tf
```diff
@@ -232,3 +232,32 @@ module "prow_test_n4" {
...
+module "prow_test_c4d" {
+  source = "../modules/gke-nodepool"
+
+  name         = "istio-test-pool-c4d"
+  cluster_name = "prow"
+  location     = "us-west1-a"
+
+  machine_type  = "c4d-standard-16"
+  initial_count = 1
+  max_count     = 20
```
keithmattix, line 248: I'm guessing the bump is to account for delays with spot instances?

howardjohn: actually I was more hoping we could move more of our e2 instances off and scale that down. The c4d are much cheaper anyways (since we use spot) and like 5x faster. But didn't want to do it all at once unless I misunderstood the quota we got (gcp has like 5 quotas on cpu :upside_down_face: ) and we don't have enough

keithmattix (approved):

PR #5820, opened 2026-02-02
Always keep 1 node in the nodepool for istio/proxy jobs

keithmattix:
We're seeing a ton of jobs time out on istio/proxy due to momentary lack of capacity in our region. So, despite the 3x cost, we should always keep 1 node around, at least until the 1.29 release is over


fjglira (approved):
LGTM, we need to avoid issues while running release 1.29 jobs

PR #5835, opened 2026-02-25
Add Steven as Infra Admin

Stevenjin8:
Going to be working on migrating away from GCP, so would like access to the API.

I am part of PSGW.

### DIFF
--- a/infra/gcp/common.tf
+++ b/infra/gcp/common.tf
@@ -8,7 +8,6 @@ locals {
     "keithmattix@microsoft.com", # keithmattix
     "daniel.hawton@solo.io",     # dhawton
     "paul@tetrate.io",           # pmerrison 
-    "sjinxuan@microsoft.com",    # Stevenjin8 
   ])
 
   terraform_infra_admins = toset([
@@ -16,6 +15,7 @@ locals {
     "johnbhoward96@gmail.com", # howardjohn
     "daniel.hawton@solo.io",   # dhawton
     "keithmattix2@gmail.com",  # keithmattix
+    "stevenjin8gmail.com",     # Stevenjin8 
   ])
 
 }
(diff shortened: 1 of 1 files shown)


Stevenjin8:
cc @keithmattix

infra/gcp/common.tf
```diff
@@ -8,6 +8,7 @@ locals {
     "keithmattix@microsoft.com", # keithmattix
     "daniel.hawton@solo.io",     # dhawton
     "paul@tetrate.io",           # pmerrison 
+    "sjinxuan@microsoft.com",    # Stevenjin8
```
keithmattix: This is only going to give you access to the private PSWG infrastructure. You need to see the istio-io project for this. I'm also not sure you can sign into GCP without a gmail account

keithmattix: Maybe try modifying [this file](https://github.com/istio/test-infra/blob/c42b4bf7b6092535f2f5eeb90ac7c8cf9bb4b2e1/infra/gcp/istio-io/iam.tf) to give yourself a viewer role

Stevenjin8: I think I got the file right, but the wrong list. The file you linked has no emails

```
resource "google_project_iam_member" "project-admins" {
  for_each = toset(local.terraform_infra_admins)
  project  = local.project_id
  role     = "roles/resourcemanager.projectIamAdmin"
  member   = "user:${each.key}"
}

resource "google_project_iam_member" "owners" {
  for_each = toset(local.terraform_infra_admins)
  project  = local.project_id
  role     = "roles/owner"
  member   = "user:${each.key}"
}
```

Also, the [common.tf](https://github.com/istio/test-infra/blob/c42b4bf7b6092535f2f5eeb90ac7c8cf9bb4b2e1/infra/gcp/istio-io/common.tf) for the istio-io project is just a symlink to the file I edited.

keithmattix: Yes but that gives you admin access. If you just want viewer you'll need to create a custom iam role in terraform and add yourself as a user there

keithmattix (approved):

PR #5846, opened 2026-03-09
Make ambient multi-cluster tests mandatory

krinkinmu:
Ambient multi-network multicluster feature is now in beta, so we want to make the integration tests covering it mandatory.

We've fixed all the known issues with the enabled tests and also reduced flakiness of the ambient tests for both multi-cluster and single-cluster cases.

At this point, I believe that additional tests will not introduce any additional significant burden on the community and will help to prevent degradation of a beta feature in the future.

+cc @grnmeira @keithmattix @therealmitchconnors

### DIFF
--- a/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -3916,7 +3916,6 @@ presubmits:
     cluster: private
     decorate: true
     name: integ-ambient-mc_istio_pri
-    optional: true
     path_alias: istio.io/istio
     rerun_command: /test integ-ambient-mc
     spec:
--- a/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
+++ b/prow/cluster/jobs/istio/istio/istio.istio.master.gen.yaml
@@ -3373,7 +3373,6 @@ presubmits:
     - ^master$
     decorate: true
     name: integ-ambient-mc_istio
-    optional: true
     path_alias: istio.io/istio
     rerun_command: /test integ-ambient-mc
     spec:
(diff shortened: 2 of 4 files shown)


dhawton:
> Ambient multi-network multicluster feature is now in beta, so we want to make the integration tests covering it mandatory.
> 
> We've fixed all the known issues with the enabled tests and also reduced flakiness of the ambient tests for both multi-cluster and single-cluster cases.
> 
> At this point, I believe that additional tests will not introduce any additional significant burden on the community and will help to prevent degradation of a beta feature in the future.
> 
> +cc @grnmeira @keithmattix @therealmitchconnors

With it being in beta as of 1.29, can you also update the 1.29 job to have it be non-optional?

krinkinmu:
> > Ambient multi-network multicluster feature is now in beta, so we want to make the integration tests covering it mandatory.
> > We've fixed all the known issues with the enabled tests and also reduced flakiness of the ambient tests for both multi-cluster and single-cluster cases.
> > At this point, I believe that additional tests will not introduce any additional significant burden on the community and will help to prevent degradation of a beta feature in the future.
> > +cc @grnmeira @keithmattix @therealmitchconnors
> 
> With it being in beta as of 1.29, can you also update the 1.29 job to have it be non-optional?

@dhawton I'd love to, but can you give me some hints on how to go about it?

I looked at the configs for the master branch and what is generated from those, but I don't see any similar modifiers or optional fields for the multicluster tests on 1.29 configs. I went through readmes and tried to find PRs that did something similar, but I just cannot figure out what part of the configuration actually makes these tests optional on 1.29 release branch.

UPDATE: I tried to add a `presubmit_optional` modifier to 1.29 configs and regenerate and with that I can see jobs marked as optional, so it looks as if according to the current configuration these tests should not be optional on 1.29 - I'm clearly missing something here.

keithmattix:
Yeah it looks like when the branch was cut, the optional modifier wasn't persisted. Which actually increases my confidence in the change on master

keithmattix (approved):

PR #5865, opened 2026-03-23
Exclude copilot from branch protections

keithmattix:
Don't activate branch protections for copilot branches. We will still require explicit approval for GH actions and prow to execute though

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -153,6 +153,8 @@ branch-protection:
       required_status_checks:
         contexts:
         - EasyCLA
+      exclude:
+        "copilot/" # Exclude copilot created branches from branch protection
       # By default, required 1 approval from a CODEOWNER
       required_pull_request_reviews:
         required_approving_review_count: 1
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -153,8 +153,6 @@ branch-protection:
       required_status_checks:
         contexts:
         - EasyCLA
-      exclude:
-      - "copilot/" # Exclude copilot created branches from branch protection
       # By default, required 1 approval from a CODEOWNER
       required_pull_request_reviews:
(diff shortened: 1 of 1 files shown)


dhawton (approved):
LGTM, put a DNM in case @howardjohn has some input

howardjohn (commented):
hmm I am a bit confused. I was on board with allowing co-pilot but this seems to disable all checks on copilot which seems quite wrong?

I can now just create a `copilot/blah` branch and have no checks and merge anything? Ah but only to the `copilot/blah` branch... does copilot not have the ability to work in forks like a normal workflow? or specifically on copilot modifying dependabot there is no fork to use at all?

dhawton:
> hmm I am a bit confused. I was on board with allowing co-pilot but this seems to disable all checks on copilot which seems quite wrong?
> 
> I can now just create a `copilot/blah` branch and have no checks and merge anything? Ah but only to the `copilot/blah` branch... does copilot not have the ability to work in forks like a normal workflow? or specifically on copilot modifying dependabot there is no fork to use at all?

Copilot, like dependabot, don't do any forks, they only work on branches within the repo. The checks and merges would still need PR approval to go into the core branches, though we might be able to configure so copilot/ branches have their own set of rules, so that only copilot can push to copilot/ prefixed branches.

Maybe that might work better actually, thinking about it. Copilot/ fixed branches is restricted to copilot, dependabot/ prefixed branches are restricted to dependabot?

keithmattix:
And members? I imagine we want folks to be able to push to their copilot branch? Or no?

keithmattix:
Actually tried to do this and turns out this isn't possible since copilot at least isn't really a user that you can give push access to (I tried in the UI). So think the best we can do is just exclude those branches in general

PR #5868, opened 2026-04-07
remove 1.27 jobs

zirain:


dhawton (changes requested):
Don't delete from sail-operator, this is an istio-ecosystems repo and handled by them.

dhawton (approved):

PR #5878, opened 2026-04-13
Use wildcards instead of regex

Stevenjin8:
Apparently, our Cloudflare plan does not support regex... If this fails, we can try remove the firxt n characters of the path...

### DIFF
--- a/infra/cf/cncf-istio/r2.tf
+++ b/infra/cf/cncf-istio/r2.tf
@@ -97,14 +97,14 @@ resource "cloudflare_ruleset" "blob_redirect" {
 
   rules = [for bucket in local.r2_public_buckets : {
     ref         = "blob_redirect_${bucket}"
-    description = "Redirect blob.istio.io/${bucket}/* to ${bucket}.r2.istio.io/*"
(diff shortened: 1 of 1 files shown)


infra/cf/cncf-istio/r2.tf
```diff
@@ -97,14 +97,14 @@ resource "cloudflare_ruleset" "blob_redirect" {
 
   rules = [for bucket in local.r2_public_buckets : {
     ref         = "blob_redirect_${bucket}"
-    description = "Redirect blob.istio.io/${bucket}/* to ${bucket}.r2.istio.io/*"
+    description = "Redirect blob.istio.io/${bucket}/* to ${bucket}.r2.istio.io/${1}"
     expression  = "(http.host eq \"blob.istio.io\" and starts_with(http.request.uri.path, \"/${bucket}/\"))"
     action      = "redirect"
     action_parameters = {
       from_value = {
-        status_code = 301
```
Stevenjin8, line 105: 308 doesnt change the method of the request

keithmattix (approved):

PR #5881, opened 2026-04-14
add r2-related secrets to prow setup

Stevenjin8:
I have created the secrets in the istio-testing GCP project. This makes the release postsubmit job reference these secrets, we can then configure the release job to use these secrets to upload to r2.


howardjohn (approved):

prow/config/jobs/istio.yaml
```diff
@@ -2,6 +2,18 @@ org: istio
 repo: istio
 support_release_branching: true
 image: gcr.io/istio-testing/build-tools:master-817db1b7909fb411f9c6d357b831b658285c26fb
+requirement_presets:
+  cf-r2-auth:
+    env:
+      - name: CF_ACCOUNT_ID
+        value: "1eb8152ceed73d3ee6f66c3557819d4c"
```
kfaseela, line 9: hope it is appropriate to hardcode

Stevenjin8: yeah its the account id, so not a secret

PR #5882, opened 2026-04-14
Change Istio release registry from Docker to GCR

Stevenjin8:
We get DDOS rate limited by docker. This caused some degredation of istio.io for ~4.5 starting today at noon EST until 4:30 EST


keithmattix (approved):
We should follow up with the docker folks on this?

PR #5887, opened 2026-04-20
Add R2 secrets to ztunnel and proxy postsubmit publish jobs

Stevenjin8:
This changes the name of secrets. I think we should have one credential set for the public buckets (istio-build) and one for the private ones. (or maybe we just want one) regardless, I think its better to rename from `cf_r2_istio-build_...` to `cf_r2_...` now.

This also adds these secrets to ztunnel and proxy release jobs so they can upload binaries.


prow/config/jobs/istio.yaml
```diff
@@ -2,18 +2,6 @@ org: istio
 repo: istio
 support_release_branching: true
 image: gcr.io/istio-testing/build-tools:master-817db1b7909fb411f9c6d357b831b658285c26fb
-requirement_presets:
```
Stevenjin8, line 5: this just go moved to .base.yaml so we can share it better.

prow/cluster/jobs/istio-private/proxy/istio-private.proxy.master.gen.yaml
```diff
@@ -88,6 +88,8 @@ postsubmits:
         - name: BAZEL_BUILD_RBE_INSTANCE
         - name: BUILD_WITH_CONTAINER
           value: "0"
+        - name: CF_ACCOUNT_ID
+          value: 1eb8152ceed73d3ee6f66c3557819d4c
```
Stevenjin8, line 92: This is just the account id, which we will end up sharing with the private credentials anyways

prow/config/jobs/ztunnel.yaml
```diff
@@ -23,8 +23,8 @@ jobs:
     architectures: [amd64, arm64]
     service_account_name: prowjob-testing-write
     types: [postsubmit]
-    command: [make, release]
-    requirements: [cratescache]
+    command: [entrypoint, make, release]
```
Stevenjin8, line 26: so secrets get picked up

keithmattix (approved):

Stevenjin8:
going to terraform apply

PR #5893, opened 2026-04-20
External secrets in build clusters

Stevenjin8:
Ok so the setup is a little weird, but we have a "control" prow cluster in the `istio-testing/prow` cluster and we have builder/worker clusters in `istio-prow-build/{prow,arm}`. Right now, we only create the external secret on `istio-testing/prow`, but but tests/CI run in `istio-prow-build/{prow,arm}`. Even worse, we don't have an external secret operator in `istio-prow-build/{prow,arm}`. So pods in `istio-prow-build` can't reference the necessary secrets to start.

This PR add the external secret operator and the corresponding workload identity bindings. It also moves the external secrets around a bit.


Stevenjin8:
i manually edited `default/config` so ci would run.

prow/cluster/arm/kubernetes-external-secrets_crd.yaml
```diff
@@ -0,0 +1,238 @@
+---
```
Stevenjin8, line 1: same as  prow/cluster/kubernetes-external-secrets_crd.yaml

prow/cluster/arm/kubernetes-external-secrets_deployment.yaml
```diff
@@ -0,0 +1,41 @@
+---
```
Stevenjin8, line 1: same as prow/cluster/kubernetest-external-secrets_deployment.yaml

prow/cluster/arm/kubernetes-external-secrets_rbac.yaml
```diff
@@ -0,0 +1,67 @@
+apiVersion: v1
```
Stevenjin8, line 1: same as prow/cluster/kubernetest-external-secrets_rbac.yaml, but with a slightly different GCP SA binding

prow/cluster/arm/kubernetes-external-secrets_service.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
Stevenjin8, line 1: same as same as prow/cluster/kubernetest-external-secrets_service.yaml

howardjohn (commented):
This is 100% working as designed. The idea was to not use secrets and instead store everything in GCP secrets manager and lookup at runtime

TBH I forget _why_ though...

Stevenjin8:
@howardjohn we can do that for the jobs that produce artifacts (postsubmit, etc), but I don't see a great way of doing this given how the prow sidecars/initContainers work. Not to mention that we're being very explicit about which secrets the worker clusters can access.

Stevenjin8:
The only reason we didn't ExternalSecrets is because we have workload identity enabled, so the kubernetes service account on the pod can authenticate to GCS. As we remove our infra out of the GCP ecosystem, I think we're going to have to have the external secrets operator.

Stevenjin8:
@howardjohn TLDR: we only removed teh External Secrets Operator because there were not secrets to sync, not some fundamental security risk the ESO posed.

I looked a bit #4863.

Before #4863, we had the External Secrets Operator (ESO) in all our clusters because we ran postsubmit community jobs in the untrusted clusters, and that needed a fairly privileged GitHub token. This was the only external secret

#4863 moved this job to a trusted cluster. Now, the ESO in the untrusted cluster was unecesssary. There were not external secrets to sync. So you removed it to "simplify things".

I think we need to add it back now. I (think I) am being fairly careful and granting access to one secret. And (as of a few minutes ago) that secret only has access to the istio-build, istio-prow, and istio-testgrid buckets.

Now, from reading #2526, there are a lot of reasons to prefer WI over GCP secrets, but I don't think we can have prow use WI to authenticate to R2. We need static credentials for R2.

keithmattix (approved):

PR #5898, opened 2026-04-23
Cloudflare secret rotation

Stevenjin8:
Ok, after a few too many PRs and some WG meeting discussion I think I've settled on a good way to do secrets.

We have one set of admin api credientials in google secret manager that has read write access to all buckets in cloudflare (Cloudflare doesn't have bucket-level granularity for admin credentials; also, only admin credential keys can create temp credential keys which are scoped). 

Now, we have a secret rotator that runs in our trusted cluster. This cronjob reads the admin api credentials and creates temp read-write credentials (24h) for each bucket we care about (right now, istio-prow and istio-build). The secret rotator uploads these secrets to Google Secret Manager in the form

```
{
  "access_key": ...
  "secret_key": ...
  "session_token": ...
  "endpoint": ...
  "s3_force_path_style": true,
  "region": auto
}
```

this is the format prow expects and by putting all our sesitive values in one secret, we have fewer secrets to worry about. We have more control over non-prow jobs, so we can have those jobs parse secrets as necessary.

I'll also put this in the parent issue.

### DIFF
--- a/infra/gcp/istio-testing/iam.tf
+++ b/infra/gcp/istio-testing/iam.tf
@@ -151,24 +151,13 @@ module "cloudflare_rotator_account" {
   // secretAccessor (read) on the permanent admin token and on the
   secrets = [
     { name = "cf_r2_admin_token" },
-    { name = "cf_r2_istio-prow_access_key_id" },
   ]
 }
 
-// secretVersionManager (add new versions + enable/disable/destroy old ones)
-// on the per-bucket ephemeral secrets. Manager is needed (rather than the
-// append-only secretVersionAdder) because the rotator disables prior versions
-// after each successful upload to keep stale credentials from being read.
-//
-// Note: cf_r2_istio-prow_access_key_id is intentionally omitted. Cloudflare's
-// temp-access-credentials endpoint returns the parent's accessKeyId unchanged,
-// so that secret only needs to be read (granted via the module above) and is
-// never rewritten by the rotator.
 locals {
   cloudflare_rotator_writable_secrets = toset([
-    "cf_r2_istio-prow_access_key_secret",
     "cf_r2_istio-prow_credentials",
-    "cf_r2_istio-prow_session_token",
+    "cf_r2_istio-build_credentials",
   ])
 }
 
--- a/infra/gcp/istio-testing/iam.tf
+++ b/infra/gcp/istio-testing/iam.tf
@@ -139,16 +139,16 @@ resource "google_project_iam_member" "owners" {
 }
 
 // GSA used by the cloudflare-rotator CronJob in the trusted cluster.
-// The KSA is "cloudflare-rotator" in the "default" namespace.
+// The KSA is "cloudflare-rotator" in the "cloudflare-secret-rotation" namespace.
 module "cloudflare_rotator_account" {
   source            = "../modules/workload-identity-service-account"
   project_id        = local.project_id
   name              = "cloudflare-rotator"
   description       = "Rotates ephemeral Cloudflare R2 credentials in GSM."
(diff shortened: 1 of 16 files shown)


infra/gcp/istio-testing/iam.tf
```diff
@@ -138,3 +138,33 @@ resource "google_project_iam_member" "owners" {
   member   = "user:${each.key}"
 }
 
+// GSA used by the cloudflare-rotator CronJob in the trusted cluster.
+// The KSA is "cloudflare-rotator" in the "default" namespace.
```
keithmattix: nit: can we use a more locked down namespace? With RBAC controls locking it down to only certain GCP accounts? How much work would that be

prow/cluster/cloudflare_rotator_cronjob.yaml
```diff
@@ -0,0 +1,51 @@
...
+      ttlSecondsAfterFinished: 86400
+      template:
+        spec:
+          serviceAccountName: cloudflare-rotator
+          restartPolicy: Never
+          nodeSelector:
+            kubernetes.io/arch: amd64
+          containers:
+            - name: rotator
+              image: gcr.io/istio-testing/build-tools:master-63fd6eec6f3df5a3ed4e190e60153ef3425f66b4
```
keithmattix, line 25: nit: can we use registry.istio.io for this yet? Or add a TODO to come back and swap it out?

infra/gcp/istio-testing/iam.tf
Stevenjin8: Just updated. You can pass the module a `cluster_serviceaccount_name` argument to specify which Kubernetes service accounts can bind to the GCP service account, but if you don't pass it anything, it will default to the `name`. So right now, only `cloudflare-secret-rotation/cloudflare-rotator` can bind to the `cloudflare-rotator` gcp service account.

prow/cluster/cloudflare_rotator_cronjob.yaml
Stevenjin8: I updated `tools/automator/scripts/update-images.sh` to include thispath, so whenever we switch to `registry.istio.io`, this will get automatically updated.

infra/gcp/istio-testing/iam.tf
keithmattix: Sounds good; the namespace is still default though unless I'm missing something?

Stevenjin8: I just changed it to `cloudflare-secret-rotation`

infra/gcp/istio-testing/iam.tf
```diff
@@ -138,3 +138,33 @@ resource "google_project_iam_member" "owners" {
...
+module "cloudflare_rotator_account" {
+  source            = "../modules/workload-identity-service-account"
+  project_id        = local.project_id
+  name              = "cloudflare-rotator"
+  description       = "Rotates ephemeral Cloudflare R2 credentials in GSM."
+  cluster_namespace = "default"
```
keithmattix: I still see default here

prow/cluster/cloudflare_rotator_configmap.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cloudflare-rotator
+  namespace: default
```
keithmattix: And default namespace here

Stevenjin8: my bad didn't push

keithmattix (approved):

PR #5905, opened 2026-04-27
remove node_count for prow node pool

Stevenjin8:
IDK how this wasn't a problem before, but terraform doesn't let me create a node pool with both a `node_count` and `initial_node_count`.it also doesnt make much sense

### DIFF
--- a/infra/gcp/istio-testing/cluster-prow.tf
+++ b/infra/gcp/istio-testing/cluster-prow.tf
@@ -115,7 +115,6 @@ resource "google_container_node_pool" "prow_pool" {
     }
   }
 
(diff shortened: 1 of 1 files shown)


kfaseela:
node_count = 4 was originally added when cluster-prow.tf was introduced in the terraform backfill PR: https://github.com/istio/test-infra/pull/4762. With autoscaling enabled (min_node_count = 4, max_node_count = 8), pinning node_count to 4 doesn’t make much sense since the autoscaler should manage desired size. So this change LGTM

keithmattix (approved):

PR #5906, opened 2026-04-27
Fix permissions to secret rotator service account

Stevenjin8:
We need to give the secret rotator read permissions to the secrets.

Also, clean up the secret rotation script a little.


prow/cluster/cloudflare_rotator_cronjob.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: cloudflare-rotator
   namespace: cloudflare-secret-rotation
 spec:
-  schedule: "0 * * * *"  # every hour while debugging; switch back to "0 */12 * * *" once stable. 
+  schedule: "0 */12 * * *"  # every hour while debugging; switch back to "0 */12 * * *" once stable.
```
keithmattix, line 7: Nit: remove the comment since we made the change

keithmattix (approved):

PR #5908, opened 2026-04-28
Bump some images

Stevenjin8:
Some images are out of date, so we should bump them


prow/cluster/gcsweb/gcsweb_deployment.yaml
```diff
@@ -23,7 +23,7 @@ spec:
       terminationGracePeriodSeconds: 30
       containers:
       - name: nginx-proxy
-        image: nginx:1.21-alpine
+        image: nginx:1.30-alpine
```
keithmattix, line 26: nit: we should probably move away from dockerhub images

keithmattix (approved):

PR #5909, opened 2026-04-28
Give service account secret access to r2 secrets

Stevenjin8:
TBH, I think a better option is to remove the `kubernetes.io/enforce-mountable-secrets: "true"`. We don't have any secrets and the annotation is deprecated https://kubernetes.io/docs/concepts/security/service-accounts/#enforce-mountable-secrets


Stevenjin8:
btw, I already applied this and CI is working on CF

keithmattix (approved):

PR #5914, opened 2026-05-04
Use long lived tokens for some of prow. Fix health

Stevenjin8:
So we had to update our gcsweb version, which no longer had a healthz endpoint.
The GCP load balancer healthchecker was a bit disappointing started defaulting to `/` and failing because it returned a 404
I just decided to hardcode it to /robots.txt to not have to deal with it guessing.

Also, lots of our images still havne't picked up the changes that allow using temp s3 tokens... So for now (and for ONLY PROW) we use a long-lived token.

### DIFF
--- a/infra/gcp/istio-testing/iam.tf
+++ b/infra/gcp/istio-testing/iam.tf
@@ -139,13 +139,13 @@ resource "google_project_iam_member" "owners" {
 }
 
 // We need to read from these secrets to get the key id, which lets us get new ephemeral credentials and refresh the secrets.
+// TODO Also rotate cf_r2_istio-prow_credentials when the rest of prow supports short-lived creds
 locals {
   cloudflare_rotator_r2_secrets = toset([
     "cf_r2_istio-build_credentials",
     "cf_r2_istio-build-private_credentials",
     "cf_r2_istio-prerelease_credentials",
     "cf_r2_istio-prerelease-private_credentials",
-    "cf_r2_istio-prow_credentials",
     "cf_r2_istio-prow-private_credentials",
     "cf_r2_istio-testgrid_credentials",
   ])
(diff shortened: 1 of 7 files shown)


infra/gcp/istio-testing/iam.tf
```diff
@@ -139,13 +139,13 @@ resource "google_project_iam_member" "owners" {
 }
 
 // We need to read from these secrets to get the key id, which lets us get new ephemeral credentials and refresh the secrets.
+// TODO Also rotate cf_r2_istio-prow_credentials when the rest of prow supports short-lived creds
```
Stevenjin8, line 142:
```suggestion
// TODO (Stevenjin8) Also rotate cf_r2_istio-prow_credentials when the rest of prow supports short-lived creds
```

prow/cluster/gcsweb/gcsweb_backendconfig.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: cloud.google.com/v1
+kind: BackendConfig
+metadata:
+  name: gcsweb
+  namespace: gcs
+spec:
+  # Just explicitly set the LB healthcheck path to not have to deal with
+  # back inferences (e.g. defaults to /)
```
Stevenjin8, line 8:
```suggestion
  # healthcheck inferences (e.g. defaults to /)
```

prow/cluster/deck_deployment.yaml
```diff
@@ -68,8 +69,8 @@ spec:
         - mountPath: /etc/kubeconfig
           name: kubeconfig
           readOnly: true
-        - mountPath: /etc/cf-r2-istio-prow-credentials
-          name: cf-r2-istio-prow-credentials
+        - mountPath: /etc/cf-r2-public-buckets-ro-credentials
```
Stevenjin8, line 72: deck doesnt support long lived tokens. We should give it RO tokens anyways.

infra/gcp/istio-testing/iam.tf
Stevenjin8: For more context, crier doesn't yet support temp tokens. It should just be a version bump upstream.

keithmattix (approved):

PR #5928, opened 2026-05-21
arm prow pool: drop spot to fix preemption-driven test flakes

PetrMc:
The c4a arm test pool is spot-only, while the amd64 cluster has a non-spot backbone. Across recent master PRs, arm jobs failed ~21% of attempts vs ~1.5% on amd64, with ~37% of arm failures being mid-run SIGTERM (Entrypoint received interrupt, no preceding error, completion_time null) consistent with spot preemption. Removing spot=true so arm jobs land on on-demand c4a-standard-16 instances. Pool name kept as c4a-16-spot for now to avoid destroy+recreate; rename can follow once stable. Confirm c4a-standard-16 quota in us-central1-f is sufficient before merging.

### DIFF
--- a/infra/gcp/istio-prow-build/cluster-prow-arm.tf
+++ b/infra/gcp/istio-prow-build/cluster-prow-arm.tf
@@ -125,8 +125,6 @@ module "prow_arm_test_spot_preview" {
 
   arm          = true
   machine_type = "c4a-standard-16"
-  # Spot instances are used as quota is capped for ARM nodes, and its cheaper.
-  spot = true
 
   service_account = "istio-prow-jobs@istio-prow-build.iam.gserviceaccount.com"
 


kfaseela:
@PetrMc what is the last line in the description? "Confirm c4a-standard-16 quota in us-central1-f is sufficient before merging."

PetrMc:
> @PetrMc what is the last line in the description? "Confirm c4a-standard-16 quota in us-central1-f is sufficient before merging."

Thanks @kfaseela Spot and on-demand c4a quotas in istio-prow-build are separate in GCP. After this change, scale-up needs on-demand c4a vCPU quota in us-central1-f to cover max_count * 16 = 128. Someone with project-level access needs to confirm or request a raise. My assumption it's already there but being careful here.

fjglira:
Spot instances can be claimed at any time, causing failures but usually costing up to 60% less than regular instances. We need to be sure that failures are caused by these claims of the spot clusters and that we can spend more also  by using regular instances on arm jobs also (I don't know how this is managed to be honest)

PetrMc:
thanks @fjglira few points to support this change.

- "60%" math was true when arm was expensive (when this was added ~3 years ago). I think these days c4a, on-demand arm is roughly on par with on-demand amd per vCPU. we can analyze more here [GCP pricing](https://cloud.google.com/compute/all-pricing) but the original premise no longer holds.

- the change would bing `arm` in line with how `amd` already runs: the workhorse amd pool [istio-test-pool-e2](https://github.com/istio/test-infra/blob/master/infra/gcp/istio-prow-build/cluster-prow.tf) (max 60 nodes) is non-spot. Only the small specialized `amd` pools (c3/n4/c4d) are spot. `arm` currently has no non-spot fallback at all.

- found plenty of concrete preemption signature and the run you just retested https://prow.istio.io/view/s3/istio-prow/pr-logs/pull/istio_istio/60341/integ-basic-arm64_istio_release-1.30/2058853677350785024 is exactly the shape that needs the  fix. you can see that the tests progressing normally through `TestGatewayConformance` for 13 minutes, then abrupt Entrypoint received interrupt: terminated, no preceding failure, `finished.json: ABORTED` while prow records state: failure. seems to me that mismatch points to the pod-level SIGTERM not initiated by prow. Tide-aborts carry description: "Aborted by trigger plugin.", this one says "Job failed.".

- the `ar`m pool already has known reliability issues. there is the terraform comment for the pool that says "autoscaling is disabled due to ongoing networking issues on ARM." Stacking spot on top of that just compounds the problem.
- also lets look of the hidden cost when we spin `arm` instance over and over to /retest plus our engineer-hours easily dominate any vCPU savings.

hope it justifies.

howardjohn:
Spot was used not because of ARM pricing but because of quota (impossible
to get non-spot quota) and a general desire to reduce cost (not specific to
ARM). the ARM networking issues are also resolved.

FWIW we are moving off GCP soon anyways so this may not matter too much.

But main concern is if we can get quota

PetrMc:
thanks @howardjohn - closing lets see if it gets better when we are migrated off GCP.

PR #5936, opened 2026-06-02
prow: add optional IPv6 docs CI job for istio.io

kfaseela:
Adds a new optional presubmit job `doc.test.ipv6.profile-default` that runs the standard `doc.test.profile-default` test suite against an IPv6-only kind cluster (`IP_FAMILIES=IPv6`).

The job is marked `presubmit_optional` + `presubmit_skipped` so it does not block PRs while we validate coverage. It can be triggered manually with:

  /test doc.test.ipv6.profile-default

Part of the IPv6 promote-to-Beta effort tracked in istio/istio#60425.

### DIFF
--- a/prow/cluster/jobs/istio-private/istio.io/istio-private.istio.io.master.gen.yaml
+++ b/prow/cluster/jobs/istio-private/istio.io/istio-private.istio.io.master.gen.yaml
@@ -87,7 +87,7 @@ postsubmits:
           value: "8"
         - name: IP_FAMILIES
           value: IPv6
-        image: gcr.io/istio-testing/build-tools:master-6de2ce5813071b34e0ca033dbac7f79ffc1644be
+        image: gcr.io/istio-testing/build-tools:master-1aae82591b3f663bba15d501ae7ce9c2d329aa70
         name: ""
         resources:
           limits:
@@ -105,9 +105,6 @@ postsubmits:
         - mountPath: /lib/modules
           name: modules
           readOnly: true
-        - mountPath: /sys/fs/cgroup
-          name: cgroup
-          readOnly: true
         - mountPath: /var/lib/docker
           name: docker-root
       nodeSelector:
@@ -122,10 +119,6 @@ postsubmits:
           path: /lib/modules
           type: Directory
         name: modules
-      - hostPath:
-          path: /sys/fs/cgroup
-          type: Directory
-        name: cgroup
       - emptyDir: {}
         name: docker-root
   - annotations:
@@ -648,7 +641,7 @@ presubmits:
           value: "8"
         - name: IP_FAMILIES
           value: IPv6
-        image: gcr.io/istio-testing/build-tools:master-6de2ce5813071b34e0ca033dbac7f79ffc1644be
+        image: gcr.io/istio-testing/build-tools:master-1aae82591b3f663bba15d501ae7ce9c2d329aa70
         name: ""
         resources:
           limits:
@@ -666,9 +659,6 @@ presubmits:
         - mountPath: /lib/modules
           name: modules
           readOnly: true
-        - mountPath: /sys/fs/cgroup
-          name: cgroup
-          readOnly: true
         - mountPath: /var/lib/docker
           name: docker-root
       nodeSelector:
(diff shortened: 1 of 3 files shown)


fjglira (approved):
LGTM, let's merge the job and check if with the default profile and the current test will work on a only IPv6 cluster

dhawton:
We probably do not want `presubmit_skipped`, let's have it run on PRs so we *can* validate coverage. Having it optional will already keep PRs from being blocked.

This will allow someone to open a PR to fix what might need to be fixed, and keep retesting until it's complete, instead of only having it be a postsubmit job requiring PRs to merge to run. Our general procedure is marking optional only.

fjglira:
> We probably do not want `presubmit_skipped`, let's have it run on PRs so we _can_ validate coverage. Having it optional will already keep PRs from being blocked.
> 
> This will allow someone to open a PR to fix what might need to be fixed, and keep retesting until it's complete, instead of only having it be a postsubmit job requiring PRs to merge to run. Our general procedure is marking optional only.

But this configuration will not make the job a postsubmit job only. According to the documentation that we have, it mentions:

```
modifiers: change properties of the job. Possible values are:
* presubmit_skipped: the test will only be run in presubmit by explicitly calling /test on it
* presubmit_optional: the test will not be required in presubmit
```

So, meanwhile we are sure that the test works we can create a fake PR or use any PR and run the test using the /test

keithmattix:
I lean more towards Daniel's perspective; let's run this in presubmit all the time but just not block. That will give us better signal

dhawton:
> > We probably do not want `presubmit_skipped`, let's have it run on PRs so we _can_ validate coverage. Having it optional will already keep PRs from being blocked.
> > This will allow someone to open a PR to fix what might need to be fixed, and keep retesting until it's complete, instead of only having it be a postsubmit job requiring PRs to merge to run. Our general procedure is marking optional only.
> 
> But this configuration will not make the job a postsubmit job only. According to the documentation that we have, it mentions:
> 
> ```
> modifiers: change properties of the job. Possible values are:
> * presubmit_skipped: the test will only be run in presubmit by explicitly calling /test on it
> * presubmit_optional: the test will not be required in presubmit
> ```
> 
> So, meanwhile we are sure that the test works we can create a fake PR or use any PR and run the test using the /test

Yes, but why? It's a good signal to catch issues *now*. You can have a PR try to fix up existing documentation, as well as signal new updates created while we're testing the test also potentially see their changes might fail this test and can fix those at that point as well.

kfaseela:
> > > We probably do not want `presubmit_skipped`, let's have it run on PRs so we _can_ validate coverage. Having it optional will already keep PRs from being blocked.
> > > This will allow someone to open a PR to fix what might need to be fixed, and keep retesting until it's complete, instead of only having it be a postsubmit job requiring PRs to merge to run. Our general procedure is marking optional only.
> > 
> > 
> > But this configuration will not make the job a postsubmit job only. According to the documentation that we have, it mentions:
> > ```
> > modifiers: change properties of the job. Possible values are:
> > * presubmit_skipped: the test will only be run in presubmit by explicitly calling /test on it
> > * presubmit_optional: the test will not be required in presubmit
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So, meanwhile we are sure that the test works we can create a fake PR or use any PR and run the test using the /test
> 
> Yes, but why? It's a good signal to catch issues _now_. You can have a PR try to fix up existing documentation, as well as signal new updates created while we're testing the test also potentially see their changes might fail this test and can fix those at that point as well.

OK @dhawton ! Let us do that way! I have not run this locally - so thought of trying some basic runs and validations before making it live on every presubmit.. Did not want people to get annoyed seeing a failing job on every PR. wanted to gain some confidence before doing so. But I am okay with this approach as well!

kfaseela:
Done @dhawton @keithmattix

keithmattix (approved):

dhawton:
> OK @dhawton ! Let us do that way! I have not run this locally - so thought of trying some basic runs and validations before making it live on every presubmit.. Did not want people to get annoyed seeing a failing job on every PR. wanted to gain some confidence before doing so. But I am okay with this approach as well!

docs is quieter than main istio, I'd rather we work through it personally.

PR #5937, opened 2026-06-03
Add integ-gateway-api-only prow job

jaiveerk:
New presubmit_skipped job that runs `make
test.integration.pilot.gateway-api-only.kube` against istio's master branch. The job is opt-in via `/test integ-gateway-api-only` on a PR — once we have confidence in stability we can flip to optional/required in a follow-up.

The make target (added on the istio side) runs the gateway-API-relevant subset of the pilot suite under --istio.test.gatewayAPIOnly with a non-default GatewayClass (istio-alt): TestGatewayConformance, TestGateway, TestHTTPRouteAndGRPCRouteCoexistence, TestInferencePoolMultipleTargetPorts. This is designed to track regressions for "Gateway API Only" mode.


prow/config/jobs/istio.yaml
```diff
@@ -331,6 +331,17 @@ jobs:
     - prow/config/ambient-sc.yaml
     - test.integration.ambient.kube
 
+  - name: integ-gateway-api-only
+    modifiers: [presubmit_skipped]
```
dhawton:
```suggestion
    modifiers: [presubmit_optional]
```

dhawton:
> New presubmit_skipped job that runs `make test.integration.pilot.gateway-api-only.kube` against istio's master branch. The job is opt-in via `/test integ-gateway-api-only` on a PR — once we have confidence in stability we can flip to optional/required in a follow-up.
> 
> The make target (added on the istio side) runs the gateway-API-relevant subset of the pilot suite under --istio.test.gatewayAPIOnly with a non-default GatewayClass (istio-alt): TestGatewayConformance, TestGateway, TestHTTPRouteAndGRPCRouteCoexistence, TestInferencePoolMultipleTargetPorts. This is designed to track regressions for "Gateway API Only" mode.

It's better to have it run so it can be tested more thoroughly and easily is the preferred and standard flow we have. We start with optional, work out the kinks, then make it required.

prow/config/jobs/istio.yaml
jaellio: @jaiveerk I agree with Daniel on this change. Let's make it optional

jaellio (commented):
Could you explain the reasoning behind the non default gateway class name? Is it to verify Istio still behaves correctly when the gateway class name is customized or is it required?

jaiveerk:
> Could you explain the reasoning behind the non default gateway class name? Is it to verify Istio still behaves correctly when the gateway class name is customized or is it required?

Yeah it's just to verify that it behaves correctly when gwclass name is customized. It's not required, and the tests fall back to just using gwclass istio when it's not specified.

prow/config/jobs/istio.yaml
jaiveerk: Cool! Updated.

jaellio (approved):

prow/config/jobs/istio.yaml
```diff
@@ -331,6 +331,17 @@ jobs:
...
+  - name: integ-gateway-api-only
+    modifiers: [presubmit_optional]
+    requirements: [kind]
+    env:
+    - name: INTEGRATION_TEST_FLAGS
+      value: "--istio.test.gatewayAPIOnly --istio.test.kube.gatewayClassName=istio-alt"
+    command:
+    - entrypoint
+    - prow/integ-suite-kind.sh
+    - test.integration.pilot.gateway-api-only.kube
```
Stevenjin8: I dont think this exists.

Stevenjin8 (changes requested):
_see comment

prow/config/jobs/istio.yaml
jaiveerk: yeah adding it in the actual istio pr. Just made that -- https://github.com/istio/istio/pull/60453/changes

prow/config/jobs/istio.yaml
```diff
@@ -337,10 +337,12 @@ jobs:
     env:
     - name: INTEGRATION_TEST_FLAGS
       value: "--istio.test.gatewayAPIOnly --istio.test.kube.gatewayClassName=istio-alt"
+    - name: T
+      value: '-run=TestGatewayConformance|TestGateway$$|TestHTTPRouteAndGRPCRouteCoexistence|TestInferencePoolMultipleTargetPorts'
     command:
     - entrypoint
     - prow/integ-suite-kind.sh
-    - test.integration.pilot.gateway-api-only.kube
+    - test.integration.pilot.kube
```
Stevenjin8, line 345: you only want the gwapi tests and the gie test right? In this case, I think you can add `SINGLE_PACKAGE=true` in your env and also pass in test.integration.pilot.kube.gie as arguments

prow/config/jobs/istio.yaml
```diff
@@ -337,10 +337,12 @@ jobs:
     env:
     - name: INTEGRATION_TEST_FLAGS
       value: "--istio.test.gatewayAPIOnly --istio.test.kube.gatewayClassName=istio-alt"
+    - name: T
+      value: '-run=TestGatewayConformance|TestGateway$$|TestHTTPRouteAndGRPCRouteCoexistence|TestInferencePoolMultipleTargetPorts'
```
Stevenjin8:
```
      value: '-run="TestGatewayConformance|TestGateway$$|TestHTTPRouteAndGRPCRouteCoexistence|TestInferencePoolMultipleTargetPorts"'
```

prow/config/jobs/istio.yaml
jaiveerk: I think that would mess with the make command right? 

https://github.com/istio/istio/blob/e2896d1a2c66fac62e2c68d45eab0d0258e632c3/prow/integ-suite-kind.sh#L226-L227

This would just concatenate the two args and then make would look for one target called "test.integration.pilot.kube test.integration.pilot.gie.kube" right?

Can try to make this work if absolutely necessary but not sure what the value add is. Plus this would make it easier to add future additions in pilot, right? Where we'd just need to include them in the T val vs also needing to potentially add a new subpkg?

prow/config/jobs/istio.yaml
jaiveerk: added quotes!

Stevenjin8 (approved):

Stevenjin8:
@jaiveerk need to make gen

PR #5948, opened 2026-06-23
AWS Infrastructure

Stevenjin8:
The migration order will be:

* create the infrastructure (terrafrom resources)
* create prow resources
* move master jobs/new release jobs to aws
* move bots over (this will be super annoying)

This just does the raw infrastructure resources. A few maybe controversial points

* we are managing helm charts in helm
* we are using community-maintained `terraform-aws-modules`
* we are create a prow s3 bucket, but not using it (just in case we need it in the future)
* We are not (yet) creating a certificate

### DIFF
--- a/infra/aws/istio/cluster_autoscaler.tf
+++ b/infra/aws/istio/cluster_autoscaler.tf
@@ -1,13 +1,3 @@
-# Cluster Autoscaler. One deployment per cluster; each one scales only the
-# managed node groups of its own cluster via ASG auto-discovery. EKS managed
-# node groups automatically tag their Auto Scaling Groups with
-# k8s.io/cluster-autoscaler/enabled and k8s.io/cluster-autoscaler/<cluster>, so
-# no extra tagging is required for discovery.
-#
-# Each controller's IAM role is delivered via EKS Pod Identity: the association
-# binds the role to the kube-system/cluster-autoscaler service account, scoped
-# to that cluster's ASGs only.
-
 locals {
   cluster_autoscaler_chart_version = "9.58.0"
   cluster_autoscaler_image_tag     = "v1.35.0"
--- a/infra/aws/istio/cluster_autoscaler.tf
+++ b/infra/aws/istio/cluster_autoscaler.tf
@@ -1,0 +1,143 @@
+# Cluster Autoscaler. One deployment per cluster; each one scales only the
+# managed node groups of its own cluster via ASG auto-discovery. EKS managed
+# node groups automatically tag their Auto Scaling Groups with
+# k8s.io/cluster-autoscaler/enabled and k8s.io/cluster-autoscaler/<cluster>, so
+# no extra tagging is required for discovery.
+#
+# Each controller's IAM role is delivered via EKS Pod Identity: the association
+# binds the role to the kube-system/cluster-autoscaler service account, scoped
+# to that cluster's ASGs only.
+
+locals {
+  cluster_autoscaler_chart_version = "9.58.0"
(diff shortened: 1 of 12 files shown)


infra/aws/istio/eks.tf
```diff
@@ -0,0 +1,160 @@
...
+      }
+    }
+
+    # Private / PSWG build cluster (GKE `prow` in istio-prow-private).
+    prow-private = {
+      node_groups = {
+        # Test pool (e2-standard-16).
+        test = {
+          ami_type       = "AL2023_x86_64_STANDARD"
+          instance_types = ["t3.small"]
```
Stevenjin8, line 84: We can optimize these later.

infra/aws/istio/cluster_autoscaler.tf
```diff
@@ -0,0 +1,133 @@
+locals {
+  cluster_autoscaler_chart_version = "9.58.0"
+  cluster_autoscaler_image_tag     = "v1.35.0"
+}
+
+module "cluster_autoscaler_identity" {
```
keithmattix, line 6: nit: do we really need to prefix this with cluster_autoscaler each time? I was initially confused about what was being installed (especially when compared with the below helm_releases)

infra/aws/istio/eks.tf
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    # Core build cluster (for both amd and arm)
+    prow-build = {
+      node_groups = {
+        # Large build pool
+        build = {
+          ami_type       = "AL2023_x86_64_STANDARD"
+          instance_types = ["m6i.4xlarge"]
+          capacity_type  = "ON_DEMAND"
```
keithmattix, line 27: Do we really want on demand for everything? I guess maybe this is just to start before we fully migrate?

infra/aws/istio/eks.tf
```diff
@@ -0,0 +1,159 @@
...
+  subnet_ids = module.vpc[each.key].private_subnets
+
+  # Auth via EKS access entries (API). This is the boundary mechanism that
+  # replaces GCP project isolation: each principal is granted RBAC only on the
+  # clusters it may use. The Terraform principal gets admin to bootstrap.
+  authentication_mode                      = "API"
+  enable_cluster_creator_admin_permissions = true
+
+  # Skeleton: public endpoint so the API is reachable while building out.
+  # Lock this down (private endpoint + CIDR allow-list) before production.
```
keithmattix, line 137: Agree

infra/aws/istio/eks.tf
Stevenjin8: The other option is spot, which i'd generally like to avoid

keithmattix: Is on demand not spot? I assumed it was based on the intel pool comment

infra/aws/istio/cluster_autoscaler.tf
Stevenjin8: I'm not great with terrafrom ,but my understanding is that all these names share a namespace across all aws files, so the prefix makes it clear that this identity if for cluster_autoscaler, rather than ESO.

keithmattix: Ah, I realized later on that AWS doesn't ship with cluster autoscaler by default, so this is a common identity (name) for each cluster. Makes perfect sense

infra/aws/istio/eks.tf
Stevenjin8: they are different things. I will update the comment.

keithmattix (approved):

PR #5954, opened 2026-07-07
Set Up AWS Cluster with Prow infra

Stevenjin8:
@keithmattix 

I tried to make this easy to review by  having organized commits.

My first commit just moves prow files into prow/gcp

The second commit creates the necessary stuff for prow to work on eks.
Its pretty much a mirror image of gcp other than
* we use the external secrets operator instead of the deprecated kubernetes external secrets
* We don't have to explitcity do identity binding on kubernetes service accounts
* **there are no actual prowjobs** this
* we use aws-specific ingress, which isn't fully set up yet
* we don't have monitoring set up
* we don't have a dedicated arm cluster

third commit updates the Makefiles

the other ones are just cleanup/typos


prow/aws/cluster/gcsweb/gcsweb_ing.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
```
keithmattix, line 2: This makes me sad, but I get it 😜

keithmattix (approved):

PR #5986, opened 2026-07-29
Remove jobs for istio/enhancements repo

jewertow:
The validate-features_enhancements_release-* job fails since 1.19, because tools/schema-validator and validate_schema.sh were removed intentionally in these PRs: https://github.com/istio/tools/pull/2533, https://github.com/istio/test-infra/pull/4830. PRs to istio/enhancements were approved regardless of the validate-features job failures, so I do not see any sense in keeping this job in prow config.


fjglira:
I agree that we should remove the job at least until we have a proper fix or solution to the real issue described in the issue linked. If we create or reopen an issue to onboard the addition again of the job I'm ok for removing the job, at the end we have being having the job failing for over two years

jewertow:
Hey @howardjohn, do you remember why you [removed](https://github.com/istio/tools/pull/2533) schema-validator if it's still used by istio/enhancements?

fjglira:
Can you rebase, I'0ll approve the PR. It make sense to remove and keep the issue until an alternative is set

keithmattix:
Agreed; let's move forward here

PR #5987, opened 2026-07-29
Fix envoy branch for proxy 1.31 job

jwendell:


jwendell:
This is causing the proxy 1.31 being updated with Envoy main, instead of 1.39 => https://github.com/istio/proxy/pull/7183

cc @jewertow

keithmattix (approved):

PR #5989, opened 2026-07-29
Start moving jobs to AWS

Stevenjin8:
I can see the light!

This PR bumps up the size of our nodes, gives them more storage, and installs the aws storage driver.
It also adds a secrets to push to ghcr (unused righ now)

It moves all private jobs to the aws cluster
It also moves all ztunnel, api, ~and istio.io~ jobs on master and release-1.31


Makefile.core.mk
```diff
@@ -32,24 +32,23 @@ gen-check: gen check-clean-repo
...
-# 	@(cd tools/prowgen/cmd/prowgen; go run main.go --input-dir=$(repo_root)/prow/aws/config/jobs --output-dir=$(repo_root)/prow/aws/cluster/jobs write)
-# 	@go run tools/prowtrans/cmd/prowtrans/main.go --requirement-presets=./prow/aws/config/jobs/.base.yaml --configs=./prow/aws/config/istio-private_jobs --input=./prow/aws/config/jobs
-# 	# experimental jobs are intentionally not generated for EKS (config removed)
-# 	# @go run tools/prowtrans/cmd/prowtrans/main.go --requirement-presets=./prow/aws/config/jobs/.base.yaml --configs=./prow/aws/config/experimental --input=./prow/aws/config/jobs
+generate-config-aws: generate-config
+	@rm -fr prow/aws/cluster/jobs/*/*/*.gen.yaml
+	@(cd tools/prowgen/cmd/prowgen; go run main.go --input-dir=$(repo_root)/prow/aws/config/jobs --output-dir=$(repo_root)/prow/aws/cluster/jobs write)
+	@go run tools/prowtrans/cmd/prowtrans/main.go --requirement-presets=./prow/aws/config/jobs/.base.yaml --configs=./prow/aws/config/istio-private_jobs --input=./prow/aws/config/jobs
+	# @go run tools/prowtrans/cmd/prowtrans/main.go --requirement-presets=./prow/aws/config/jobs/.base.yaml --configs=./prow/aws/config/experimental --input=./prow/aws/config/jobs
```
keithmattix, line 44: Maybe we should swap this and start aws on experimental jobs first? Do we have enough of those?

infra/aws/istio/eks.tf
```diff
@@ -1,5 +1,10 @@
 locals {
   cluster_version = "1.36"
+  gp3_root_volume = {
+    delete_on_termination = true
```
keithmattix, line 4: What are the implications of this?

keithmattix (commented):
It looks like we're deleting more jobs from GCP than what we replace with aws. Can you double check and ensure those are all intended?

Makefile.core.mk
Stevenjin8: I don't think we use the `experimental-*` branches anymore.

infra/aws/istio/eks.tf
Stevenjin8: This just expands the root fs to have more storage. We keep a go/rust cache on a hostVolume and generally don't want to keep these volumes around if the node gets spun down. This is also the default (at least on EC2)

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configure-root-volume-delete-on-termination.html

Stevenjin8:
@keithmattix good catch on the private jobs.

testgrid/config.yaml
```diff
@@ -1,8 +1,6 @@
 # Dashboards need to be specified here
 # A prow annotation will be invalid if it references a dashboard that doesn't exist
 dashboards:
-- name: istio_api
```
Stevenjin8, line 4: we ignore testgrid for now and I'll move it once all the jobs have moved.

Makefile.core.mk
keithmattix: I know windows at least is on there but, fair enough

keithmattix (approved):

PR #5991, opened 2026-07-30
Migrate more jobs from GCP to AWS

Stevenjin8:
everything except for tools, release-builder, and ~istio.io~ istio because those are the ones that push to container registries.


keithmattix:
Let's maybe see how the current set of jobs are going

Stevenjin8:
@keithmattix I don't mind waiting, but I would rather merge this one and then wait. Currently we only watch istio/api and istio/ztunnel, which doesn't give us a lot of signal. The repos this PR introduces don't have a lot of traffic, so its not much of a change from a contributors point of view. This PR also leaves istio/istio, istio/tools, and istio/release-building out, which are the ones that might give us more trouble.

keithmattix:
Ahh I was reading the description and thought we were moving istio over. Ok this set is fine

keithmattix (approved):

PR #6004, opened 2026-08-12
Add presubmit job for pilot integ tests with nativeNftables

sridhargaddam:
This PR adds a new optional presubmit job to validate all Istio pilot integration tests with the nativeNftables backend and CNI enabled. The existing pilot integration tests mainly validate the default iptables backend, leaving a significant gap in coverage for the nftables backend.

This complements the existing integ-pilot job, which uses the default iptables backend, and helps catch regressions in the nftables traffic redirection path early in the development cycle.

The job is marked as presubmit_optional to give it soak time before promoting it to a required check.

Corresponding PR in the Istio repo: https://github.com/istio/istio/pull/61216

### DIFF
--- a/prow/gcp/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
+++ b/prow/gcp/cluster/jobs/istio-private/istio/istio-private.istio.master.gen.yaml
@@ -5356,6 +5356,91 @@ presubmits:
     - ^master$
     cluster: private
     decorate: true
+    name: integ-pilot-cni-nftables_istio_pri
+    optional: true
+    path_alias: istio.io/istio
+    rerun_command: /test integ-pilot-cni-nftables
+    spec:
+      automountServiceAccountToken: false
+      containers:
+      - command:
+        - entrypoint
+        - prow/integ-suite-kind.sh
+        - test.integration.pilot.kube
+        env:
+        - name: BAZEL_BUILD_RBE_INSTANCE
+        - name: BUILD_WITH_CONTAINER
+          value: "0"
+        - name: GOMAXPROCS
+          value: "5"
+        - name: INTEGRATION_TEST_FLAGS
+          value: --istio.test.istio.enableCNI=true --istio.test.nativeNftables=true
+        - name: DEPENDENCIES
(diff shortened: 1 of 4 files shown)


fjglira:
I'm in favor of adding this job, and I think having it at the beginning having it as optional to test it and see how stable is

Stevenjin8:
@sridhargaddam @fjglira Any concerns about making this a postsubmit? I suspect that we'll have to cut down on presubmit jobs to save money? Regardless, I'll likely start worrying about this in a few weeks. For now +1 on optional.

sridhargaddam:
> @sridhargaddam @fjglira Any concerns about making this a postsubmit? I suspect that we'll have to cut down on presubmit jobs to save money? Regardless, I'll likely start worrying about this in a few weeks. For now +1 on optional.

I can understand the concern. Let’s start with it as an optional presubmit job, and once we have some confidence that the tests are passing, we can either move it to postsubmit or explore modifying one of the existing presubmit jobs (f.e., `integ-cni_istio`) to use the nftables backend instead of the iptables backend. WDYS?

**Edit:** Just to give you one more perspective, currently only a small subset of the Pilot integration tests are being validated against the nftables backend as part of the `integ-pilot_istio` CI job. I [pushed this dummy PR](https://github.com/istio/istio/pull/61251) to show how we could extend the existing `integ-pilot_istio` CI job to validate the tests with the nftables backend as well. As you can see, this approach adds quite a bit of complexity to the test code and doesn’t feel like a great long-term solution.

That’s why I think introducing a new optional CI presubmit for nftables, at least for now, would be a cleaner approach and we can revisit this job once we have some stats.

sridhargaddam:
@keithmattix @Stevenjin8 Please take a look at my previous comment about keeping this as an optional presubmit job for now. We can revisit and update it later if needed based on how things evolve. Happy to hear your thoughts and discuss further. Thanks!

keithmattix:
This looks good for now, but we should consider setting up our e2e framework to run multiple tests in succession on the same infra (especially for some that are experimental or have been around for awhile)

Stevenjin8 (approved):

PR #6005, opened 2026-08-12
pack nodes better and remove duplicate extensions

Stevenjin8:
PR introduces various hints/pdb/node selectors that will help us pack our pods better and save some money $$$$$. Clusters now scale down to one node.

also remove duplicate extensions.


prow/gcp/plugins.yaml
```diff
@@ -21,78 +21,19 @@ config_updater:
...
-    plugins:
-    - golint
-    - hold
-
-  istio-ecosystem/sail-operator:
-    plugins:
-    - assign
-    - hold
-
-external_plugins:
```
keithmattix, line 76: Why are we removing this?

Stevenjin8: This is already present in the aws cluster, so we don't need it twice. The main one was the cherry picker because it was trying to cherrypick twice (see https://github.com/istio/istio/pull/61228#issuecomment-5274733213)

keithmattix (approved):

PR #6011, opened 2026-08-20
Add test coverage for k8s 1.36 on master branch

jewertow:
The [support status](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases) table shows that we support k8s 1.36, but 1.35 is the highest version used in tests for now, so I upgraded the default node version and added job for 1.35.


jewertow:
cc: @dhawton @kfaseela

jewertow:
Sorry, I see that it shouldn't be updated here.

PR #6041, opened 2026-09-07
Moving Sail Opeartor jobs to AWS

fjglira:
The testgrid config validator (testgrid/config_test.go) only reads `prow/gcp/cluster/jobs` to generate testgrid tabs. After sail-operator jobs for main, release-1.29, release-1.30, and release-1.31 moved to the AWS prow cluster in the PR, the four dashboards explicitly declared in `testgrid/config.yaml` had zero tabs, causing validation to fail with "contains no tabs".
Added an `awsJobPath` flag that default: `../prow/aws/cluster/jobs/istio-ecosystem` and a `mergedJobDir()` helper that symlinks gen.yaml files from both GCP and AWS into a single temp directory. The testgrid configurator now sees jobs from both clusters and can populate tabs for all declared dashboards.
Scoped to istio-ecosystem only to avoid surfacing a pre-existing, unrelated issue where the AWS release-builder master `gen.yaml` references dashboards not yet declared in `testgrid/config.yaml`.


prow/aws/config/jobs/sail-operator-release-1.30.yaml
```diff
@@ -4,7 +4,7 @@ branches:
   - release-1.30
 
 support_release_branching: false
-image: gcr.io/istio-testing/build-tools:release-1.30-d868c62ea286fa9370c5202c29c25e755eaa71d6
+image: registry.istio.io/testing/build-tools:release-1.30-d868c62ea286fa9370c5202c29c25e755eaa71d6
```
FilipB, line 7: Note that this image does not exist:
`Error: unable to copy from source docker://registry.istio.io/testing/build-tools:release-1.30-d868c62ea286fa9370c5202c29c25e755eaa71d6: initializing source docker://registry.istio.io/testing/build-tools:release-1.30-d868c62ea286fa9370c5202c29c25e755eaa71d6: reading manifest release-1.30-d868c62ea286fa9370c5202c29c25e755eaa71d6 in registry.istio.io/testing/build-tools: manifest unknown`

It's only available in `gcr.io`, not in the new registry where the `registry.istio.io` is pointing to now.

FilipB (changes requested):

prow/aws/config/jobs/sail-operator-release-1.30.yaml
Stevenjin8: @fjglira I would keep everything <= 1.30 in gcp

FilipB: @Stevenjin8 what about common [files](https://github.com/istio/common-files/blob/release-1.30/files/.devcontainer/devcontainer.json#L3)? Is there going to be a new build tool image available in `registry.istio.io` or we need to switch to gcp? Part of our automation is considering the image referenced in common files as source of truth.

Stevenjin8: right... I need to do that. switching that to gcp would fix it, but I think I need to update 1.30 CI to also push to ghcr.io/istio/testing (which is what backs registry.istio.io/testing)

FilipB (approved):

prow/aws/config/jobs/sail-operator.yaml
```diff
@@ -4,7 +4,7 @@ branches:
   - main
 
 support_release_branching: false
-image: gcr.io/istio-testing/build-tools:master-344978549023107b14c57d2ef8523d41b82ccc76
+image: registry.istio.io/testing/build-tools:master-344978549023107b14c57d2ef8523d41b82ccc76
```
Stevenjin8, line 7: can we use the <account number>.dck.xxx registry? Its configured on all other istio jobs

Stevenjin8 (approved):

prow/aws/config/jobs/sail-operator.yaml
Stevenjin8: oops, I'll do this

PR #6045, opened 2026-09-08
Lower Resource Request and Use Pull Through Cache, etc

Stevenjin8:
* Pull through cache saves us hundreds a day because the build tools image is heavy and we have a lot of node churn.
* I suspect we can lower resource requests even more. Our integration tests are a bit unfortunate in that they both build istio and run integration tests. Build has very high CPU utilization (using all available cores), but running the kind tests is usually ~200 milli cpus. I suspect we can bump it down more.
* updates the automator script to keep the build-tools image pointing at the pull through cache.
* bumps our spot pool down so it can scale to 0 (and hopefully stay there)
* change our main pool to accept different similar VM SKUs so it is less likely we get capped by AWS capacity constraints
* Enabled metrics server so its easier to track resource utilization
* Give crane permissions to patch pods in remote clusters


infra/aws/istio/cluster_autoscaler.tf
```diff
@@ -151,5 +171,15 @@ resource "helm_release" "cluster_autoscaler_prow_private" {
...
+  set {
+    name  = "extraArgs.scale-down-unneeded-time"
+    value = "3m"
+  }
+
+  set {
+    name  = "extraArgs.scale-down-delay-after-add"
+    value = "3m"
+  }
+
```
Stevenjin8, line 183: Hav nodes scale down after 3 minutes of inactivity

fjglira:
Regarding this point:
```
I suspect we can lower resource requests even more. Our integration tests are a bit unfortunate in that they both build istio and run integration tests. Build has very high CPU utilization (using all available cores), but running the kind tests is usually ~200 milli cpus. I suspect we can bump it down more.
```

A better approach used in other projects is to build and push the PR image once in a pre-step, then have the tests pull that shared image. While pulling a large image adds a small overhead, sharing a single build will still save us significant CI time and money.

Stevenjin8:
@fjglira WDYM pre step? I don't think prow allows defining dependency graphs.

fjglira:
> @fjglira WDYM pre step? I don't think prow allows defining dependency graphs.

Yes, Prow doesn't allow dependency steps, but it can be worked around by using the Istio-bot or a GitHub Action that monitors when the first jobs pass (light jobs like linter, unit, gen check, etc) and adds the build and push of the image. Once those pass, the bot will comment /test integ-xxxxx to trigger the rest of the tests with SKIP_BUILD and the image tagged as PR_NUMBER_SHA, for example, so the build can be done once and the integration tests can pull those images to use in their tests, saving time and money. and it will allow to lower CPU resources because the build is going to be done once only

infra/aws/istio/eks.tf
```diff
@@ -62,11 +62,11 @@ locals {
           }
           labels = { testing = "test-pool" }
         }
-        test2 = {
+        testspot = {
```
fjglira, line 65: This will behave as a terraform destroy and create, so if there is any CI running at that point it will be killed

fjglira: I mean, it's not a real issue but just to taking into account

infra/aws/istio/cluster_autoscaler.tf
fjglira: Do you know what the default value is? I'm wondering if 3 minutes is too low, but in anycase we can adjust this again in case is too low

infra/aws/istio/eks.tf
```diff
@@ -62,11 +62,11 @@ locals {
...
           ami_type       = "AL2023_x86_64_STANDARD"
           instance_types = ["m7a.4xlarge", "m6a.4xlarge", "m7i.4xlarge", "m6i.4xlarge"]
           capacity_type  = "SPOT"
-          min_size       = 1
+          min_size       = 0
```
fjglira, line 69: If the desired size is one, a min size 0 it will provision a cluster and drain after 3 minutes when we apply this configuration. It's not a blocker but it's what is going to happen from what I can see

prow/aws/README.md
fjglira, line 1: Why remove this file instead of populating with the information?

tools/automator/scripts/update-images.sh
```diff
@@ -23,7 +23,7 @@ ROOT="$(cd -P "$(dirname -- "$0")" && pwd -P)"
 source "$ROOT/../utils.sh"
 
 # Defaults
-image='registry.istio.io/testing/build-tools.*'
+image='\(registry[.]istio[.]io/testing\|678412441677[.]dkr[.]ecr[.]us-west-2[.]amazonaws[.]com/ghcr/istio/testing\)/build-tools.*'
```
fjglira: Checking if this regex applies, I see:

```
OLD_SHA="7f5d77ade26da4b04ed42f353d9c26fbb3a4e7f1"
NEW_SHA="deadbeefcafebabe1234567890abcdef12345678"
image_broken='\(registry[.]istio[.]io/testing\|678412441677[.]dkr[.]ecr[.]us-west-2[.]amazonaws[.]com/ghcr/istio/testing\)/build-tools.*'

echo "image: 678412441677.dkr.ecr.us-west-2.amazonaws.com/ghcr/istio/testing/build-tools:master-${OLD_SHA}" \
  | sed -E "s,(image: $image_broken:)master-[0-9a-f]{40},\1${NEW_SHA},g"
```
Output SHA unchanged, no substitution occurred:
```
image: 678412441677.dkr.ecr.us-west-2.amazonaws.com/ghcr/istio/testing/build-tools:master-7f5d77ade26da4b04ed42f353d9c26fbb3a4e7f1
```

From what I see the correct regex should be:
`image='(registry[.]istio[.]io/testing|678412441677[.]dkr[.]ecr[.]us-west-2[.]amazonaws[.]com/ghcr/istio/testing)/build-tools[^:]*'`

fjglira (commented):
lgtm in general, just some minor comments and one error that I see in the update image bash script

tools/automator/scripts/update-images.sh
fjglira: Please do a test because I'm on macOS and sometimes sed tricks me

infra/aws/istio/cluster_autoscaler.tf
dhawton: I think default is 10

infra/aws/istio/eks.tf
dhawton: won't provision a cluster, just a node pool.. no reason to pay for a node sitting around doing nothing.

infra/aws/istio/eks.tf
Stevenjin8: Yep, I already did it

prow/aws/README.md
Stevenjin8: It was a copilot generated file. Planning on document this better when everythign is in place

tools/automator/scripts/update-images.sh
Stevenjin8: good catch.

fjglira (approved):
LGTM

PR #6048, opened 2026-09-10
Less network ingress, separate trusted nodes, and trusted prow caches

Stevenjin8:
Ok, this ended up ballooning into a bunch of things. Happy to split this up, but want to make a lot of progress before the end of next week. **I have tried to split the commits cleanly so they are easy to review**. These are the tings we do

* We set environment variables to we pull various images from a pull-through cache
* We create a few new node pool for our release-esque jobs (anything that pushes or has access to secrets)
* We make a testing a release bazel s3 bucket and mount the cache as a sidecar so we can use pod identity to authenticate.
* Use new test targets introduced by https://github.com/istio/istio/pull/61677
* Remove k8s integration tests with versions <= 1.32
* Use athens goproxy to reduce network ingress
* Increase disk throughput to make compilation and kind setup faster
* Add a cronjob that rolls out crier and deck private (credentials become stale)


infra/aws/istio/eks.tf
```diff
@@ -33,23 +34,29 @@ locals {
         # Large build pool
         build = {
           ami_type       = "AL2023_x86_64_STANDARD"
-          instance_types = ["m6i.16xlarge"]
+          instance_types = ["m7i.16xlarge", "m7i.16xlarge"]
```
fjglira, line 37: The instance type is duplicated here; it's a typo, and we want to set another instance type?

infra/aws/istio/eks.tf
```diff
@@ -33,23 +34,29 @@ locals {
...
           capacity_type  = "ON_DEMAND"
           min_size       = 0
           max_size       = 5
           desired_size   = 1
           block_device_mappings = {
             root = {
               device_name = "/dev/xvda"
-              ebs         = merge(local.gp3_root_volume, { volume_size = 2000 })
+              ebs         = merge(local.gp3_root_volume, { volume_size = 512 })
```
fjglira, line 45: Can we hit disk pressure because the disk reduction is 75%? Do you know which can be the current disk utilization is?

infra/aws/istio/eks.tf
```diff
@@ -221,13 +355,21 @@ module "eks" {
...
-        try(config.iam_role_additional_policies, {}),
-        {
-          ecr_pull_through_cache_read = "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly"
-        },
-      )
-    })
+    for name, config in each.value.node_groups : name => merge(
+      config,
+      # AWS does this dumb thing where it rebalances nodes and kills our pods.
+      # This pins the node group to a single AZ that AWS won't auto rebalance.
```
fjglira, line 361: If we ping both prow-build and prow-private to a single AZ, we can introduce a single point of failure. Maybe it's worth checking if we can introduce a pod disruption budget to avoid this kind of issue.

fjglira (commented):
Some initial review, I need to take more time to check more

prow/aws/config/jobs/istio.yaml
```diff
@@ -422,162 +422,6 @@ jobs:
     command: [entrypoint, prow/integ-suite-kind.sh, --node-image, 678412441677.dkr.ecr.us-west-2.amazonaws.com/ghcr/istio/testing/kind-node:v1.36.0, test.integration.helm.kube]
     requirements: [kind, ecr]
 
-    # The node image must be kept in sync with the kind version we use.
-    # See istio.io/tools/docker/build-tools to build a kind image
-  - name: integ-k8s-123
```
fjglira, line 427: Are we officially going to remove those older versions? We need to let people know that is going to happen if there are users who rely on those jobs. Maybe let them know how to run those test on those version will be good in case they wanted to continue running in some way

prow/aws/config/jobs/.base.yaml
```diff
@@ -159,6 +159,14 @@ requirement_presets:
     env:
     - name: ECR_REGISTRY
       value: 678412441677.dkr.ecr.us-west-2.amazonaws.com
+    - name: ISTIO_BASE_REGISTRY
+      value: 678412441677.dkr.ecr.us-west-2.amazonaws.com/docker/istio
+    - name: ISTIO_BUILDKIT_IMAGE
+      value: 678412441677.dkr.ecr.us-west-2.amazonaws.com/ghcr/istio/testing/buildkit:v0.11.0
```
fjglira, line 165: We will need to set update over this image also to keep it up to date I see that latest one is v0.13.1

infra/aws/istio/eks.tf
dhawton: It doesn't matter.. prow-build and prow-private never talk or work together. prow-private is specifically only for PSWG work. prow-build is used for the open/public repos.

prow/aws/config/jobs/istio.yaml
dhawton: Yes, we need to update docs and probably a blog post, to say we're no longer testing these. Our docs still list these as tested so I'd prefer that be communicated before we just drop them.

fjglira: I'll prepare today a PR for this on istio.io

infra/aws/istio/eks.tf
Stevenjin8: 512 GB is plenty plenty/ I think we'll have at most 8 prowjobs running on one node at once.

prow/aws/config/jobs/.base.yaml
Stevenjin8: I think we have historically manually updated this one. I only see one right now

```
10:12:17 (prow-build) /home/sj/test-infra $ crane ls ghcr.io/istio/testing/buildkit
v0.11.0
```

infra/aws/istio/eks.tf
fjglira: U just asked because it surprised that it was so much reduction :)

prow/aws/config/jobs/istio.yaml
fjglira: Blogpost PR: https://github.com/istio/istio.io/pull/17617

fjglira (approved):
lgtm. Let's leave the hold until @dhawton take a look also

fjglira:
Let's wait also until we merge the blog post

prow/aws/config/jobs/istio.yaml
```diff
@@ -422,162 +422,6 @@ jobs:
...
-      - --node-image
-      - 678412441677.dkr.ecr.us-west-2.amazonaws.com/ghcr/istio/testing/kind-node:v1.32.0
-      - --kind-config
-      - prow/config/modern.yaml
-      - test.integration.kube
-    requirements: [kind, ecr]
-    timeout: 4h
-    env:
-      - name: INTEGRATION_TEST_FLAGS
-        value: " --istio.test.retries=1 "
```
dhawton, line 579: 1.32 is a supported Kubernetes version of Istio 1.31. I would really prefer we not change this post .0 release without a lot more community buy-in. https://istio.io/latest/docs/releases/supported-releases/

Stevenjin8: @dhawton this doesn't change the istio 1.31 tests, only the master ones, but I can keep it if you still feel strongly about this

dhawton (approved):

prow/aws/config/jobs/istio.yaml
dhawton: Nope, I misunderstood.

dhawton:
@Stevenjin8 we have the blog going out and since this really only drops the old integ tests for master (and 1.32) and doesn't affect the current releases, I am happy for this to merge. I left DNM for your last look over.

PR #6050, opened 2026-09-14
Run `build` and `publish` release builder jobs on nodes with on-demand capacity

jewertow:
This is an equivalent of this PR https://github.com/istio/test-infra/pull/6007 for AWS.

This change will prevent random job failures, like recent 1.31.1 build failure: https://aws.prow.istio.io/view/s3/istio-prow/logs/release-arm64_ztunnel_release-1.31_postsubmit/2097241431558066176.


fjglira (approved):
lgtm, it will use from now on only nodes with testing: test-pool and capacityType ON_DEMAND avoiding SPOT instances

Stevenjin8:
@jewertow, I think this might have been an amd vs arm problem

PR #6054, opened 2026-09-16
Fix the `update-proxy` job

jwendell:
It needs to run in the proxy image, as it needs `bazel` now.


jwendell:
https://aws.prow.istio.io/view/s3/istio-prow/logs/update-proxy_proxy_periodic/2100239371218194432#1:build-log.txt%3A106

Stevenjin8 (approved):

PR #6059, opened 2026-09-20
Add support for k8s 1.37

jewertow:
This PR adds postsubmit jobs with k8s 1.37 to master and 1.31 branches. If it works, I will update the default kind image to 1.37 and replace `integ-k8s-137` with `integ-k8s-136`.


jewertow:
@Stevenjin8 I guess we need the kind image in ECR, right? Could you push it?

Stevenjin8 (approved):

PR #6008, opened 2026-08-18
AWS: migrate main repos

Stevenjin8:
* Move tools, istio, release-builder to aws (only master and release-1.31)
* update repository mappings to point to ghcr.io/istio-testing
* private jobs still need some work. I would like to merge this as-is and immediately follow up fixing the private jobs. Release branches older than 1.31 should not be affected
* update our signing key to use an aws one.

Since this is a big change, I'm going to wait for this pr to be approved, apply the changes, fix the things i inevitably miss, and once everything seems good, will request another review and merge.

PR #6061, opened 2026-09-21
add r2 release creds to gcp release-builder publish jobs for 1.29 and 1.30

PetrMc:
publish.sh on release-1.29/1.30 uploads to R2 since istio/release-builder#2419 and istio/release-builder#2420 but the gcp publish jobs never got cf-r2-istio-release-auth, so 1.29.8 and 1.30.5 publish failed with CF_CREDENTIALS unbound. Matches the release-1.31 aws job.


PetrMc:
also dropped the istio-ecosystem main/1.31 sail-operator dashboards from testgrid: #6041 moved those jobs to aws but the configurator (both the presubmit check and the trusted upload to gs://istio-testgrid) only reads prow/gcp/cluster/jobs, so they had no tabs and the check has been failing since sep 16. reverted the config_test.go merge from #6041 so the go test matches what prod generates, same as #6008 did for the istio jobs

fjglira (approved):
Lgtm. Let's rerun and check

